Issue #26, opened 2020-09-03
Add tags, filtering


andrewvc:
It should be possible to tag journeys, and, when running a suite, filter by tag name.

I propose we add `--include-tags tag1,tag2` and `--exclude-tags tag1,tag2` as options.

PR #60, opened 2020-09-25
Add tags to journeys + filtering

andrewvc:
Fixes elastic/synthetics#26 adds tags as metadata to journeys that can be filtered on. These tags will be searchable / filterable in the Uptime UI


vigneshshanmugam:
Closing in favor of #300

PR #111, opened 2020-10-29
Update README.md

shahzad31:
Added an example

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,7 +21,7 @@ If you'd like to stay informed about synthetic monitoring development please vis
 ### To Run Example Journey with Docker and hearbeat, 
 Data will be pushed to ES, which can be viewed using uptime app
 
-`node ./examples/docker/run.sh 7.10.0  -E output.elasticsearch.hosts=["http://localhost:9200"] -E output.elasticsearch.username=admin -E output.elasticsearch.password=changeme`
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -8,7 +8,22 @@ are published.
...
+`npm run build`
+
+`npm link`
+
+`cat examples/inline/short.js | npx @elastic/synthetics --inline`
+
+### To Run Example Journey with Docker and hearbeat, 
+Data will be pushed to ES, which can be viewed using uptime app
+
+`node ./examples/docker/run.sh 7.10.0  -E output.elasticsearch.hosts=["http://localhost:9200"] -E output.elasticsearch.username=admin -E output.elasticsearch.password=changeme`
```
vigneshshanmugam, line 24: Bettet to move this under `examples/docker/readme.md` and also remove `node` from there

shahzad31: i think since it's pretty small for now. keep it at root makes sense

andrewvc: ++ that `node` breaks this. I would prefer we send people to the docs, mainly because READMEs tend not to be updated very frequently, and the docs are mostly done.

That said, I think it'd be a great idea to improve the README for the specific example. If the README is in the same folder it's much more likely to be updated.

shahzad31: i already removed node thingy, if @vigneshshanmugam is doing a separate readme after package publgish, we can close this, he mentioned that.

PR #112, opened 2020-10-29
fix: close fd after debug logs are written

### LINKED ISSUE #110: End of descriptor error on reporter fd
+ When there are async activity and if that function internally calls our `logger`, we would be writing to that FD and that could be closed by our runner which closes the stream on `process.nextTick`. 
+ This happens only on debug mode. 

### TLDR: 

process.nextTick gets called in the event loop before the micro task queue and we have to close the stream as the last task in the queue.

vigneshshanmugam:
+ fix #110 
+ runner end should be called after reset since we are logging the output inside reset
+ Moved from `nextTick` to `setTimeout` as its called after all the micro tasks and ensure all log statements are printed before we end stream.

### DIFF
--- a/src/core/runner.ts
+++ b/src/core/runner.ts
@@ -328,8 +328,8 @@ export default class Runner {
       result[journey.name] = journeyResult;
     }
     await this.runAfterAllHook();
-    this.emit('end', {});
     this.reset();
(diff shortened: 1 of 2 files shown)


andrewvc (approved):
LGTM

PR #114, opened 2020-10-29
fix: ensure we execute tests only once per run

### LINKED ISSUE #113: Tests run multiple times if invoked via CLI and progammatically
+ When the test is invoked via CLI and the user has the programmatic API `run`, it will run the journeys multiple times based on when it got registered. 

```ts
import { journey, step, run } from '@elastic/synthetics';
import { deepStrictEqual } from 'assert';
import { join } from 'path';

// calling run here programmatically as well
run({});

journey('check that title is present', async ({ page }) => {
  step('go to app', async () => {
    await page.goto(path);
  });
});
```

Now ,run the above journey
```
npx @elastic/synthetics journey.ts -d
```

vigneshshanmugam:
+ fix #113 
+ It ensures if the current run is active, we don't schedule new runs even invoked via CLI or another programmatic run

### DIFF
--- a/__tests__/index.test.ts
+++ b/__tests__/index.test.ts
@@ -23,11 +23,40 @@
  *
  */
 
-import { run } from '../src/index';
+import { run, journey } from '../src/index';
 import { runner } from '../src/core';
 import * as ParseArgs from '../src/parse_args';
+import { generateTempPath } from '../src/helpers';
+import fs from 'fs';
 
-describe('run', () => {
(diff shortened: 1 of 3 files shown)


andrewvc (changes requested):
Can we add a test for this behavior?

vigneshshanmugam:
@andrewvc Added the test.

andrewvc (approved):
LGTM

PR #116, opened 2020-10-29
E2e test poc

shahzad31:
This is a POC for how we can make sure e2e things are working

1. It will run examples journeys through docker
2. Will setup kibana, es
2. Will run an inline journey to make sure data is being populate in kibana uptime app for those example journeys

You can test by running

npm run test inside __tests/e2e dir


.idea/misc.xml
```diff
@@ -0,0 +1,6 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
andrewvc, line 1: looks like we need to exclude IDE files in gitignore

andrewvc: Or rather that some were committed before being excluded

e2e/package-lock.json
```diff
@@ -0,0 +1,6235 @@
+{
```
vigneshshanmugam, line 1: lets get rid of lockfiles for examples and tests. Create `.npmrc` file with `package-lock=false`

e2e/package.json
```diff
@@ -0,0 +1,10 @@
+{
+  "name": "e2e-synthetics",
+  "version": "1.0.0",
```
vigneshshanmugam, line 3: private: true and remove versioning and all other unnecessary details.

e2e/run-e2e.sh
```diff
@@ -0,0 +1,145 @@
...
+mkdir -p ${TMP_DIR}
+
+
+# Start synthetics docker examples
+echo "" # newline
+echo "${bold}Starting synthetics docker examples${normal}"
+echo "" # newline
+cd ${DOCKER_DIR}
+
+./run.sh 7.10.0 \
```
vigneshshanmugam, line 43: better to just pass it to the `run.sh` script instead of handling the version and heartbeat args on both places.

e2e/run-e2e.sh
```diff
@@ -0,0 +1,145 @@
...
+    -E output.elasticsearch.hosts=["localhost:9201"] \
+    -E output.elasticsearch.username=admin \
+    -E output.elasticsearch.password=changeme \
+     > ../../e2e/tmp/synthetics.log 2>&1 &
+
+cd ../../e2e/
```
vigneshshanmugam, line 49: Hmmm, we can simply call `examples/docker/run.sh` instead of going inside and out.

e2e/run-e2e.sh
```diff
@@ -0,0 +1,145 @@
...
+#
+# run e2e tests journey
+##################################################
+
+
+
+npx @elastic/synthetics uptime.journey.ts
+
+
+if [ $? == 1 ]; then
```
vigneshshanmugam, line 129: This wont work for the synthetics run as i mentioned in our call.

vigneshshanmugam (commented):
Thanks for starting this Shazad. Couple of suggestions. Keeping bunch of commands on `package.json` would ease up the process locally and we can even invoke them on CI. 

+ [x] npm run `setup` - would kick in and start necessary images
+ [ ] npm run `test` - would make sure kibana is running locally and starts and verifies the runner
+ [ ] move the root level `e2e` folder inside tests as we dont want to have mutiple root level directories.

__tests__/e2e/scripts/test.sh
```diff
@@ -0,0 +1,40 @@
...
+
+# formatting
+bold=$(tput bold)
+normal=$(tput sgr0)
+
+# Wait for Kibana to start
+##################################################
+echo "" # newline
+echo "${bold}Waiting for Kibana to start...${normal}"
+yarn wait-on -i 500 -w 500 http-get://admin:changeme@localhost:$KIBANA_PORT/api/status > /dev/null
```
shahzad31, line 21: @jahtalab is there a way we can improve this command?

jahtalab: I think we should refactor the setup into the docker-compose, so we can add dependency for both ES and kibana and docker-compose would make sure to bring up those before running the tests.

shahzad31: dependency won't make sure kibana is up, it only means kibana docker container is up. Kibana itself takes a while to start.

__tests__/e2e/docker-compose.yml
```diff
@@ -0,0 +1,58 @@
...
+#      -E output.elasticsearch.username=elastic
+#      -E output.elasticsearch.password=changeme"
+#
+#    networks:
+#      - elastic
+
+volumes:
+  data01:
+    driver: local
+  data02:
```
andrewvc, line 51: We don't need these extra volumes that are not in use

__tests__/e2e/package.json
```diff
@@ -0,0 +1,14 @@
+{
+  "name": "e2e-synthetics",
+  "private": true,
+  "main": "index.js",
+  "license": "MIT",
+  "scripts": {
+    "setup": "sh scripts/setup.sh",
```
andrewvc, line 7: This should either be `scripts/setup.sh` or `bash scripts/setup.sh`. That shell script requires bash features not present in the minimal `sh` binary.

__tests__/e2e/scripts/setup.sh
```diff
@@ -0,0 +1,50 @@
+#!/usr/bin/env bash
```
andrewvc, line 1: For bash scripts it's generally a good practice to add the line:

`set -e`  which will cause the script to exit on error. This [so answer](https://stackoverflow.com/questions/9952177/whats-the-meaning-of-the-parameter-e-for-bash-shell-command-line/9952249#:~:text=2%20Answers&text=The%20%2De%20option%20means%20%22if,t%20a%20real%20%22error%22.) provides a good overview of the implications of this (and a few downsides)

__tests__/e2e/package.json
```diff
@@ -0,0 +1,14 @@
...
+    "test": "npm i && npm run setup &&  sh scripts/test.sh"
```
andrewvc, line 8: Same issue here with` sh`

__tests__/e2e/uptime.journey.ts
```diff
@@ -0,0 +1,45 @@
...
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ *
+ */
+
+import { journey, step } from '@elastic/synthetics';
+
+journey('E2e test synthetics', async ({ page }) => {
```
andrewvc, line 28: I think this is a fine start, but let's follow-up soon with some more substantive tests

__tests__/e2e/scripts/test.sh
andrewvc: Yes, that's right, however, I do think it makes more sense to just use a docker compose file for the setup of synthetics, and for that matter, we should probably just get rid of run.sh eventually and replace it with docker-compose.

We can monitor for the synthetics container starting by checking the docker logs.

I'm not sure if the shell scripts from APM, in the end, are that great. I can see things being much simpler without them, we just need to:

1. Run `docker-compose up`
2. Wait for kibana to start
3. Run the synthetics tests in a loop for some length of time until they pass

We could optionally have a step between 2 and 3 where we monitor the output of heartbeat waiting for the tests to run, but I don't think we really need that. We're testing the functionality end to end after all, we just care that results eventually appear in Uptime

andrewvc (commented):
This is a great improvement, but I think the shell scripts may create more issues than they solve. Left the details of my concerns there in the comment with @jahtalab 

I also left some comments on improving the shell scripts if indeed that is the way we want to go.

Another thought, we could use `node` instead of bash to drive any orchestration. Shell gets unwieldy sometimes

__tests__/e2e/scripts/test.sh
jahtalab: @shahzad31 , we can add a healthcheck to make sure kibana is actually up before running the tests.

shahzad31: @jahtalab i have added health checks

examples/todos/package.json
```diff
@@ -5,7 +5,7 @@
   "scripts": {},
   "license": "MIT",
   "dependencies": {
-    "@elastic/synthetics": "^0.0.1-alpha.6",
+    "@elastic/synthetics": "file:../../../elastic-synthetics.tgz",
```
vigneshshanmugam, line 8: Need to be reverted, this file does not exist atm.

__tests__/e2e/package.json
```diff
@@ -0,0 +1,14 @@
...
+  "name": "e2e-synthetics",
+  "private": true,
+  "main": "index.js",
+  "license": "MIT",
+  "scripts": {
+    "setup": "bash scripts/setup.sh",
+    "test": "npm i && npm run setup &&  bash scripts/test.sh"
+  },
+  "devDependencies": {
+    "@elastic/synthetics": "../../",
```
vigneshshanmugam, line 11: seems wrong, needs the file protocol.

__tests__/e2e/package.json
```diff
@@ -0,0 +1,14 @@
...
+    "wait-on": "^5.0.1"
```
vigneshshanmugam, line 12: unused.

vigneshshanmugam (commented):
Its in a good shape, personally i would move this shell stuff to JS, but we can keep it for now and move to JS at some point to customize it. 

Have not tested it, but will take it for a spin sometime this week.

__tests__/e2e/package.json
shahzad31: we tried file protocol, i think it didn't work.

__tests__/e2e/docker-compose.yml
```diff
@@ -0,0 +1,68 @@
+version: '3.7'
+services:
+  es01:
```
jahtalab, line 3: Is there a reason to name these services with a number? I don't think we expect to add more in the future. Simply using `elasticsearch` is more readable, the same with kibana and synthetic

__tests__/e2e/docker-compose.yml
```diff
@@ -0,0 +1,68 @@
...
+      - bootstrap.memory_lock=true
+      - "ES_JAVA_OPTS=-Xms512m -Xmx512m"
+    ulimits:
+      memlock:
+        soft: -1
+        hard: -1
+    volumes:
+      - data01:/usr/share/elasticsearch/data
+    ports:
+      - 9200:9200
```
jahtalab, line 23: use `127.0.0.1:9200:9000` to avoid exposing externally.

__tests__/e2e/docker-compose.yml
```diff
@@ -0,0 +1,68 @@
...
+      - cluster.name=es-docker-cluster
+      - cluster.initial_master_nodes=es01
+      - bootstrap.memory_lock=true
+      - "ES_JAVA_OPTS=-Xms512m -Xmx512m"
+    ulimits:
+      memlock:
+        soft: -1
+        hard: -1
+    volumes:
+      - data01:/usr/share/elasticsearch/data
```
jahtalab, line 21: IMHO the data volume is unnecessary, since the data in the container is temporary anyway.

__tests__/e2e/docker-compose.yml
```diff
@@ -0,0 +1,68 @@
...
+    ports:
+      - 9200:9200
+    networks:
+      - elastic
```
jahtalab, line 25: I think the network is also unnecessary, since the default networks works fine for testing purposes.

__tests__/e2e/docker-compose.yml
```diff
@@ -0,0 +1,68 @@
...
+  kib01:
+    image: docker.elastic.co/kibana/kibana:${STACK_VERSION}
+    container_name: kib01
+    healthcheck:
+      test: ["CMD", "curl", "-I", "http://localhost:5601/status"]
+      interval: 30s
+      retries: 10
+      start_period: 30s
+    ports:
+      - 5601:5601
```
jahtalab, line 35: The same thing here with mapping to `127.0.0.1`

jahtalab (commented):
Thanks @shahzad31 , I left some comments. 

Also would you please have a look at the merge check list, it is failing at the moment?

__tests__/e2e/package.json
vigneshshanmugam: Interesting, I just tested it and it works for me. file:../../ linking to the repo's `package.json` directory. Can you try once again?

vigneshshanmugam (approved):

PR #119, opened 2020-10-30
fix: use millisecond for timestamps

jahtalab:
@andrewvc , This is only a draft PR to start using millis for timestamp, let's discuss if this is the direction we want to go.

Fixes #119


jahtalab:
Closing this one as we decided not change the unit for now.

PR #121, opened 2020-11-02
fix: use filmstrips option for capturing trace

### LINKED ISSUE #143: Feature flag to control filmstrip image capture
Until filmstrips are implemented (https://github.com/elastic/kibana/issues/83816), we want to change the default behaviour of the Synthetic Agent to stop capturing filmstrip images, to avoid any impact of them, or incorrect timestamps.  Similarly, we want the ability to do filmstrip capture testing, so will be able to switch them on with a feature flag.

**Acceptance criteria**
- [ ] By default, the Synthetics Agent will not capture filmstrip images
- [ ] There is a configuration option that, when enabled, will make the Synthetics Agent capture filmstrip images (in the same way that it does currently)

vigneshshanmugam:
+ fix #143 
+ Use `filmstrips` view for capturing the screenshot trace events instead of using the `screenshot` event, This should speed up the run from heartbeat side.

### DIFF
--- a/__tests__/index.test.ts
+++ b/__tests__/index.test.ts
@@ -62,6 +62,7 @@ describe('run', () => {
       environment: 'debug',
       headless: true,
       screenshots: true,
+      filmstrips: false,
       dryRun: true,
       journeyName: 'There and Back Again',
       network: true,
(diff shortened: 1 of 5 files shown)


jahtalab (approved):
LGTM 👍

PR #125, opened 2020-11-03
fix: generate timestamps when events are reported

### LINKED ISSUE #118: Journey start/end timestamps are incorrect
It looks like we're setting the start/end timestamps for journeys in the JSON to the same value from the journey context. This makes calculating journey duration impossible. This can be confirmed with the following command

```bash
andrewvc@x1e:~/p/s/e/todos|example-reliability⚡*?
➤ npx @elastic/synthetics . --json | jq '{t: .type, ts: .["@timestamp"]} | select(.t | contains("journey"))' -c
{"t":"journey/start","ts":1604025562618983}
{"t":"journey/end","ts":1604025562618983}
{"t":"journey/start","ts":1604025563206505.2}
{"t":"journey/end","ts":1604025563206505.2}
{"t":"journey/start","ts":1604025563766296.5}
{"t":"journey/end","ts":1604025563766296.5}
{"t":"journey/start","ts":1604025564166952}
{"t":"journey/end","ts":1604025564166952}
```

vigneshshanmugam:
+ fix #118 
+ Timestamps are calculated for every event when the JSON is written to the FD for most of the events as we want the intermediately steps like `network`, `screenshot`, `filmstrips` timestamps to fall under their respective steps timestamps and does not extend beyond that. 

+ Way to test this works
```
node dist/cli.js examples/todos/basics.journey.ts -j --metrics --screenshots | jq '{t: .type, ts: .["@timestamp"]}'
```

We would still use the `start`  and `end` timings for other calculations which is based on the monotonic clock based on the devtools protocol.

### DIFF
--- a/__tests__/core/runner.test.ts
+++ b/__tests__/core/runner.test.ts
@@ -84,14 +84,12 @@ describe('runner', () => {
       expect(event).toMatchObject({
         journey,
         step,
-        timestamp: expect.any(Number),
       });
     });
     runner.on('step:end', event => {
       expect(event).toMatchObject({
         journey,
         step,
-        timestamp: expect.any(Number),
         status: 'succeeded',
         url: 'about:blank',
         start: expect.any(Number),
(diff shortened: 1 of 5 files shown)


andrewvc (approved):
LGTM

Issue #123, opened 2020-11-03
Rename @elastic/synthetics bin to elastic-synthetics


andrewvc:
We should rename installed binary from synthetics -> elastic-synthetics to resolve issues with `npx` invocation and provide an unambiguous name for the executable.

We will need to quickly follow-up in heartbeat before this is merged to un-break the docker image.

PR #126, opened 2020-11-04
fix: handle symlinks and make examples work

### LINKED ISSUE #124: Switch examples to use `../../` for @elastic/synthetics lib
This will let us stop using `npm link` and get more consistent package behavior

vigneshshanmugam:
+ fix #124 #123 
This PR fixes multiple issues 
+ Handle symlinks and ignore them while running suites as it would crash the process by recursively searching for the same files
+ Ignore `node_modules` by default when running journey
+ Always install the latest version of synthetics package on the docker image and invoke from heartbeat side using `elastic-synthetics` executable without using the `npx` command as it would resolve the problem with local/global package collision
+ TS files are only compiled on the fly without type checking which is slower and not required for journey files from the users.

## Steps to run
+ `rm -rf node_modules`
+ `npm unlink` if you have linking before
+ `npm install` on the root folder

### To run examples locally, 
+ `cd examples/todos/ && npm install` 
+ `node dist/cli.js examples/todos` in the root or you can do `npm link` and run `elastic-synthetics examples/todos`

### For running inside heartbeat
**Note: heartbeat has to be changed to invoke the CLI through `elastic-synthetics`, but for testing purpose we can do this**
+ `cd examples/docker`
+ run `./bash-build-local.sh` which should give bash environment
+ `npm link` as we want the image to know about our unpublished new `elastic-synthetics` executable
+ `elastic-synthetics examples/todos`

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -19,9 +19,6 @@ RUN  cd /usr/share/heartbeat/.node \\
       && mkdir node \\
       && curl https://nodejs.org/dist/v12.18.4/node-v12.18.4-linux-x64.tar.xz | tar -xJ --strip 1 -C node
 ENV PATH="/usr/share/heartbeat/.node/node/bin:$PATH"
-# Install playwright first since it speeds up the install of elastic-synthetics*.tgz, since it doesn't need to re-download the
-# browsers every time the code there changes
-RUN npm i -g playwright-chromium
-COPY elastic-synthetics-*.tgz /opt/elastic-synthetics.tgz
-RUN npm install -g /opt/elastic-synthetics.tgz
-ENV HEARTBEAT_SYNTHETICS_TGZ=/opt/elastic-synthetics.tgz
+# Install the latest version of @elastic/synthetics, so heartbeat can
+# call the executable directly
+RUN npm i -g @elastic/synthetics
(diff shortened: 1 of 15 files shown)


Dockerfile
```diff
@@ -19,9 +19,6 @@ RUN  cd /usr/share/heartbeat/.node \\
...
 ENV PATH="/usr/share/heartbeat/.node/node/bin:$PATH"
-# Install playwright first since it speeds up the install of elastic-synthetics*.tgz, since it doesn't need to re-download the
-# browsers every time the code there changes
-RUN npm i -g playwright-chromium
-COPY elastic-synthetics-*.tgz /opt/elastic-synthetics.tgz
-RUN npm install -g /opt/elastic-synthetics.tgz
-ENV HEARTBEAT_SYNTHETICS_TGZ=/opt/elastic-synthetics.tgz
+# Install the latest version of @elastic/synthetics, so heartbeat can
+# call the executable directly
+RUN npm i -g @elastic/synthetics
```
vigneshshanmugam, line 24: Once we publish the new version, this should make `elastic-synthetics` in the docker image work out of the box.

examples/docker/bash.sh
```diff
@@ -14,6 +14,6 @@ docker run \
   --net=host \
   --security-opt seccomp=seccomp_profile.json \
   --volume="$(pwd)/heartbeat.docker.yml:/usr/share/heartbeat/heartbeat.yml:ro" \
-  --volume="$(pwd)/../:/opt/examples:rw" \
+  --volume="$(pwd)/../../:/opt/elastic-synthetics:rw" \
```
vigneshshanmugam, line 17: copying the whole project to make file links work in examples.

andrewvc (approved):
LGTM

PR #127, opened 2020-11-05
Install alpha by default

andrewvc:
This fixes the issue with broken NPM tags, where we'd publish the oldest alpha as latest.


vigneshshanmugam (approved):
Looks good

PR #129, opened 2020-11-09
Report URL on errors

andrewvc:
Prior to this patch we'd only report the URL in JSON output when the step passed.


vigneshshanmugam (approved):
Good catch

PR #132, opened 2020-11-09
chore: publish under two names

vigneshshanmugam:
+ this should make the `npx @elastic/synthetics` work. We have decided to keep the executable available under two names. We will revisit during the beta phase


andrewvc (approved):
LGTM

PR #135, opened 2020-11-11
Use tini init daemon (correctly) in Docker (#134)

andrewvc:
Without this chrome zombies sometimes appear. We should use an init daemon since we spawn subprocesses.

This is a redo of #134 which was reverted due to a bug. This should be tested with `./run-build-local.sh` and .`./bash-build-local.sh` (you can run `ps` in bash to check for `tini`)


vigneshshanmugam (approved):
Tested locally, works as expected.

PR #145, opened 2020-11-23
chore: update playwright to 1.6.2

vigneshshanmugam:
+ Allows running playwright on macOS Big Sur. Fails to launch browsers on older versions.


paulb-elastic (approved):
LGTM

PR #146, opened 2020-11-23
feat: add browser service to launch browsers

jahtalab:
This PR adds a browser service that can be used to launch browser on demand. It also adds `--ws-endpoint` option for the cli.

- To run the browser service simply run the JS file: `node dist/browser-service.js`
- To point the tests to use the browser service: `node dist/cli.js examples/todos/ --ws-endpoint=ws://localhost:9322`

### DIFF
--- a/src/browser-service.ts
+++ b/src/browser-service.ts
@@ -23,6 +23,7 @@
  *
  */
 
+import { URL } from 'url';
 import { createServer } from 'http';
 import { createProxyServer } from 'http-proxy';
 import { chromium } from 'playwright-chromium';
@@ -33,12 +34,10 @@ const proxyServer = createServer();
 proxyServer.on('upgrade', async function (req, socket, head) {
   const browserServer = await chromium.launchServer({ headless: true });
   const wsEndpoint = browserServer.wsEndpoint();
-  const parts = wsEndpoint.split('/');
-  const guid = parts.pop();
-  req.url = `/${guid}`;
-  const target = parts.join('/');
+  const { pathname, origin } = new URL(wsEndpoint);
+  req.url = pathname;
   console.log(`New browser: ${wsEndpoint}`);
-  proxy.ws(req, socket, head, { target });
+  proxy.ws(req, socket, head, { target: origin });
   const closeBrowser = async () => {
     await browserServer.close();
     console.log(`Socket closed: ${wsEndpoint}`);
--- a/src/browser-service.ts
+++ b/src/browser-service.ts
@@ -23,14 +23,12 @@
  *
  */
 
-/* eslint-disable @typescript-eslint/no-var-requires */
+import { createServer } from 'http';
+import { createProxyServer } from 'http-proxy';
+import { chromium } from 'playwright-chromium';
 
-const httpProxy = require('http-proxy');
-const http = require('http');
-const { chromium } = require('playwright-chromium');
-
-const proxy = httpProxy.createProxyServer();
-const proxyServer = http.createServer();
+const proxy = createProxyServer();
+const proxyServer = createServer();
 
 proxyServer.on('upgrade', async function (req, socket, head) {
   const browserServer = await chromium.launchServer({ headless: true });
@@ -41,10 +39,12 @@ proxyServer.on('upgrade', async function (req, socket, head) {
   const target = parts.join('/');
   console.log(`New browser: ${wsEndpoint}`);
   proxy.ws(req, socket, head, { target });
-  socket.on('close', async () => {
+  const closeBrowser = async () => {
     await browserServer.close();
     console.log(`Socket closed: ${wsEndpoint}`);
-  });
+  };
+  socket.on('close', closeBrowser);
+  socket.on('error', closeBrowser);
 });
 
 const port = 9322;
--- a/src/browser-service.ts
+++ b/src/browser-service.ts
@@ -1,0 +1,53 @@
+/**
+ * MIT License
+ *
+ * Copyright (c) 2020-present, Elastic NV
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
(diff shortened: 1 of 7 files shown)


vigneshshanmugam (commented):
Thanks @jahtalab , I have not tested them in local, But would like to give some feedback so its easier to understand the picture clearly
+ Can you improve the PR description and add how would you test this service locally by adding the steps you did while developing the PR. 
+ Could you tests for the browser service by using our run API with the new `wsEndpoint` option?

jahtalab:
Thanks @vigneshshanmugam , I've updated the description, but to summarise, simply running the js file and pointing the tests to endpoint should work, everything else is exactly the same.

I was planning to add E2E tests after it's PR is merged, but I think some unit tests would be useful as well 👍

jahtalab:
@vigneshshanmugam , On second thought, I think the browser service test doesn't fit in our unit tests. The E2E tests are a better match, Let's add those once the E2E PR is merged.

jahtalab:
Summary of conversation with @vigneshshanmugam:
- We discussed the options for testing, E2E vs unit tests.
- The trade-off of starting the browser-service for every unit test run
- We can potentially have a global option to run all unit tests against the browser service

Considering that the browser-service is in the POC phase, I suggest to add E2E tests for now (in separate PR) and look into the unit tests once we actually decide on the final approach.

@paulb-elastic , @andrewvc what are your thoughts on this?

src/core/gatherer.ts
```diff
@@ -46,9 +46,18 @@ export type Driver = {
...
  */
 export class Gatherer {
-  static async setupDriver(headless?: boolean): Promise<Driver> {
-    log('Gatherer: launching chrome');
-    const browser = await chromium.launch({ headless });
+  static async setupDriver(
+    headless?: boolean,
+    wsEndpoint?: string
+  ): Promise<Driver> {
+    let browser;
```
vigneshshanmugam, line 53: need to add types for the browser `ChromiumBrowser`

src/browser-service.ts
```diff
@@ -0,0 +1,53 @@
...
+
+const httpProxy = require('http-proxy');
+const http = require('http');
+const { chromium } = require('playwright-chromium');
+
+const proxy = httpProxy.createProxyServer();
+const proxyServer = http.createServer();
+
+proxyServer.on('upgrade', async function (req, socket, head) {
+  const browserServer = await chromium.launchServer({ headless: true });
```
vigneshshanmugam, line 36: Needs to be handled in different way as we are assuming `headless` service for now.

src/browser-service.ts
```diff
@@ -0,0 +1,53 @@
...
+  const wsEndpoint = browserServer.wsEndpoint();
+  const parts = wsEndpoint.split('/');
```
vigneshshanmugam, line 38: better to use `url.parse` and pass the target host and path. Any change in playwright would break this assumption otherwise.

src/browser-service.ts
```diff
@@ -0,0 +1,53 @@
...
+  const guid = parts.pop();
+  req.url = `/${guid}`;
+  const target = parts.join('/');
+  console.log(`New browser: ${wsEndpoint}`);
+  proxy.ws(req, socket, head, { target });
```
vigneshshanmugam, line 43: We may have to listen for error events and handle them correctly https://github.com/http-party/node-http-proxy#listening-for-proxy-events

src/browser-service.ts
```diff
@@ -0,0 +1,53 @@
...
+  socket.on('close', async () => {
```
vigneshshanmugam, line 44: `close` wont be called for `error` events in Node.js, lets call the same callback on both events.

src/browser-service.ts
```diff
@@ -0,0 +1,53 @@
...
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ *
+ */
+
+/* eslint-disable @typescript-eslint/no-var-requires */
+
+const httpProxy = require('http-proxy');
```
vigneshshanmugam, line 28: move to `import` statements.

src/browser-service.ts
```diff
@@ -0,0 +1,53 @@
+/**
+ * MIT License
+ *
```
vigneshshanmugam, line 3: can you move this file to `src/core`

vigneshshanmugam (changes requested):
@jahtalab  Thanks, Did some tests in local and examples works smoothly, However i have not tested all the edge cases like 
+ If browser handles are closed correctly when there is error on the any step/journey
+ handling error when proxy server is closed in the middle ( triggers unhandle rejection scenario and might need in our `browser-service.ts` class)

I am in favour of having at-least few unit tests for this `browser-service` file and add more tests along with our API if we want to go forward with this approach.

src/browser-service.ts
jahtalab: @vigneshshanmugam , would you please elaborate on this? needs to be different in what way?

src/browser-service.ts
jahtalab: I don't think the browser service belongs in the core, since it's not a component that we use directly but rather a service that is executed on it's own. In that sense it's more similar to the cli.

src/browser-service.ts
jahtalab: The proxy errors are passed to the downstream, the error event is useful only if we want to return a custom error instead.

src/browser-service.ts
vigneshshanmugam: This was more of a question, do we want to launch the service in headless mode for all cases?

paulb-elastic:
@jahtalab I agree and think that we can hold off adding tests right now until we decide to move ahead with this approach.

src/browser-service.ts
jahtalab: Primarily we want to run this service with `headless: true` flag, but for testing and local development it might be useful to have the option, but let's add these after the POC phase.

jahtalab:
Thanks @paulb-elastic , I've added a task for adding the tests once we have decided on the approach in the original issue.

src/browser-service.ts
vigneshshanmugam: I am not sure if we want to do `headless: true`, Did we already decide on this, Not able to find any relevant discussion? Running in headless mode would not get some of the performance metrics. I am okay leaving it for now, but its not only for testing/local dev. Ideally we should support both modes and by default should be run on non-headless mode as it allows for capturing more perf metrics.

src/browser-service.ts
```diff
@@ -0,0 +1,52 @@
...
+const proxy = createProxyServer();
+const proxyServer = createServer();
+
+proxyServer.on('upgrade', async function (req, socket, head) {
+  const browserServer = await chromium.launchServer({ headless: true });
+  const wsEndpoint = browserServer.wsEndpoint();
+  const { pathname, origin } = new URL(wsEndpoint);
+  req.url = pathname;
+  console.log(`New browser: ${wsEndpoint}`);
+  proxy.ws(req, socket, head, { target: origin });
```
vigneshshanmugam, line 40: why not pass the full url object? Thought proxy lib supports that option.

vigneshshanmugam (approved):

src/browser-service.ts
jahtalab: We haven't decided it, but It doesn't make a difference for the POC and adding the option is easy to do later on.

src/browser-service.ts
jahtalab: I think I tried that but didn't work, but we'll have another look for the final version.

vigneshshanmugam: Yeah ++

PR #147, opened 2020-11-24
fix: sync browser version in docker and examples

### LINKED ISSUE #142: Run Playwright post-install script to fix chrome binary sometimes unavailable
It seems that sometimes the synthetics package can get into a state where our code runs, but the version of the playwright library used references a chrome version that is not installed. @paulb-elastic got his system in such a state once, but we couldn't repro it.

My hunch is this can happen on only some systems after playwright releases a new version. I'm somewhat unsure beyond that, since my system can't repro this.

My hunch is that it would be a good thing to run the playwright post-install script on each run to ensure that the chrome binary is present, and attempt download if not. I can't think of a reason not to try downloading it vs. a straight-up failure.

This only seems to happen with `npx` running a globally installed package. For a local project this circumstance can be fixed with `npm i && npm install playwright`. See https://github.com/microsoft/playwright/issues/3712 for more context on that situation

vigneshshanmugam:
+ fix #142 
+ Forcefully reinstalls our `@elastic/synthetics` package by ignoring the cache incase any versions have changed. We have not used the `docker build --no-cache` as it will also clear out other dependency caches
+ By having the latest version on both official docker image and examples pointing to the latest published version. We synchronise the playwright versions inside the docker image and do not have a broken state. 
+ We can also remove the intermin solution that is put inside the heartbeat code https://github.com/elastic/beats/blob/heartbeat-synthetics/x-pack/heartbeat/monitors/browser/synthexec/synthexec.go#L57 as the problem is not related to playwright being available instead a broken state.  

Steps to test this flow. 

```sh
// remove the previous docker image if you want to test in a clean environment
docker rmi heartbeat-synthetics-local
npm run docker 
cd examples/docker
./run-build-local.sh -E output.console={}
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -26,6 +26,6 @@ RUN  cd /usr/share/heartbeat/.node \\
       && mkdir node \\
       && curl https://nodejs.org/dist/v12.18.4/node-v12.18.4-linux-x64.tar.xz | tar -xJ --strip 1 -C node
 ENV PATH="/usr/share/heartbeat/.node/node/bin:$PATH"
-# Install the latest version of @elastic/synthetics, so heartbeat can
-# call the executable directly
-RUN npm i -g @elastic/synthetics@alpha
+# Install the latest version of @elastic/synthetics forcefully ignoring the previously
+# cached node_modules, hearbeat then calls the global executable to run test suites
+RUN npm i -g -f @elastic/synthetics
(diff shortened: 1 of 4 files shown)


paulb-elastic (approved):
I've run this locally and it works, I'm getting the JSON output from the synthetic monitors that are running.  I had to authorise with the docker registry before this would work.

vigneshshanmugam:
Will revisit later with better alternatives if we decide to do docker cache cleanup at some point for building new images.

PR #150, opened 2020-11-26
chore: disallow auto update of chrome browser

vigneshshanmugam:
+ This is a follow up of #142 and allows us to control the update of playwright effectively browsers to tie it with the proper version. Having `^` or `~` would allow minor/patch versions to be installed on user's machine/test environments if lockfiles are not taken in to account.

### DIFF
--- a/package.json
+++ b/package.json
@@ -39,7 +39,7 @@
   "dependencies": {
     "commander": "^6.0.0",
     "kleur": "^4.1.1",
-    "playwright-chromium": "^1.6.2",
+    "playwright-chromium": "=1.6.2",
     "snakecase-keys": "^3.2.0",
     "sonic-boom": "^1.1.0",
     "source-map-support": "^0.5.19",


jahtalab (commented):
Thanks @vigneshshanmugam, Can we at least allow the patch versions? Is there a precedent for patch versions making an issue?

vigneshshanmugam:
yeah patch versions also updates the browser https://github.com/microsoft/playwright/releases which would be an issue for us as it would trigger removing the old browser version in favor of new one. So, we need to stick with exact match.

package.json
```diff
@@ -39,7 +39,7 @@
   "dependencies": {
     "commander": "^6.0.0",
     "kleur": "^4.1.1",
-    "playwright-chromium": "^1.6.2",
+    "playwright-chromium": "=1.6.2",
```
jahtalab, line 42: Just specifying the exact version would pin the version. I wasn't sure about the `=` syntax since it's not mentioned in the npm docs.

```suggestion
    "playwright-chromium": "1.6.2",
```

jahtalab (approved):
Sounds good! I think we should separate the update cycle of the browser from the update of playwright (we can do that as part of the packaging ticket).

package.json
vigneshshanmugam: Its part of semver https://github.com/npm/node-semver

Issue #152, opened 2020-11-30
capture more network information to improve waterfall


vigneshshanmugam:
Couple of improvements that we can do to improve the step level waterfall in the UI
+ [x] Move the overall timing calculation logic in the UI to the agent to improve the performance #168 
+ [x] Use `requestSentime` and `responseReceivedTime` in favor of `start/end` time and use the monotonic clock in start
and end time
+ [x] Capture the step index info for all the netinfo events to make querying easier 
+ [x] Fix the timestamp to get the value of when the event got generated vs when the event is written to stdout which is after journey end.
~+ [ ] Enhance the additional headers information by listening to `requestWillBeSentExtraInfo` and `responseReceivedExtraInfo` event~
~+ [ ] Figure out if an entry is from the cache by using `requestServedFromCache` event.~

paulb-elastic:
Removing `capture web vitals and other RUM metrics - #120` from the list of requirements as we're tracking #120 separately.

andrewvc:
Since we don't actually need the last two items for our UI right now I've crossed this out, and am closing the issue.

PR #154, opened 2020-12-01
fix: associate network info with each step

vigneshshanmugam:
+ part of #152 
+ Associate the network information with each step as its required for the waterfall chart creation
+ fix the timestamp information as it was pointing to the journey end timestamp as opposed to the step start/end timestamp
+ Move `start/end` in favor of `requestSentTime/loadEndTime` as its when events are written and not the actual duration. 
/cc @Kerry350

### DIFF
--- a/src/common_types.ts
+++ b/src/common_types.ts
@@ -24,6 +24,7 @@
  */
 
 import { Protocol } from 'playwright-chromium/types/protocol';
+import { Step } from './dsl';
 
 export type VoidCallback = () => void;
 
@@ -35,6 +36,11 @@ export type FilmStrip = {
   ts: number;
 };
 
+export type DefaultPluginOutput = {
+  step?: Partial<Step>;
+  timestamp: number;
+};
+
 export type NetworkInfo = {
   url: string;
   method: string;
@@ -42,7 +48,7 @@ export type NetworkInfo = {
   request: Protocol.Network.Request;
   response: Protocol.Network.Response;
   isNavigationRequest: boolean;
-  start: number;
-  end: number;
+  requestSentTime: number;
+  loadEndTime: number;
   status: number;
-};
+} & DefaultPluginOutput;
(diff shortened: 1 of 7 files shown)


src/plugins/plugin-manager.ts
```diff
@@ -78,6 +79,11 @@ export class PluginManager {
     return this.plugins.get(Type.name) as T;
   }
 
+  onStep(step: Step) {
+    this.get(BrowserConsole) && (this.get(BrowserConsole)._currentStep = step);
```
vigneshshanmugam, line 83: Will replace this with EventEmitter at some point if we require more integration between runner and plugins.

src/plugins/network.ts
```diff
@@ -50,14 +53,16 @@ export class NetworkManager {
...
 
     this.waterfallMap.set(requestId, {
+      step: this._currentStep,
+      timestamp: getTimestamp(),
       url,
       request,
       type,
       method,
-      start: timestamp,
+      requestSentTime: timestamp,
```
vigneshshanmugam, line 62: This change was done after some investigation with the chrome network events and how we will use it in the waterfall. copy pasting from internal gist

So here is the gist that i found and we need to change

* we will use requestTime  for the waterfall stuff
* start/end represents the timestamp of when request is sent and when request completed, so we can calculate the queue time as follows
 ```js
const queuedTime = payload.response.timing.request_time - payload.response.start
if (queuedTime > 0) {
  // convert in ms and use it in waterfall
}
```
* we discard end timing for now as its no useful. Also i think we can rename it to something like requestStartTime  and responseEndTime to make it easier

File: src/common_types.ts [hunk 3]
```diff
@@ -35,14 +36,19 @@ export type FilmStrip = {
...
   url: string;
   method: string;
   type: string;
   request: Protocol.Network.Request;
   response: Protocol.Network.Response;
   isNavigationRequest: boolean;
-  start: number;
-  end: number;
+  requestSentTime: number;
+  loadEndTime: number;
```
Kerry350, line 52: Much clearer 👍

Kerry350 (approved):
LGTM 👍 

Thanks for making things easier here 🙏

Issue #120, opened 2020-10-30
Capture relevant perf metrics to close the gap with RUM agent


vigneshshanmugam:
This has come up multiple times in the slack channel. We have to measure and report all the performance metrics for each step session. 

#### Metrics to meausre
   - [x] First Contentful Paint (FCP)
   - [x] Core Web Vital: Largest Contentful Paint (LCP)
   - [x] Core Web Vital: Cumulative Layout Shift (CLS)
   - [x] onLoad (document complete)
   - [x] DOM Content Loaded
   - [x] User timing marks & measures

**Note: The above marked items are fixed by PR #194**

More details - https://github.com/elastic/uptime/issues/250

paulb-elastic:
@vigneshshanmugam can you convert the above list into checkboxes and mark off the ones that have been done, to make it easier to identify what status this is in, thanks.

paulb-elastic:
I have removed these from the main issue as they can all be calculated.  If necessary, a separate issue will be created if these need to be optimised when we come to use them:
* Number of Requests
* Bytes Over the Wire
* Bytes Uncompressed
* Connected Domains

paulb-elastic:
Split off some of the metrics into separate issue https://github.com/elastic/synthetics/issues/283

PR #155, opened 2020-12-01
chore: improve tracer time by consuming as events

vigneshshanmugam:
+ Changes the Tracing.start protocol command to use the `ReportEvents` instead of `ReturnAsStream`. This makes the trace gathering much faster, as observed in some automated run timings:

### Average timings - With filmstrips alone
Before - 3 seconds 
After -  2.5 seconds 

This would be useful as we add more categories to trace when we start doing this #120

### DIFF
--- a/src/plugins/tracing.ts
+++ b/src/plugins/tracing.ts
@@ -43,8 +43,7 @@ export function filterFilmstrips(
     return (
       name === 'Screenshot' &&
       cat === 'disabled-by-default-devtools.screenshot' &&
-      args &&
-      args.snapshot
+      args?.snapshot
     );
   });
 
@@ -68,6 +67,10 @@ export class Tracing {
       'disabled-by-default-devtools.screenshot',
     ];
     await client.send('Tracing.start', {
+      /**
+       * Using `ReportEvents` makes gathering trace events
+       * much faster as opposed to using `ReturnAsStream` mode
+       */
       transferMode: 'ReportEvents',
       traceConfig: {
         includedCategories,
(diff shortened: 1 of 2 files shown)


src/plugins/tracing.ts
```diff
@@ -80,26 +61,34 @@ export function filterFilmstrips(buffer: Buffer): Array<FilmStrip> {
...
-
+    const includedCategories = [
+      // exclude all default categories
+      '-*',
+      // capture screenshots
+      'disabled-by-default-devtools.screenshot',
+    ];
     await client.send('Tracing.start', {
-      transferMode: 'ReturnAsStream',
+      transferMode: 'ReportEvents',
```
jahtalab, line 71: The transferMode `ReportEvents` is the default mode, so we can remove this.
also, might be worth adding a comment in the code to explain choosing the default mode over `ReturnAsStream`

jahtalab (commented):
Thanks @vigneshshanmugam , generally looks good, added a few comments.

src/plugins/tracing.ts
vigneshshanmugam: Yeah its the default, i wanted to keep it as they are deprecating some of the fields and might change the default anytime. I will add a comment though about the perf implication.

jahtalab: Out of interest, do you have a link for the that announcement?

jahtalab (approved):

src/plugins/tracing.ts
vigneshshanmugam: No i meant the deprecation notices in the latest tip of tree in CDP https://chromedevtools.github.io/devtools-protocol/tot/Tracing/#method-start. None of them are at the stable RC at the moment. Anything could change, so this was just us being defensive, nothing announced yet.

Issue #156, opened 2020-12-01
Docker example doesn't work well with uids != 1000


andrewvc:
In elastic/synthetics#153  @paulb-elastic discovered that the docs only work in the rather common scenario where the your user on the host machine is 1000 which matches the heartbeat user in the container. The issue, essentially, is that the `heartbeat` user in docker has a uid of `1000`, and if this doesn't match up with the uids of the shared mounts you'll get perms issues.

Possible solutions:

1. Document this, recommend users users change their local file perms
2. Start using [fixuid](https://github.com/boxboat/fixuid) in the container
3. Something else???

paulb-elastic:
During tech sync we have discussed the option of using a copy on start up, as opposed to a shared volume.

In this way, the host folder would be copied into the container at start up (which only needs read permissions) and then the runtime Synthetics Agent uses that folder inside the container (not as a shared volume).  It will of course be ephemeral and do the same during the next launch.

The [Remote Zip URL support](https://github.com/elastic/uptime/issues/271) feature will still provide a mechanism for the suites to be updated without needing a shared volume.

Similarly, [hot reloading](https://github.com/elastic/beats/issues/22511) would still be supported in that it would just periodically copy the host folder(s) to the running container at the given interval.

PR #160, opened 2020-12-03
Add offline docker example for air gapped networks

andrewvc:
This adds an example of how to use heartbeat / synthetics in an air gapped network, where our calls to NPM won't work. Users will have to pre-bake a container on a machine with internet access and ship that behind their firewall.


paulb-elastic (commented):
@andrewvc I tried running this (`build-offline-dockerfile.sh`) on a machine with no images, for it to build one locally, but get the following error when it gets to step 6:
```
Step 6/6 : RUN cd /opt/synthetics-suites/todos && npm install && npm install playwright-chromium
 ---> Running in 5561c4d6f064
npm WARN checkPermissions Missing write access to /opt/synthetics-suites/todos
npm ERR! code EACCES
npm ERR! syscall access
npm ERR! path /opt/synthetics-suites/todos
npm ERR! errno -13
npm ERR! Error: EACCES: permission denied, access '/opt/synthetics-suites/todos'
npm ERR!  [Error: EACCES: permission denied, access '/opt/synthetics-suites/todos'] {
npm ERR!   errno: -13,
npm ERR!   code: 'EACCES',
npm ERR!   syscall: 'access',
npm ERR!   path: '/opt/synthetics-suites/todos'
npm ERR! }
npm ERR! 
npm ERR! The operation was rejected by your operating system.
npm ERR! It is likely you do not have the permissions to access this file as the current user
npm ERR! 
npm ERR! If you believe this might be a permissions issue, please double-check the
npm ERR! permissions of the file and its containing directories, or try running
npm ERR! the command again as root/Administrator.

npm ERR! A complete log of this run can be found in:
npm ERR!     /usr/share/heartbeat/.npm/_logs/2020-12-08T11_18_21_312Z-debug.log
```

This is on a VM where I can run the normal docker example, so I don't think is related to https://github.com/elastic/synthetics/issues/156

andrewvc:
@paulb-elastic this should be fixed in the latest commits, however, it won't work quite right until this PR is merged, because I made some changes to the base image, and the example references the published images. However, if you run the `examples/docker/run-build-local.sh` script, then use the commented out `FROM` line in the todos `Dockerfile`, you can simulate this experience, which should work.

examples/todos/run-offline-docker.sh
```diff
@@ -0,0 +1,2 @@
+#!/bin/sh
+../docker/run.sh my-custom-heartbeat
```
jahtalab, line 2: nit: It seems to me that this could be move to docs or to the readme

jahtalab (approved):
Thanks @andrewvc , LGTM 👍

examples/todos/Dockerfile
```diff
@@ -0,0 +1,30 @@
...
+
+# Copy our heartbeat config directly to the image.
+# This could be done as a shared mount instead, but if we're
+# baking an image anyway, this may be something that may be easier
+# to do in this manner.
+COPY heartbeat.docker.yml /usr/share/heartbeat/heartbeat.yml
+
+RUN mkdir -p $SUITES_DIR/todos
+# Copy your custom synthetics tests into a folder on the image
+COPY ./* $SUITES_DIR/todos/
```
jahtalab, line 25:
```suggestion
COPY . $SUITES_DIR/todos/
```

jahtalab (commented):
@andrewvc , I've built the image locally. The build is successful but the test suite fails, the reason is that when the suite directory is copied into the image it also flattens the file structure therefore it can't load the `index.html` at the specified location in the tests. I've made a suggestion that fixes this.

examples/todos/package.json
```diff
@@ -6,6 +6,7 @@
   "license": "MIT",
   "dependencies": {
     "@elastic/synthetics": "*",
+    "playwright-chromium": "^1.6.2",
```
vigneshshanmugam, line 9: This would straight away blow up and would not work. An easy way to test the breakage would be to change the

```
"@elastic/synthetics": "0.5.0" //does not depend on 1.6.2 chromium
"playwright-chromium": "^1.6.2"
```

The reason why this would break is because of the installation of different versions of chromium browser on global installation path. 

We would need to assume the image has already preinstalled node_modules directory with specified playwright version from installing `@elastic/synthetics` and remove `playwright-chromium` from deps. You could also do force install of `@elastic/synthetics` if you want to be absolutely sure about installing the right version and triggering pre install hook from playwright.

examples/todos/Dockerfile
```diff
@@ -0,0 +1,30 @@
...
+
+# Install NPM deps locally on this image
+# Please note that it's important to run both `npm install` AND `npm install playwright`
+# for more see this issue: https://github.com/microsoft/playwright/issues/3712
+RUN cd $SUITES_DIR/todos && npm install && npm install playwright-chromium
```
vigneshshanmugam, line 30: would remove the `playwright-chromium` from here and also from the deps. reason given below.

examples/todos/run-offline-docker.sh
andrewvc: ++ I'll remove this

Issue #136, opened 2020-11-11
Add option to disable chromium sandbox


andrewvc:
Disabling the chrome sandbox is useful for containers. I propose we add a `--no-sandbox` option to allow this.

PR #167, opened 2020-12-09
added --no-sandbox option

shahzad31:
Fixes: https://github.com/elastic/synthetics/issues/136
added --no-sandbox option into cli


andrewvc (approved):
Works perfectly (and I like the improved types). Confirmed that this does disable the sandbox:

image

PR #174, opened 2020-12-17
Update npm deps

andrewvc:
this updates ini among other things


examples/todos/package.json
```diff
@@ -6,6 +6,6 @@
   "license": "MIT",
   "dependencies": {
     "@elastic/synthetics": "*",
-    "playwright-core": "^1.5.2"
+    "playwright-core": "^1.7.0"
```
jahtalab, line 9: I think we should keep the examples version of playwright the same as the agents' version, which is currently at `1.6.2`
```suggestion
    "playwright-core": "=1.6.2"
```

PR #179, opened 2020-12-22
Fix Dockerfile by removing slashes

ruflin:
I cannot run `docker build .` as there are `\\` in the Dockerfile instead of `\`. Not sure if this is on purpose or not?


vigneshshanmugam (approved):
Probably a docker version thing, we haven't seen any error for so long. But PR looks good.

PR #181, opened 2021-01-06
feat: add seccomp profiler

jahtalab:
This is the first draft for generating seccomp profile from any runtime environment. This approach uses the syslog generated from seccomp violations and parses these logs to create a seccomp profile in yaml format. By default it runs the synthetic agent to generate the profile but it can be configured to work for other executables.

To generate a profile use  `docker-compose up seccomp-profiler` (in `seccomp` folder)

Remaning issues:
- The rsyslog service seems to be running a bit late which creates inconsistency when generating the profile
- We should use docker syslog driver


jahtalab:
@andrewvc , I've created this draft so we can discuss the approach in more detail. Please have a look and let me know what you think.

andrewvc (approved):
This approach seems reasonable for a start here. LGTM

PR #186, opened 2021-01-12
test: run unit tests against browser service

jahtalab:
This is part of browser service integration effort. This PR runs our tests against browser service to make sure the application logic works correctly. This effectively makes sure that the synthetics agent behaves the same in both running with or without browser service.


jest.config.js
```diff
@@ -30,5 +30,7 @@ module.exports = {
   coverageDirectory: 'coverage',
   coverageProvider: 'v8',
   testPathIgnorePatterns: ['dist'],
-  modulePathIgnorePatterns: ['/e2e/']
+  modulePathIgnorePatterns: ['/e2e/'],
+  globalSetup: './utils/jest-global-setup.ts',
```
vigneshshanmugam, line 34: nit: maybe we should move all these inside `__tests__`, WDYT?  Including server.ts and test related files.

utils/jest-global-setup.ts
```diff
@@ -0,0 +1,50 @@
...
+import { wsEndpoint } from './test-config';
+
+module.exports = async () => {
+  if (wsEndpoint) {
+    return new Promise((resolve, reject) => {
+      console.log(`\nRunning BrowserService ${wsEndpoint}`);
+      const process = spawn('node', ['./dist/browser-service.js']);
+      (global as any).__browserServiceProcess = process;
+      process.stdout.on('data', data => {
+        if (data.indexOf('Listening on port: 9322') >= 0) {
```
vigneshshanmugam, line 36: Instead of doing this, We should get rid of the console logs in here https://github.com/elastic/synthetics/blob/b55ab0a3812cdb301f3eef58e85ebf4718c56ccf/src/browser-service.ts#L39-L46

and move them as debug logs. We don't want to show those logs in production anyway.

utils/jest-global-setup.ts
```diff
@@ -0,0 +1,50 @@
...
+          console.log(`${data}`);
+          resolve(process);
+        }
+      });
+
+      process.stderr.on('data', data => {
+        const message = `BrowserService: ${data}`;
+        reject(message);
```
vigneshshanmugam, line 44: doing this would hang the tests and would make jest teardown not work correctly (browser service would not be closed). 

we can `resolve` always with process here and let tests take care of failing functionality/timeouts in service. If we do this, then 
```js
spawn('node', ['./dist/browser-service.js'], {stdio: 'inherit'});
```
should be enough.

vigneshshanmugam (commented):
Generally looks good, Added some comments on closing child process correctly. 

+ need to add `wsEndpoint` to the missed tests in `gatherer.test.ts` and `network.test.ts`

jest.config.js
jahtalab: If we move them to `__test__` we would need to ignore them similar to `/e2e/` tests. I don't have a strong preference here.

jahtalab: I've moved them to `__test__`

utils/jest-global-setup.ts
jahtalab: I think we can add a debug mode for these logs, I've added an item to the original issue for this.
We still need a way to confirm that browserService is up otherwise tests will be flaky.

utils/jest-global-setup.ts
jahtalab: I've added `process.kill()` if there's an error, I prefer this over waiting for timeouts since you get clear errors and it also fails fast instead of waiting for the timeout.

utils/jest-global-setup.ts
vigneshshanmugam: We already have a listening log here which should tell if its up https://github.com/elastic/synthetics/blob/b55ab0a3812cdb301f3eef58e85ebf4718c56ccf/src/browser-service.ts#L51

For other places, so you can just use our internal logger. https://github.com/elastic/synthetics/blob/b55ab0a3812cdb301f3eef58e85ebf4718c56ccf/src/core/logger.ts#L40

vigneshshanmugam:
@jahtalab Once you add the missing tests for `wsEndpoint` to the tests in `gatherer.test.ts` and `network.test.ts`, you can see the process hanging error that i mentioned in review.

jahtalab:
@vigneshshanmugam, the `wsEndpoint` is already added to `gatherer.test.ts` and `network.test.ts` and tests run successfully both on my machine as well as the CI. 

Do you see this issue locally on your machine? Also can you elaborate on the issue that you're experiencing?

vigneshshanmugam:
@jahtalab Its for these tests, I can reproduce the process not being killed issue - 

1. https://github.com/elastic/synthetics/blob/b55ab0a3812cdb301f3eef58e85ebf4718c56ccf/__tests__/core/gatherer.test.ts#L39-L49

2. https://github.com/elastic/synthetics/blob/b55ab0a3812cdb301f3eef58e85ebf4718c56ccf/__tests__/plugins/network.test.ts#L55-L73

You can add `wsEndpoint` to those tests and then `WSENDPOINT=ws://localhost:9322 npx jest __tests__/plugins/network.test.ts` should reproduce the process issue.

jahtalab:
@vigneshshanmugam , the problem with tests hanging is not related to the how the tests are running against browser service, but it's related to browser service not handling `SIGTERM`

We still need to handle process events properly in the browser service and as part of the [Browser service integration](https://github.com/elastic/uptime-dev/issues/38), I've added a task for this.

jahtalab:
jenkins retest this

vigneshshanmugam (approved):

PR #187, opened 2021-01-12
fix: calculate blocking time properly for missing timing data

### LINKED ISSUE #185: Unit tests for network/waterfall timing calculation
As we approach the beta, its good to add the tests for `network.ts` which includes the waterfall timing calculation logic and ensure we have all the fields. Also its a good way to have the tests before we migrate fields to ECS format.

vigneshshanmugam:
+ Blocking time was set to `Infinity` when timing data was unavailable. Found it while adding tests, I have added more descriptive tests for timing calculation. 
+ fix #185


andrewvc (approved):
LGTM

PR #192, opened 2021-01-20
feat: add init/dispose to BrowserService

jahtalab:
Part of https://github.com/elastic/uptime-dev/issues/38

- Refactor BrowserService into a class
- Add browser service test


src/run-browser-service.ts
```diff
@@ -0,0 +1,37 @@
...
+
+const browserService = new BrowserService();
+browserService.init();
+
+const exitFn = async () => {
+  await browserService.dispose();
+  process.exit(0);
+};
+
+process.on('SIGTERM', exitFn);
```
vigneshshanmugam, line 36: add `uncaughtexception` listener as well, just to make it resilient.

src/run-browser-service.ts
```diff
@@ -0,0 +1,37 @@
...
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ *
+ */
+
+import { BrowserService } from './core/browser-service';
+
+const browserService = new BrowserService();
+browserService.init();
```
vigneshshanmugam, line 29: this should be moved inside 
```
 if (require.main === module) {
    const browserService = new BrowserService();
    browserService.init();
  }
```

to make sure we are running this file only when invoked via node src/run-browser-service

__tests__/utils/jest-global-setup.ts
```diff
@@ -30,18 +30,21 @@ module.exports = async () => {
...
-      const process = spawn('node', ['./dist/browser-service.js']);
-      (global as any).__browserServiceProcess = process;
-      process.stdout.on('data', data => {
+      const browserServiceProcess = spawn(
+        'node',
+        ['./dist/run-browser-service.js'],
+        { env: { ...process.env, DEBUG: 'true' } }
+      );
+      (global as any).__browserServiceProcess = browserServiceProcess;
+      browserServiceProcess.stdout.on('data', data => {
```
vigneshshanmugam, line 39: Why don't we switch to `spawn` event and then resolve the spawned process. https://nodejs.org/api/child_process.html#child_process_event_spawn

__tests__/utils/jest-global-setup.ts
```diff
@@ -30,18 +30,21 @@ module.exports = async () => {
...
         if (data.indexOf('Listening on port: 9322') >= 0) {
-          console.log(`${data}`);
-          resolve(process);
+          resolve(browserServiceProcess);
         }
       });
 
-      process.stderr.on('data', data => {
+      browserServiceProcess.stderr.on('data', data => {
```
vigneshshanmugam, line 45: same here, listen for `exit`, `error`, close event and kill the process.

__tests__/core/browser-service.test.ts
```diff
@@ -0,0 +1,48 @@
...
+    browserService.init();
+  });
+
+  afterEach(async () => {
+    await browserService.dispose();
+  });
+
+  it('should create browser pages', async () => {
+    const browser = await chromium.connect({
+      wsEndpoint: 'ws://localhost:9323',
```
vigneshshanmugam, line 42: can we add a field to the browserservice class to extract the endpoint, would be good to have it there and use it everywhere for tests.

__tests__/core/browser-service.test.ts
```diff
@@ -0,0 +1,48 @@
...
+    });
+    const context = await browser.newContext();
+    const page = await context.newPage();
+    await page.goto('https://elastic.co');
```
vigneshshanmugam, line 46: close the browser process as well here.

src/run-browser-service.ts
jahtalab: I don't expect that we would require `run-browser-service.ts` if we do it's a bug, it's not meant to be imported. I see this similar to how the `cli.ts` works, it's expected to executed directly and not imported.

__tests__/utils/jest-global-setup.ts
jahtalab: The reason is that the `spawn` event does not guarantee that the browser service has started listening on the port and therefore it add a race condition to the tests.

__tests__/utils/jest-global-setup.ts
jahtalab: See the above comment.

__tests__/core/browser-service.test.ts
jahtalab: The browser service does not necessarily know on which interface it is accessible, e.g. if it's running inside docker, so I don't think having the wsEndpoint there is a good idea.

src/run-browser-service.ts
vigneshshanmugam: Its a good practise to wrap things inside this pattern to avoid the consequence of auto initialising browser service. CLI is different. However this is a not a must do.

__tests__/core/browser-service.test.ts
vigneshshanmugam: Probably i wasn't clear. I meant this
```
export class BrowserService {
  closeCallbacks = [];
  proxyServer = createServer();
  options: Record<string, unknown>;
  host: string;

  constructor(options = { port: 9322 }) {
    this.options = options;
    this.host = `ws:localhost:${this.options.port}`;
  }
}

```

We could use the `host` instead everywhere.

src/core/browser-service.ts
```diff
@@ -0,0 +1,77 @@
...
+    });
+
+    this.proxyServer.listen(this.options.port, () => {
+      log(`Listening on port: ${this.options.port}`);
+    });
+  }
+
+  async dispose() {
+    for (let i = 0; i < this.closeCallbacks.length; i++) {
+      const callback = this.closeCallbacks[i];
```
vigneshshanmugam, line 68: we have to find a better pattern for this. let's improve this later.

__tests__/core/browser-service.test.ts
jahtalab: Yeah, that's what I meant as well, the browser service isn't necessarily hosted on `localhost` therefore having the host on this class is not appropriate IMO.

src/core/browser-service.ts
jahtalab: Alternatively we can keep the browser references in the array but it's more or less similar to this. This gives us a bit more flexibility in terms of disposing of multiple resources in one function though.

src/run-browser-service.ts
jahtalab: Let's keep it as is then, I don't anticipate an issue here and it's an easy fix to do later on. There are plenty of libraries like us without this. FWIW I don't think this is a bad practice though, just unnecessary IMO.

vigneshshanmugam (approved):

PR #193, opened 2021-01-20
fix: update docker-compose to use heartbeat docker image

kuisathaverat:
The 7.11.0 heartbeat Docker image contains synthetics so we can use it.


mdelapenya (approved):

__tests__/e2e/docker-compose.yml
```diff
@@ -40,7 +40,7 @@ services:
     networks:
       - elastic
   synthetic:
-    image: docker.elastic.co/experimental/synthetics:${STACK_VERSION}-synthetics
+    image: docker.elastic.co/beats/heartbeat:${STACK_VERSION}
```
v1v, line 43: We don't need to support previous versions? Only 7.11/master/7.x? If so ok, otherwise maybe the docker-compse should support the previous behaviour?

kuisathaverat: synthetic is experimental (https://www.elastic.co/what-is/synthetic-monitoring) the first version of heartbeat with synthetic is 7.11, so we do not need to support older versions

cachedout (approved):

__tests__/e2e/docker-compose.yml
kuisathaverat, line 43:
```suggestion
    image: ${STACK_IMAGE:-docker.elastic.co/beats/heartbeat}:${STACK_VERSION}
```

vigneshshanmugam:
This PR https://github.com/elastic/synthetics/pull/196 will fix the test failure issue.

vigneshshanmugam:
@kuisathaverat Did you investigate already why the E2E fails with the new image?

kuisathaverat:
the heartbeat Docker image does not start with the following error, it looks like a configuration issue, also the e2e will test a Docker image already generated not the current commit, I will check if it is possible to test the current commit instead.

```
2021-04-21T11:45:45.570Z	ERROR	instance/beat.go:953	Exiting: could not create monitor: job err could not parse suite config: config 'source' must be specified for this monitor, if upgrading from a previous experimental version please see our new config docs accessing 'heartbeat.monitors.0' (source:'heartbeat.yml')
Exiting: could not create monitor: job err could not parse suite config: config 'source' must be specified for this monitor, if upgrading from a previous experimental version please see our new config docs accessing 'heartbeat.monitors.0' (source:'heartbeat.yml')
```

kuisathaverat:
superseded by https://github.com/elastic/synthetics/pull/257

PR #194, opened 2021-01-20
feat: capture performance metrics from trace events

### LINKED ISSUE #120: Capture relevant perf metrics to close the gap with RUM agent

vigneshshanmugam:
+ fixes #120 
+ Captures FirstContentfulPaint, LongestContentful Paint, Navigationorigin, Dom content loaded, Load event, Layout shift events, Cumulative Layout shift from chrome trace events
+ Replace `trace` flag with `filmstrips` in the CLI. 
+ Writes to the JSON reporter as `journey/metrics` to reflect the trace output timing information. 
+  Uses the Lighthouse Trace processor to extract the events specific to the current main process. 
+ included `Notice.txt` which ppints to Lighthouse license as we are using their trace processor for extracting key metrics. 

Trace output is in this format
```
{
  "type": "journey/metrics",
  "@timestamp": 1622578544529441,
  "journey": {
    "name": "check if input placeholder is correct",
    "id": "check if input placeholder is correct"
  },
  "root_fields": {
    "browser": {
      "experience": {
        "name": "navigationStart",
        "type": "mark",
        "start": 1014038.407051
      }
    },
    "os": {
      "platform": "darwin"
    },
    "package": {
      "name": "@elastic/synthetics",
      "version": "1.0.0-beta.2"
    }
  },
  "package_version": "1.0.0-beta.2"
}

```

### start, end - Timings that are relative to step/journey start and end times.


src/reporters/json.ts
```diff
@@ -124,6 +127,19 @@ export default class JSONReporter extends BaseReporter {
             });
           });
         }
+        const traces = userTiming.concat(experience, layoutShift);
+        traces.forEach(({ startTime, ts, name, type, duration, endTime }) => {
+          this.writeJSON('journey/timing', journey, {
+            payload: {
```
vigneshshanmugam, line 133: @andrewvc as we cannot use ECS fields for these new type of metrics. Do you have any concern here exposing these metrics via payload or we look for another way to represent these metric data?

src/parse_args.ts
```diff
@@ -60,7 +60,7 @@ program
     'pause on error until a keypress is made in the console. Useful during development'
   )
   .option('--screenshots', 'take screenshot for each step')
-  .option('--filmstrips', 'record detailed filmstrip info for all journeys')
+  .option('--trace', 'record chrome trace events for all journeys')
```
vigneshshanmugam, line 63: we might need a way to disable filmstrips alone as it's going to increase the doc size to a larger extent.

andrewvc: Why not keep the filmstrips arg and add the `--trace` arg to separate that then?

vigneshshanmugam: Hmmm was afraid of introducing more flags. Let me have a look at the commander library for 2 level objects like trace.filmstrips and `trace-perf` as such.

src/sdk/trace-processor.ts
```diff
@@ -0,0 +1,113 @@
...
+  lcpInvalidated: boolean;
+  processEvents: Array<TraceEvent>;
+  mainThreadEvents: Array<TraceEvent>;
+};
+
+/**
+ * Extends the lighthouse trace processor which extracts all the meaningful trace
+ * events in chronological order from the tab's process
+ */
+export class TraceProcessor extends LighthouseTraceProcessor {
```
andrewvc, line 85: Why do we extend the lighthouse TraceProcessor instead of using composition? Given that modern OO code tends to avoid inheritance where possible it would be good to give an explanation. Could this be accomplished more cleanly without inheritance?

__tests__/index.test.ts
```diff
@@ -66,7 +66,7 @@ describe('Run', () => {
       },
       headless: false,
       screenshots: true,
-      filmstrips: false,
+      trace: false,
```
andrewvc, line 69: Why did we remove the filmstrips option? It feels like these two things should be separate. I know there are similarities in the implementation but from a command line standpoint they do feel separate

andrewvc (commented):
Leaving a partial review here will follow up with a complete review ASAP

package.json
```diff
@@ -6,7 +6,7 @@
   "types": "dist/index.d.ts",
   "scripts": {
     "clean": "rm -rf dist",
-    "prepublish": "npm run clean && npm run build",
+    "prepublishOnly": "npm run clean && npm run build",
```
vigneshshanmugam, line 9: Note to self: need to revert this

src/sdk/trace-metrics.ts
```diff
@@ -0,0 +1,178 @@
...
+        measuresMap.set(name, ts);
+      } else if (phase === 'e') {
+        const startTime = measuresMap.get(name);
+        userTimings.push({
+          name,
+          ts,
+          type: 'measure',
+          startTime: convertTraceTimestamp(startTime),
+          endTime: convertTraceTimestamp(ts),
+          duration: ts - startTime,
```
vigneshshanmugam, line 74: note to self: check if this is correct. or endTime - startTime

src/parse_args.ts
vigneshshanmugam: We will keep both trace and filmstrips and address the heartbeat config handling in https://github.com/elastic/synthetics/issues/289

andrewvc: I think we have a solution for the too many flags issue in #289

src/core/gatherer.ts
```diff
@@ -73,10 +73,10 @@ export class Gatherer {
    */
   static async beginRecording(driver: Driver, options: RunOptions) {
     log('Gatherer: started recording');
-    const { filmstrips, network, metrics } = options;
+    const { trace, network, metrics } = options;
     const pluginManager = new PluginManager(driver);
     pluginManager.start('browserconsole');
-    filmstrips && (await pluginManager.start('trace'));
+    trace && (await pluginManager.start('trace'));
```
andrewvc, line 79: At some point does it make sense to use a `Promise.all` to init the plugins concurrently? Doesn't need to be in this PR

andrewvc (changes requested):

andrewvc:
FYI I did reach out to @drewpost and discuss the fact that metrics we collect are not aligned up with lighthouse or many other tools due to the fact that specific browser versions and platform differences will prevent numbers from lining up exactly.

Once this is merged we should be sure sure to document these differences and explain performance in the synthetics documentation. I've opened https://github.com/elastic/observability-docs/issues/735 to track

vigneshshanmugam:
@andrewvc Thanks Andrew for taking care of that and the info.

__tests__/index.test.ts
vigneshshanmugam: Brought it back.

src/sdk/trace-metrics.ts
```diff
@@ -0,0 +1,178 @@
...
+      firstContentfulPaintEvt,
+      largestContentfulPaintEvt,
+      loadEvt,
+      timeOriginEvt,
+      lcpInvalidated,
+    } = trace;
+
+    experienceMetrics.push(this.buildMetric(timeOriginEvt));
+    experienceMetrics.push(this.buildMetric(firstContentfulPaintEvt));
+    !lcpInvalidated &&
```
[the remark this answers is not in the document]
vigneshshanmugam: Added explanation and comment block.

src/sdk/trace-metrics.ts
```diff
@@ -0,0 +1,178 @@
...
+        name === 'requestStart' ||
+        name === 'navigationStart' ||
+        name === 'paintNonDefaultBackgroundColor' ||
+        args.frame !== undefined
+      ) {
+        continue;
+      }
+
+      /**
+       * user timing mark events falls under `R` and i or I
```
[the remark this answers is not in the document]
vigneshshanmugam: Added a doc link which explains in detail.

src/sdk/trace-processor.ts
vigneshshanmugam: Added explanation.

src/core/gatherer.ts
vigneshshanmugam: Done.

src/common_types.ts
```diff
@@ -31,10 +31,29 @@ export type VoidCallback = () => void;
...
+
+export type Filmstrip = TraceOutput & {
+  snapshot: string;
+};
+
+export type LayoutShift = {
+  exists: boolean;
+  score: number;
+  name: string;
+  ts?: number;
```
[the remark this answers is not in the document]
vigneshshanmugam: Got rid of `ts`, not required. Kept it to `start` which is the relative time based on `step/journey` start and end times.

src/sdk/trace-metrics.ts
vigneshshanmugam: Fixed.

src/reporters/json.ts
andrewvc: Coming back to take a second look I'm really glad I caught this comment. These are the sorts of metrics users will want to query with custom dashboards and putting them inside the payload will prevent that.

Beyond user dashboards this prevents us from doing our own aggregations and charts over time it is also a big problem.

Let's take the time to add these fields. The good news is that we can use the new root fields to index these into elasticsearch without having to modify the heartbeat go code. We will have to update the elasticsearch mappings in heartbeat but that is much simpler. You will also have to update the new synthetics mappings in the fleet integration

I am okay with us merging this PR just to get the code in however we should not build any features in the kibana UI based on it until we've moved to values into properly indexed fields

andrewvc (approved):
LGTM however I'll leave it up to you to determine whether you want to move the fields out of the payload now or in a subsequent PR. We can't build GUI features based on this until we have proper elasticsearch fields defined in our mappings.

src/reporters/json.ts
vigneshshanmugam: Yeah ++ to all your points, Since these features are anyway behind feature flag and heartbeat right now does not even use them, we are better off here and start building the UI once we move them out in the next release. I am also fine with adding them under `root_fields` in the same PR. Do you have any suggestion on the naming?

andrewvc: It's tricky, I think let's come up with a name that would make sense to potentially get adopted into ECS extended fields at some point. That said, I do think they likely are too specific for ECS. 

What about `browser.*` as a base for these fields. So, we could use `browser.user_timing.*` for user timing metrics. Perhaps `browser.core_vitals.*` etc. WDYT?

vigneshshanmugam: ++, I like that idea of nesting them under `browser.` Are you fine with them lying under `journey/metrics` type?

vigneshshanmugam: Moved all fields under `browser.*`

andrewvc (approved):
LGTM

PR #195, opened 2021-01-21
fix: transpile suites to latest target

### LINKED ISSUE #172: Transpiled source is displayed instead of raw typescript
Without a `tsconfig.json` file like the one in the root of this project present synthetics will only show the transpiled (ugly) JS.  I think the solution here is to bring back what we used to do, where we hard-coded the tsconfig settings when requiring 'typescript'. This was introduced here: https://github.com/elastic/synthetics/pull/98/files#diff-fa8d4e24d8399e8350f1c8bad05df53e8032ea995835bf911507015e2db61cddR115 but later backed out.

@vigneshshanmugam I'd be curious as to your thoughts here

vigneshshanmugam:
+ fix #172 
+ Moved the source map support during run time instead of import time.

### DIFF
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -120,13 +120,8 @@ async function prepareSuites(inputs: string[]) {
   } else {
     /**
      * Preload modules before running the suites
-     * we support `.ts` files out of the box by invoking
-     * the `ts-node` via `transpile-only` mode which only compiles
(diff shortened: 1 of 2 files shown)


andrewvc (approved):
LGTM, tested locally

PR #196, opened 2021-01-25
chore: update test page to fix tests

vigneshshanmugam:
+ With this change #186 , we moved our server code to `__tests__` but forgot to move the pages which resulted in status of 404 instead of 200. 
+ Updated to use `matchObject` which provided the error in a better way.


andrewvc (approved):
LGTM

PR #189, opened 2021-01-18
Fix linted shellcheck defects

v1v:
### What

- Add pre-commit
- Fix linted defects from https://github.com/koalaman/shellcheck/
- Use shebang with `#!/usr/bin/env bash`
- Update the contributions with the pre-commit

### Tests

- UI test output for the pre-commit linting 

image

image

PR #200, opened 2021-01-27
Fix broken overwrite of heartbeat args

andrewvc:
This was broken in https://github.com/elastic/synthetics/pull/189 which quoted this variable that should not have been quoted


jahtalab (approved):
LGTM

PR #203, opened 2021-01-28
fix: handle exit code correctly

### LINKED ISSUE #191: exit with non-zero code when any step fails
+ Right now, we always exit the process with `0` even when there are step that contains failures. 

```
journey('test', async ({ page }) => {
  step('error step', async () => {
    throw new Error('boom');
  });
});
```

We already capture and show this error on both of our reporters, but we have to exit with non-zero code when `result.status == 'failed'`

vigneshshanmugam:
+ fix #191 
+ Handle the exit code from the CLI runs correctly, now we exit with >0 error codes when any of the journey fails which include steps as well. If any error is thrown via assertion or error - CLI will exit with 1 exit code
+ Added tests by creating a CLI mock that runs and waits for the exit code.
+ Cleaned up some unwanted CLI args parsing code which is not relevant for our run API.

### DIFF
--- a/__tests__/index.test.ts
+++ b/__tests__/index.test.ts
@@ -23,16 +23,16 @@
  *
  */
 
+import fs from 'fs';
 import { run, journey } from '../src/index';
 import { runner } from '../src/core';
-import * as ParseArgs from '../src/parse_args';
+import { RunOptions } from '../src/core/runner';
 import { generateTempPath } from '../src/helpers';
-import fs from 'fs';
 
 describe('Run', () => {
   const dest = generateTempPath();
   afterEach(() => {
-    fs.unlinkSync(dest);
+    fs.existsSync(dest) && fs.unlinkSync(dest);
   });
 
   it('multiple run invokes runner only once', async () => {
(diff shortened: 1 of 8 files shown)


__tests__/core/browser-service.test.ts
```diff
@@ -24,26 +24,25 @@
...
     await browserService.dispose();
   });
 
   it('should create browser pages', async () => {
-    const browser = await chromium.connect({
-      wsEndpoint: 'ws://localhost:9323',
-    });
-    const context = await browser.newContext();
-    const page = await context.newPage();
-    await page.goto('https://elastic.co');
```
vigneshshanmugam, line 46: missed this in my last review. we use local dev server to test. fixed it now.

__tests__/index.test.ts
```diff
@@ -56,7 +56,7 @@ describe('Run', () => {
   });
 });
 
-describe('Run - cli args', () => {
+describe.skip('Run - cli args', () => {
```
vigneshshanmugam, line 59: These tests tie up our run API with CLI which is incorrect, I will clean this up or will do a follow up PR to get rid of this and add proper tests for CLI for arguments.

src/cli.ts
```diff
@@ -175,6 +175,16 @@ async function prepareSuites(inputs: string[]) {
...
+
+  /**
+   * Exit with error status if any journey fails
+   */
+  for (const result of Object.values(results)) {
+    if (result.status === 'failed') {
+      process.exit(1);
+    }
+  }
+  return results;
```
vigneshshanmugam, line 187: note to self: remove

src/index.ts
```diff
@@ -36,34 +35,22 @@ export async function run(options: RunOptions) {
   sourceMapSupport.install({
     environment: 'node',
   });
-
-  const cliArgs = parseArgs();
```
vigneshshanmugam, line 40: Not sure how this got introduced, but this is wrong. Our CLI already parses and sends the parsed options to our run API.

__tests__/index.test.ts
vigneshshanmugam: Fixed now.

andrewvc (approved):
LGTM. Tested manually as well, works as advertised

PR #204, opened 2021-01-28
chore: add functional CLI tests

### LINKED ISSUE #139: Add Functional test for CLI
Currently we test the CLI functionality via the API, but we should add tests that:

1. Build the package
2. Install it
3. Actually invoke the CLI app and check for basic functionality

vigneshshanmugam:
+ fix #139 
+ Added tests for few flags that can be passed via CLI


andrewvc (approved):
LGTM

PR #215, opened 2021-02-04
feat: move fields to ECS format

### LINKED ISSUE #164: Move network fields to ECS where possible  (starting with network data)
We currently write a lot of our JSON data to the `payload` section of the documents in ES. I propose we add `http` and `tls` keys matching the ECS spec to our JSON output, as well as a `http.timings` field that contains the calculated times for the various waterfall colors (connect, send, receive, etc.)

See the ECS reference for [http](https://www.elastic.co/guide/en/ecs/current/ecs-http.html), [URL](https://www.elastic.co/guide/en/ecs/current/ecs-url.html) and [tls](https://www.elastic.co/guide/en/ecs/current/ecs-tls.html) fields for more detail

vigneshshanmugam:
+ fix #164 
+ All ecs related field are nested under `root_fields` as discussed.

### DIFF
--- a/src/plugins/network.ts
+++ b/src/plugins/network.ts
@@ -42,6 +42,10 @@ export class NetworkManager {
       'Network.requestWillBeSent',
       this._onRequestWillBeSent.bind(this)
     );
+    client.on(
+      'Network.requestWillBeSentExtraInfo',
+      this._onRequestWillBeSentExtraInfo.bind(this)
+    );
     client.on('Network.responseReceived', this._onResponseReceived.bind(this));
     client.on('Network.loadingFinished', this._onLoadingFinished.bind(this));
     client.on('Network.loadingFailed', this._onLoadingFailed.bind(this));
@@ -107,6 +111,23 @@ export class NetworkManager {
     });
   }
 
+  _onRequestWillBeSentExtraInfo(
+    event: Protocol.Network.requestWillBeSentExtraInfoPayload
+  ) {
+    const { requestId, headers } = event;
+    const record = this.waterfallMap.get(requestId);
+    if (!record) {
+      return;
+    }
+    /**
+     * Enhance request headers with additional information
+     */
+    record.request.headers = {
+      ...record.request.headers,
+      ...headers,
+    };
+  }
+
   _onResponseReceived(event: Protocol.Network.responseReceivedPayload) {
     const { requestId, response, timestamp } = event;
     const record = this.waterfallMap.get(requestId);
@@ -118,15 +139,6 @@ export class NetworkManager {
       response,
       responseReceivedTime: timestamp,
     });
-    /**
-     * Enhance request headers with additional information
-     */
-    if (response.requestHeaders) {
-      record.request.headers = {
-        ...record.request.headers,
-        ...response.requestHeaders,
-      };
-    }
   }
 
   _onLoadingFinished(event: Protocol.Network.loadingFinishedPayload) {
--- a/src/plugins/network.ts
+++ b/src/plugins/network.ts
@@ -25,15 +25,19 @@
 
 import { CDPSession } from 'playwright-chromium';
 import { Protocol } from 'playwright-chromium/types/protocol';
-import { NetworkInfo } from '../common_types';
+import { NetworkInfo, BrowserInfo } from '../common_types';
 import { Step } from '../dsl';
 import { getTimestamp } from '../helpers';
 
 export class NetworkManager {
+  private _browser: BrowserInfo;
   _currentStep: Partial<Step> = null;
   waterfallMap = new Map<string, NetworkInfo>();
 
   async start(client: CDPSession) {
+    const { product } = await client.send('Browser.getVersion');
+    const [name, version] = product.split('/');
+    this._browser = { name, version };
     await client.send('Network.enable');
     /**
      * Listen for all network events
@@ -95,6 +99,7 @@ export class NetworkManager {
     }
 
     this.waterfallMap.set(requestId, {
+      browser: this._browser,
       step: this._currentStep,
       timestamp: getTimestamp(),
       url,
--- a/src/reporters/json.ts
+++ b/src/reporters/json.ts
@@ -72,18 +72,20 @@ function getMetadata() {
   };
 }
 
-function formatHttpVersion(protocol: string) {
+function formatHTTPVersion(protocol: string) {
   if (protocol === 'h2') {
     return 2;
   } else if (protocol === 'http/1.1') {
     return 1.1;
   } else if (protocol === 'http/1.0') {
     return 1.0;
+  } else if (protocol && protocol.startsWith('h3')) {
+    return 3;
   }
   return;
 }
 
-function formatECSFields(network: NetworkInfo) {
+export function formatNetworkFields(network: NetworkInfo) {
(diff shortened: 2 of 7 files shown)


src/plugins/network.ts
```diff
@@ -107,6 +111,23 @@ export class NetworkManager {
...
+  ) {
+    const { requestId, headers } = event;
+    const record = this.waterfallMap.get(requestId);
+    if (!record) {
+      return;
+    }
+    /**
+     * Enhance request headers with additional information
+     */
+    record.request.headers = {
```
vigneshshanmugam, line 125: this is done here instead of on response as `response` event can sometimes be not fired for dropped requests, so moved it before. Also this lets us do our next iteration in our waterfall which is `Add flyout in waterfall`

src/reporters/json.ts
```diff
@@ -27,18 +27,119 @@ import BaseReporter from './base';
...
+  timestamp?: number;
+  url?: string;
+  step?: Partial<Step>;
+  error?: Error;
+  root_fields?: Record<string, unknown>;
+  payload?: Record<string, unknown>;
+  blob?: string;
+};
+
+function getMetadata() {
```
vigneshshanmugam, line 57: @andrewvc I picked only minimal things for metadata, if you have any ideas for other fields, we can incorporate that as well.

src/reporters/json.ts
```diff
@@ -93,8 +198,11 @@ export default class JSONReporter extends BaseReporter {
...
         if (networkinfo) {
           networkinfo.forEach(ni => {
-            this.writeJSON('journey/network_info', journey, {
+            this.writeJSON({
+              type: 'journey/network_info',
+              journey,
               timestamp: ni.timestamp,
+              root_fields: formatNetworkFields(ni),
               step: ni.step,
               payload: snakeCaseKeys(ni),
```
andrewvc, line 207: So, it looks like we're writing the ECS fields in addition to the raw network data. Seems like we could possibly just drop payload with this patch?

src/reporters/json.ts
```diff
@@ -27,18 +27,119 @@ import BaseReporter from './base';
...
+        ? {
+            body: {
+              bytes: response.encodedDataLength,
+            },
+            mime_type: response.mimeType,
+            status_code: response.status,
+          }
+        : undefined,
+    },
+    tls: tls
```
andrewvc, line 117: We should have this match what heartbeat does more closely, specifically nesting under `tls.server` instead of `tls.client` and enumerating more fields if possible. The `subject` field is particularly imporant.

Here's an example of the fields heartbeat provides.

```json
{
  "cipher": "ECDHE-RSA-AES-128-GCM-SHA256",
  "server": {
    "x509": {
      "subject": {
        "common_name": "r2.shared.global.fastly.net",
        "distinguished_name": "CN=r2.shared.global.fastly.net,O=Fastly\\, Inc.,L=San Francisco,ST=California,C=US"
      },
      "serial_number": "25044290308248398422640150363",
      "public_key_size": 2048,
      "public_key_exponent": 65537,
      "not_before": "2021-02-05T17:27:13.000Z",
      "issuer": {
        "common_name": "GlobalSign CloudSSL CA - SHA256 - G3",
        "distinguished_name": "CN=GlobalSign CloudSSL CA - SHA256 - G3,O=GlobalSign nv-sa,C=BE"
      },
      "not_after": "2021-06-20T21:11:25.000Z",
      "signature_algorithm": "SHA256-RSA",
      "public_key_algorithm": "RSA"
    },
    "hash": {
      "sha256": "bf2b471bc6d729f0b0359bbaaf7cead9c1dffd7b8e7a15ea136e7468ef8b7d87",
      "sha1": "e005c41f94da793d7f812c092e1a432b6c58498e"
    }
  },
  "established": true,
  "rtt": {
    "handshake": {
      "us": 31066
    }
  },
  "version_protocol": "tls",
  "version": "1.2"
}
```

src/reporters/json.ts
```diff
@@ -27,18 +27,119 @@ import BaseReporter from './base';
...
+      version: formatHTTPVersion(response?.protocol),
+      request: {
+        body: {
+          bytes: postdata.length,
+          content: postdata,
+        },
+        method: request.method,
+        referrer: request.headers?.Referer,
+      },
+      response: response
```
andrewvc, line 107: I notice a common pattern here of `value ? {...fields} : undefined`. It'd be better to have the key not be present if there are no fields since ES stores the full JSON source.

andrewvc (commented):
Generally looks good, left some notes on areas where it could be improved

src/reporters/json.ts
```diff
@@ -27,18 +27,119 @@ import BaseReporter from './base';
...
+  step?: Partial<Step>;
+  error?: Error;
+  root_fields?: Record<string, unknown>;
+  payload?: Record<string, unknown>;
+  blob?: string;
+};
+
+function getMetadata() {
+  return {
+    process: {
```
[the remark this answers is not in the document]
vigneshshanmugam: Sure, let's get rid of it.

src/reporters/json.ts
```diff
@@ -27,18 +27,119 @@ import BaseReporter from './base';
...
+
+export function formatNetworkFields(network: NetworkInfo) {
+  const { request, response, url } = network;
+  const postdata = request.postData || '';
+  const tls = response?.securityDetails;
+
+  return {
+    // URL and USER AGENT would be parsed and mapped by heartbeat
+    url,
+    user_agent: request.headers?.['User-Agent'],
```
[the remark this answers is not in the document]
vigneshshanmugam: APM server uses ES https://www.elastic.co/guide/en/elasticsearch/reference/current/user-agent-processor.html to parse the user agent fields, I would like heartbeat do the same thing if possible when writing our docs.

src/reporters/json.ts
vigneshshanmugam: Discussed offline, we will add all payload network fields for now to root_fields and drop payload on next version.

src/reporters/json.ts
vigneshshanmugam: Great point, will dig more and add these.

src/reporters/json.ts
vigneshshanmugam: when `undefined` gets `JSON.stringify(undefined)`, we wont see them on the output. So we shouldnt need to worry about them.

src/reporters/json.ts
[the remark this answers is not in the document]
andrewvc: Hmmm, while this is possible heartbeat actually doesn't use ingest node at all at the moment. I'd hate to introduce it for a single field. Additionally, I don't know if there's a huge benefit to this field for synthetics, we could get the browser version in other ways without parsing the UA since we know which browser we're on.

Could we just set `user_agent.original` and `user_agent.version` and `user_agent.name` as fields? That data should be readily available in synthetics and should simplify things greatly.

andrewvc (approved):
LGTM pending the UA change

src/reporters/json.ts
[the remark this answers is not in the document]
vigneshshanmugam: Hmmm that should be a minimal change and we can keep it like that for now until we need the full blown searching and other capability .

PR #227, opened 2021-02-23
fix: disable chromium sandboxing by default

### LINKED ISSUE #225: Default sandbox to off, add `--sandbox` option
We currently support  a `--no-sandbox` option, but we should make this behavior the default since the general use case for synthetics is not security sensitive, and users frequently desire to run on headless servers which essentially require this. 

We should keep the explicit `--no-sandbox` option but also add a positive `--sandbox` option.

vigneshshanmugam:
+ fix #225
+ By default the chromium sanboxing is disabled and can be enabled with `--sandbox` CLI option or using our runner API.
+ Cleaned up the driver arguments. 
```ts
import { run } from "@elastic/synthetics";

await run({ sandbox: true })
```

### DIFF
--- a/__tests__/core/browser-service.test.ts
+++ b/__tests__/core/browser-service.test.ts
@@ -41,7 +41,9 @@ describe('BrowserService', () => {
   });
 
   it('should create browser pages', async () => {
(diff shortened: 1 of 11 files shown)


andrewvc (approved):
LGTM

PR #228, opened 2021-02-23
refactor: wait till load when screenshot is enabled

vigneshshanmugam:
+ Wait for the load event only when screenshots flag is enabled. 
+ waiting for the load event for normal scenarios is unnecessary and `screenshots`


andrewvc:
Works great, and about 30% faster when running the todos example!

Issue #230, opened 2021-02-24
expose `status` field as part of journey/end and step/end


vigneshshanmugam:
+ Right now we expose the status -> succeeded, failed as part of the payload field which is not searchable. 
+ We move that to the `journey` and `step` schema which would make things easier for the heartbeat and also for the UI to query these fields. 

```
{
    "type": "journey/end"
   "journey": {
       "name": "blah",
       "id": "blah",
       "status": "failed"
    },
    "error": {}
}

{
    "type": "step/end"
   "step": {
       "name": "blah",
       "id": "blah",
       "status": "failed"
    },
    "error": {}
}
```

PR #231, opened 2021-03-01
expose step status field via step and journey

shahzad31:
Fixes: https://github.com/elastic/synthetics/issues/230

Exposed status field for step/end and journey/end documents being sent to heartbeat


shahzad31:
can't figure out CI errors :/

src/reporters/json.ts
```diff
@@ -167,7 +181,7 @@ export default class JSONReporter extends BaseReporter {
         type: 'journey/start',
         journey,
         timestamp,
-        payload: { params, source: journey.callback.toString },
+        payload: { params, source: journey.callback.toString() },
```
vigneshshanmugam, line 184: good catch 👍

src/reporters/json.ts
```diff
@@ -51,7 +65,7 @@ type OutputFields = {
   step?: Partial<Step>;
   error?: Error;
   root_fields?: Record<string, unknown>;
-  payload?: Record<string, unknown>;
+  payload?: Payload | NetworkInfo;
```
vigneshshanmugam, line 68: If we have to type this, proper way would be to type it for each `type` -> `journey/start`, `journey/end` and such. I am not 100% sure if this required. 

One other way is to use the `Partial` type and keep it minimal and add Record<string, unknown> for the rest.

src/reporters/json.ts
```diff
@@ -298,11 +312,17 @@ export default class JSONReporter extends BaseReporter {
       journey: {
         name: journey.name,
         id: journey.id,
+        ...(type === 'journey/end' ? { status: payload.status } : {}),
```
vigneshshanmugam, line 315: you can just leave it as 
```
journey: {
   name: journey.name,
   id: journey.id,	       
   status: payload.status,
}
```

This would work only for `journey/end` and `step/end`

src/reporters/json.ts
```diff
@@ -298,11 +312,17 @@ export default class JSONReporter extends BaseReporter {
...
       },
       step: step
         ? {
             name: step.name,
             index: step.index,
+            ...(type === 'step/end'
```
vigneshshanmugam, line 321: same as above.

src/reporters/json.ts
shahzad31: we only need status for journey/end and step/end, for others it doesnt make sense to add status.

src/reporters/json.ts
shahzad31: i was almost about to write conditional types, but didn't want to spend too much time on this PR :D

vigneshshanmugam: yeah lets not go down the rabbit hole and spend too much time on this, we will eventually get rid of payload at some point. lets keep it as it is.

src/reporters/json.ts
vigneshshanmugam: Reverted back my suggestion.

vigneshshanmugam (approved):

PR #239, opened 2021-03-09
feat: expose driver types from the agent

vigneshshanmugam:
+ Solves the problem specified in #158
+ Exposes all of the driver types that could be used from the suites
+ It solves the problem of users needing to import `playwright-core` or `playwright-*` for using types and eventually run in to problems of version mismatch. 
+ Once we add support for multiple browsers, we would expose the types from `playwright-core` instead and rest would look the same

### DIFF
--- a/examples/todos/package.json
+++ b/examples/todos/package.json
@@ -5,7 +5,6 @@
   "scripts": {},
   "license": "MIT",
   "dependencies": {
-    "@elastic/synthetics": "*",
-    "playwright-core": "=1.6.2"
+    "@elastic/synthetics": "file:../../"
   }
 }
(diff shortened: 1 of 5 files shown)


examples/todos/package.json
```diff
@@ -5,7 +5,6 @@
   "scripts": {},
   "license": "MIT",
   "dependencies": {
-    "@elastic/synthetics": "*",
-    "playwright-core": "=1.6.2"
+    "@elastic/synthetics": "file:../../"
```
vigneshshanmugam, line 8: will change once we merge and release a new version where types would become available.

PR #247, opened 2021-03-27
feat: report number of journeys in json

vigneshshanmugam:
+ Reports the number of journeys as `synthetics/metadata` event which would be potentially used in the `--dry-run` command which reports this event as well registered journeys via `journey/register` command. 

Dry run in todos basic example 
```
{"type":"synthetics/metadata","@timestamp":1616806791018720,"root_fields":{"num_journeys":2,"os":{"platform":"darwin"},"package":{"name":"@elastic/synthetics","version":"0.0.1-alpha.11"}},"package_version":"0.0.1-alpha.11"}
{"type":"journey/register","@timestamp":1616806791019158,"journey":{"name":"check if title is present","id":"check if title is present"},"root_fields":{"os":{"platform":"darwin"},"package":{"name":"@elastic/synthetics","version":"0.0.1-alpha.11"}},"package_version":"0.0.1-alpha.11"}
{"type":"journey/register","@timestamp":1616806791019203.2,"journey":{"name":"check if input placeholder is correct","id":"check if input placeholder is correct"},"root_fields":{"os":{"platform":"darwin"},"package":{"name":"@elastic/synthetics","version":"0.0.1-alpha.11"}},"package_version":"0.0.1-alpha.11"}

```

### DIFF
--- a/src/reporters/json.ts
+++ b/src/reporters/json.ts
@@ -35,6 +35,7 @@ import { Metrics } from '../plugins';
 const { version, name } = require('../../package.json');
 
 type OutputType =
+  | 'synthetics/metadata'
   | 'journey/register'
   | 'journey/start'
   | 'step/screenshot'
@@ -59,7 +60,7 @@ type Payload = {
 
 type OutputFields = {
   type: OutputType;
-  journey: Journey;
+  journey?: Journey;
   timestamp?: number;
   url?: string;
   step?: Partial<Step>;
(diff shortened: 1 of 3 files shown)


src/reporters/json.ts
```diff
@@ -167,8 +168,51 @@ export function formatNetworkFields(network: NetworkInfo) {
...
   _registerListeners() {
+    /**
+     * report the number of journeys that exists on a suite which
+     * could be used for better sharding
+     */
+    this.runner.on('start', ({ numJourneys }) => {
+      this.writeJSON({
+        type: 'synthetics/metadata',
+        root_fields: {
+          num_journeys: numJourneys,
```
andrewvc, line 211: I'm OK with this, but we'll need to confirm that heartbeat doesn't exist this event. It shouldn't because it should only index events inside a journey context, but it's something to be aware of. 

We should follow-up by adding explicit support to `synthetics/metadata` events to heartbeat

andrewvc (approved):

src/reporters/json.ts
vigneshshanmugam: Yes thats correct, heartbeat doesn't also support `journey/register` events if i remember correctly. Would it still pipe these events to the stdout?

PR #254, opened 2021-04-14
add support for custom reporters

spalger:
In order to support getting access to runner events from the standard `run()` API I think it would be good to support custom `Runner` classes. This PR adds support for a `Runner` class being passed as the `reporter` option, `new`s up this class when the runner starts, and passes the `Runner` instance as well as the `ReporterOptions` to the constructor, and then moves on.


vigneshshanmugam (approved):
Neat and I like the idea of allowing customer reporters, Thanks @spalger 🎉

PR #258, opened 2021-04-21
feat: use jpeg for screenshots

### LINKED ISSUE #233: Add option for JPEG screenshots
For those who are space conscious JPEG screenshots would be welcome as a CLI option. We would need to add a corresponding heartbeat option.

vigneshshanmugam:
+ fix #233 
+ Defaults to jpeg images with quality 80(default for jpeg in playwright) for screenshots, Size is 2x smaller than PNG for todos example 
```
28060 vs 14936 kB
```

**Note: Verified that Kibana Uptime UI works even without the proper mime type being set**


andrewvc (approved):
LGTM

PR #261, opened 2021-04-22
fix: keep journey callback to be sync

### LINKED ISSUE #256: Journey callbacks shouldn't be async
`journey()` callbacks are currently forced to be async by the types, but best I can tell we shouldn't be doing any async work in those callbacks, right? All async work should be wrapped in a `step()` and then executed by synthetics. I think that would be more obvious if the types required a `void` return type for the callback.

vigneshshanmugam:
+ fix #256 
+ Updated tests which was expecting async all over the place.


andrewvc (approved):
LGTM, but since this is breaking, let's wait on the next release so we can do some messaging on the forums

PR #268, opened 2021-05-04
fix: exclude data uri from network events

### LINKED ISSUE #267: Don’t store data URI request details
Storing data URIs and showing them on the waterfall chart is misleading and confusing for users.  We have no requirement for these right now, so, along with the space saving, we should stop Synthetics from reporting these in the result data.

vigneshshanmugam:
+ fix #267


andrewvc (approved):
LGTM

PR #271, opened 2021-05-06
feat: add synthetic config file support

### LINKED ISSUE #270: [Proposal] allow configuration through `synthetics.config.js`
+ Suite params options is super useful for passing down params for every journey files, however it becomes difficult when there are multiple environments involved like `staging`, `ci` and `production` and the whole suite level params keep changing for each of those environments. The proposal address that by allowing a configuration file which lets users inject params depending on the environment. 

```js
module.exports = {
  // defaults to development
  params: {
    url: 'http://localhost:8080',
  },
};

// To change params based on environment, users can return
// a function

module.exports = env => {
  if (env === 'development') {
    return {
      params: {
        url: 'http://localhost:8080',
      },
    };
  }
  return {
    params: {
      url: 'http://elastic.co',
    },
  };
};
```

**Note: params is explicitly kept as key to allow extending this configuration in the future.**

This setup and validation of config would be done before invoking any journeys and we would pass down these parameters to the individual journeys. 

```js
beforeAll(({ params }) => {
  let server;
  if (params.env === "development") {
    server = await startServer();
  }
});

afterAll(({ params }) => {
  if (params.env === "development") {
    server.stop()
  }
})

journey('journey1', ({ page, params }) => {
  step('step1', async () => {
    await page.goto(params.url);
  });
});

```

Purposefully this model exists as we don't want to support any other way of loading configs, By that I mean we don't want to support `.json`, `.yml` or any other extensions. Let's stick to a single way of doing config loading.

vigneshshanmugam:
+ fix #270 
+ PR allow users to supply `--config <synthetics.config.{js|ts}>` that lets users control the params passed to the journey callbacks and also for `before` and `after` callbacks. 
+ `params` object exposed to journeys are frozen meaning users will not be able to extend it. 
+ `params.environment` defaults to "development" and can be override by specifying `-e production` or through NODE_ENV="production" when running synthetics CLI. 

### Usage
```sh
npx @elastic/synthetics examples/todos --config examples/todos/synthetics.config.ts
```

```ts
// synthetics.config.ts
export default env => {
  if (env === 'development') {
    return {
      params: {
        url: 'http://localhost:8080',
      },
    };
  }
  return {
    params: {
      url: 'http://elastic.co',
    },
  };
};


journey('journey1', ({ page, params }) => {
  before(async ({ params }) => {
    if (params.environment === 'dev') {
      await startServer();
    }
  });
  after(({ params }) => {
    if (params.environment === 'dev') {
      stopServer();
    }
  });
  step('step1', async () => {
    await page.goto(params.url);
  });
});

```

### Config lookup algorithm

+ If `--config <filename>` option is provided, we look for the exact file and throw error in case the file does not exists.
+ If the option is not supplied, we look for presence of `synthetics.config.js` and `synthetics.config.ts` in the current suites directory all the way to the root. 
    - If present, we resolve to that file
    - If it does not exists, we resolve to empty object


src/cli.ts
```diff
@@ -158,13 +159,23 @@ async function prepareSuites(inputs: string[]) {
...
   /**
+   * Use the NODE_ENV variable to control the environment if its not explicity
+   * passed from either CLI or through the API
+   */
+  const environment = options.environment || process.env['NODE_ENV'];
+  /**
+   * Validate and handle configs
+   */
+  const config = await readConfig(options.config, environment);
+  const params = config.params || JSON.parse(options.suiteParams);
```
vigneshshanmugam, line 170: TODO: Add env to params by default

__tests__/plugins/network.test.ts
```diff
@@ -128,6 +128,7 @@ describe('network', () => {
     });
 
     await driver.page.goto(server.PREFIX + '/index');
+    await driver.page.waitForLoadState();
```
vigneshshanmugam, line 131: This is irrelevant to the current PR, but fixes the flakiness in runs,

src/cli.ts
```diff
@@ -157,14 +158,23 @@ async function prepareSuites(inputs: string[]) {
...
   /**
    * use JSON reporter if json flag is enabled
    */
   const reporter = options.json ? 'json' : options.reporter;
 
   const results = await run({
-    params: JSON.parse(options.suiteParams),
-    environment: options.environment,
+    params: Object.freeze({ ...params, environment }),
```
andrewvc, line 177: Glad to see we are freezing this object here that you prevent a lot of errors. So

andrewvc (approved):
LGTM

vigneshshanmugam:
@andrewvc In this PR, i have not exposed the params to any hooks (`before`, `beforeAll` etc), I am still thinking though, Shall we expose it as part of all hooks as we don't allow users to modify it and there is no harm in doing so? We can so that as follow up PR as this PR adds only config support.

vigneshshanmugam:
Had a offline discussion, we will have it for all the hooks in separate PR.

PR #272, opened 2021-05-06
feat: expose params in all hooks

vigneshshanmugam:
+ Exposes the suite params and params passed via `--config` flag to all our hooks `before, beforeAll, after, afterAll`
+ Params in hooks cannot be modified, throws error otherwise - added a test in `cli.test.ts`. 

+ Follow up of the config support PR #271

### DIFF
--- a/src/core/runner.ts
+++ b/src/core/runner.ts
@@ -33,8 +33,9 @@ import {
   FilmStrip,
   NetworkInfo,
   HooksCallback,
-  RunParams,
+  Params,
   CliArgs,
+  HooksArgs,
 } from '../common_types';
 import { BrowserMessage, PluginManager } from '../plugins';
 import { PerformanceManager, Metrics } from '../plugins';
@@ -50,14 +51,13 @@ export type RunOptions = Omit<
   | 'require'
   | 'suiteParams'
   | 'reporter'
-  | 'environment'
 > & {
-  params?: RunParams;
+  params?: Params;
   reporter?: CliArgs['reporter'] | Reporter;
 };
 
 type BaseContext = {
-  params?: RunParams;
+  params?: Params;
   start: number;
   end?: number;
 };
@@ -93,7 +93,7 @@ interface Events {
   'journey:start': {
     journey: Journey;
     timestamp: number;
-    params: RunParams;
+    params: Params;
   };
   'journey:end': BaseContext &
     JourneyResult & {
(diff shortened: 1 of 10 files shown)


src/core/runner.ts
```diff
@@ -150,24 +149,24 @@ export default class Runner {
     this.eventEmitter.on(e, cb);
   }
 
-  async runBeforeAllHook() {
+  async runBeforeAllHook(params: RunParamaters) {
```
andrewvc, line 152: Can we rename this parameter runParams? It's confusing with both RunParams and Params existing

src/dsl/journey.ts
```diff
@@ -25,22 +25,21 @@
...
-export type Hooks = Record<HookType, Array<VoidCallback>>;
-
+export type Hooks = Record<HookType, Array<HooksCallback>>;
 export type JourneyCallback = (options: {
   page: Page;
   context: BrowserContext;
   browser: Browser;
   client: CDPSession;
-  params: Record<string, any>;
+  params: RunParamaters;
```
andrewvc, line 42: Typo, parameters not parameters, also, can we call the arg runParams?

andrewvc (commented):
Generally looks good but the naming could be improved

src/dsl/journey.ts
vigneshshanmugam: Sure, lets change it to runParams

src/core/runner.ts
vigneshshanmugam: I really have to change my keyboard :P, fixing it.

src/dsl/journey.ts
vigneshshanmugam: I really have to change my keyboard :P, fixing it.

src/core/runner.ts
vigneshshanmugam: Done

vigneshshanmugam: Probbaly I misunderstood, if we change it to `runParams` in the args, do you want to pass them as `runParams` in the hooks as well?

andrewvc: Okay I see the confusion here. I think we are on different pages. I had missed this change (https://github.com/elastic/synthetics/commit/cb01d24f6b073fd36b954ff7e16b10be4cb1e0ea#diff-fd47e6aef3e1ae139693b20fdc424b2a557e7214608e8f22e931b99e11d549bbR58) in the config refactor commit.

The question is, is the `environment` flag something that modifies a special key in `--suite-params` or is it a fully separate value passed on its own? Given that they are passed as separate flags I don't think they should be mixed into the same map. I think we should expose it as a fully separate variable as a sibling alongside params.

So, we'd have a signature like:

`beforeAll({params, env} => {})` Instead of what we have today. What do you think?

vigneshshanmugam: We agreed on splitting param and env offline. 

will make the changes here and ask for review.

vigneshshanmugam: Done.

andrewvc (approved):
LGTM

PR #273, opened 2021-05-07
fix: capture hook errors in base reporter

vigneshshanmugam:
+ When there are errors thrown in any of the `before` and `after` hooks, the JSON reporter captures them, but we dont capture it on the base reporter which is a bug. Now we render the error and also print out the duration it took to run the journeys.

### DIFF
--- a/__tests__/reporters/__snapshots__/base.test.ts.snap
+++ b/__tests__/reporters/__snapshots__/base.test.ts.snap
@@ -1,5 +1,18 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
+exports[`base reporter render hook errors without steps 1`] = `
+"
+Journey: j1
+   ---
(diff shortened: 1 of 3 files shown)


paulb-elastic:
Discussed in the team meeting yesterday, this isn't required for 7.13, so removing the label.

vigneshshanmugam:
@andrewvc You are right, we dont handle the beforeAll and afterAll hooks properly and yes it totally makes sense. But i am worried about propagating the error down the line, let me create an issue and we can discuss solutions there. 

Lets merge this PR as if you don't have any other concerns with base reporter.

paulb-elastic:
Linking to https://github.com/elastic/synthetics/issues/280

vigneshshanmugam:
`beforeAll` and `afterAll` will be handled as part of the other issue - #280

PR #274, opened 2021-05-11
perf: use context isolation for inidividual journeys

vigneshshanmugam:
+ This PR improves the test execution time when there are multiple journeys involved while running the test suites. Previously  we relied on spawning browser instance for every journey execution which is not required as context already provides enough isolation and each context is fresh (does not share any state or cache) used for each journey runs

+ This removes the time that is required for spawning browser for each run which is in order of ~100ms. 

### Results (Approx)

1. For todos directory with 4 journeys
```
Before - 3600ms
After - 2400ms
```
2. Custom with increased number of journeys to showcase the impact.

```
// 20 journeys
Before - 11200ms
After - 5800ms

// 50 journeys
Before - 28700ms
After - 14800ms
```

### DIFF
--- a/__tests__/core/browser-service.test.ts
+++ b/__tests__/core/browser-service.test.ts
@@ -41,10 +41,10 @@ describe('BrowserService', () => {
   });
 
   it('should create browser pages', async () => {
-    const driver = await Gatherer.setupDriver({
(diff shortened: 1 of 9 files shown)


andrewvc (approved):
LGTM

Issue #266, opened 2021-05-03
Spike: Reduce disk usage of screenshots / network data


andrewvc:
We currently use more disk space than desired to store synthetics screenshots this issue serves to track our efforts to reduce the total amount of disk space.

We have a number of thoughts on how to improve this:

### Lifecycle by status

We currently store the following datasets, `browser`, `browser_network` and `browser_screenshot`. We could migrate to `browser`, `browser_network_{up|down}`, and `browser_screenshot_{up|down}`. With this strategy segmenting indices by result status users could etablish a separate data lifecycle for failures vs. successes. Most users we've talked to would prefer to keep failures around longer than successes. Note that the core `browser` dataset does not have in the example listed above a namespace by up and down status. The rationale for not doing that is that it might be nice to depend on the lightweight core metadata existing and mostly focus on life cycle for the much more resource intensive network data and screenshot use cases.

This approach would require that we buffer the Elasticsearch documents from the run on the box running the synthetics tests. the reason we would need to do this is that we don't know whether the run is a success or failure until it is complete. We could simply do this with temp files on the system.

### Sampled Screenshots

We could add an option to only index successful runs a certain percentage of the time, in a way that is somewhat reminiscent to APM sampling.  Since users mostly care about failures, we would only store the core metadata for each run on most synthetics executions, say only one in 10 runs. Since most of the storage space is taken up by screenshots and network data this would result in a tenfold reduction in storage costs for monitors and mostly succeed. We would make sure to always store data for the run on any state change in other words if it had gone from a success to a failure we would store the first failure event in full.

One challenge with this approach is that we would need UI support for it to make sense, users would need to have a different experience when viewing a sampled run versus Hey fully hydrated run with all screenshots. We could point them at the most recent run with full screenshots if they wanted to see the screenshots.

### Screenshot diffs

[...]

PR #282, opened 2021-05-20
De-dupe screenshots

andrewvc:
WIP for #266 

Beats side is in: https://github.com/elastic/beats/pull/25808

### DIFF
--- a/src/core/runner.ts
+++ b/src/core/runner.ts
@@ -73,8 +73,6 @@ type JourneyContext = BaseContext & {
 export type ScreenshotRef = {
   width: number;
   height: number;
-  blockWidth: number;
-  blockHeight: number;
   blocks: Array<{
     hash: string;
     top: number;
@@ -192,7 +190,7 @@ export default class Runner {
     context: JourneyContext,
     options: RunOptions
   ): Promise<StepResult> {
-    const data: StepResult = {
+    const stepResult: StepResult = {
       status: 'succeeded',
     };
     log(`Runner: start step (${step.name})`);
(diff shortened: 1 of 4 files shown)


src/reporters/json.ts
```diff
@@ -250,28 +263,28 @@ export default class JSONReporter extends BaseReporter {
...
       'step:end',
-      ({
+      async ({
         journey,
         step,
         start,
         end,
         error,
-        screenshot,
+        screenshotRef,
```
vigneshshanmugam, line 283: Curios, what is this one used for? The last reference?

src/reporters/json.ts
```diff
@@ -285,6 +298,7 @@ export default class JSONReporter extends BaseReporter {
             url,
             status,
             metrics,
+            screenshot_ref: screenshotRef,
```
dominiqueclarke, line 301: Where does the screenshot get constructed from the individual blocks? Uptime UI?

andrewvc: That's a good question, maybe there, maybe in Kibana.

Another thought is it may be worth pre-rendering small thumbnails. Assembling 20 of these for a list of thumbnails may be too much. Ideally we'd have a caching CDN

vigneshshanmugam: If kibana in cloud already uses HTTP/2, We can just do individual reqs as we would yield better caching for both screenshots and filmstrips moving forward.

andrewvc: Let's move this out of the `payload` field and into `synthetics.screenshot_ref`. Statistics on these fields could be useful, especially in establishing cache-hit ratio, which will be useful for debugging customer storage issues.

src/core/runner.ts
```diff
@@ -68,12 +70,26 @@ type JourneyContext = BaseContext & {
   pluginManager: PluginManager;
 };
 
+export type ScreenshotRef = {
+  width: number;
+  height: number;
+  blockWidth: number;
```
andrewvc, line 76: We can omit block width / height, it's irrelevant since it's encoded per-block

src/reporters/json.ts
andrewvc: Just made the change ^^

andrewvc:
Closing in favor of https://github.com/elastic/synthetics/pull/290/files

PR #290, opened 2021-05-26
feat: post process screenshots for deduping

### LINKED ISSUE #285: De-duplicate images
Currently screenshots take up too much storage space in synthetics. While there are many approaches to improving this this issue covers our initial approach which will be de-duplication of images, or rather sections of images that repeat.

This is accomplished via a content –addressable scheme of storage where we take each image slice it up into `n` parts, hash each part and use that hash as an elasticsearch document ID. We can then describe each image as a series of image processing operations compositing these parts into a final canvas either on the client or server-side.

See this PR: https://github.com/elastic/synthetics/pull/282

vigneshshanmugam:
+ fix #285 
+ Captures screenshot for each step and writes them to the cached `.synthetics/<pid>/screenshots` folder in JSON format with `step` and `screenshot` info. The pid stuff is to make sure each process (workers in future for leveraging parallelism) is writing screenshots to individual folders to avoid mutation. 
+ Written screenshots are processed in to 8 blocks which results in 64 images blobs for `1280*720` viewport (Default) and wrote to heartbeat as `screenshot/block` and the whole screenshot reference is written as `step/screenshot_reference` which can be queried from the Uptime UI to construct all the images from the hash of individual block. 

```json
// screenshot block
{
  "type": "screenshot/block",
  "_id": "20f2cdc9e8302e94ff8d3dc09d3c359683cbcec5",
  "@timestamp": 1622076105043985.8,
  "blob": "UklGRlAAAABXRUJQVlA4IEQAAAAwBgCdASqgAFoAPm02mUmkIyKhIIgAgA2JaW7hdJAAT22IvEFRz2xF4gqOe2IvEFRz2xF4gqOe2ItgAP7/dlECgAAAAA==",
  "blob_mime": "image/webp",
  "package_version": "1.0.0-beta.2"
}
// screenshot ref

{
  "type": "step/screenshot_ref",
  "@timestamp": 1600300800000000,
  "journey": {
    "name": "j1",
    "id": "j1"
  },
  "step": {
    "name": "launch app",
    "index": 1
  },
  "root_fields": {
    "screenshot_ref": {
      "width": 1280,
      "height": 720,
      "blocks": [
        {
          "hash": "aee19c1f32131ed90e6dd34074a6e220ad97baf6",
          "top": 0,
          "left": 0,
          "width": 160,
          "height": 90
        },
        {
          "hash": "aee19c1f32131ed90e6dd34074a6e220ad97baf6",
          "top": 0,
          "left": 160,
          "width": 160,
          "height": 90
        }
      ]
    }
  },
  "package_version": "0.0.1"
}
```


src/reporters/json.ts
```diff
@@ -219,6 +250,69 @@ function stepInfo(
...
+  };
+  const blocks: Array<ScreenshotBlob> = [];
+
+  for (let row = 0; row < divisions; row++) {
+    const top = row * blockHeight;
+    for (let col = 0; col < divisions; col++) {
+      const left = col * blockWidth;
+      const buf = await img
+        .extract({ top, left, width: blockWidth, height: blockHeight })
+        .toFormat(sharp.format.webp)
```
vigneshshanmugam, line 282: self note: check the size of individual blocks using different format.

vigneshshanmugam: webp even though smaller than png by ~20ms average would yield more savings. Will leave it as webp for now. 

Processing 50 images
webp - 14.36 seconds
png - 13.119 seconds

vigneshshanmugam: Benchmarks script to play with
```
!(async () => {
  console.time('screenshots');
  for (let i = 0; i < 50; i++) {
    await gatherScreenshots(
      'path/to/.synthetics/pid/screenshots'
    );
  }
  console.timeEnd('screenshots');
})();
```

src/reporters/json.ts
```diff
@@ -292,7 +379,7 @@ export default class JSONReporter extends BaseReporter {
 
     this.runner.on(
       'journey:end',
-      ({
+      async ({
```
vigneshshanmugam, line 382: move to sync way, otherwise we would have hard time figuring out when to exactly close the underlying stream. 

Once this is sorted out, I will add tests.

src/reporters/json.ts
```diff
@@ -219,6 +250,81 @@ function stepInfo(
...
+      });
+      /**
+       * We dont write the width, height of individual blocks on the
+       * reference as we use similar sized blocks for each extraction,
+       * we would need to send the width and height here if we decide to
+       * go with dynamic block extraction.
+       */
+      reference.blocks.push({
+        hash,
+        top,
```
vigneshshanmugam, line 298: Since we dont have dynamic extraction at this point, we could save some more size by calculating the `top`, `left` using the number of blocks and fixed block width and height. 

num of blocks = divisions ^ 2
calculating top and left the same way as we do here for extraction, we do the same for composition on the UI. 

Thanks to @axw for the idea.

src/reporters/base.ts
```diff
@@ -66,26 +66,7 @@ export default class BaseReporter {
...
-    } else {
-      /**
-       * If the user has passed a custom FD we don't close the FD, but we do flush it
-       * to give them more control. This is important because FDs should only be closed
-       * once, and the primary use case for custom FDs is being called by heartbeat, which
-       * closes the FD after the process exits
-       */
-      this.stream.flush();
-    }
+    this.runner.on('end', () => this.stream.flush());
```
vigneshshanmugam, line 69: We don't need to take care of closing the stream as the process exists automatically which closes the underlying stream once program runs to completion. This is better than figuring out when to close and prematurely closing it than leaving it which does not harm.

src/reporters/json.ts
```diff
@@ -302,6 +395,32 @@ export default class JSONReporter extends BaseReporter {
...
+                blob_mime,
+              });
+            }
+
+            this.writeJSON({
+              type: 'step/screenshot_reference',
+              journey,
+              step,
+              payload: {
+                screenshot_ref: reference,
```
andrewvc, line 418: Let's pull this out of payload and put it at the top level. We want to index some of the values here, they'll be useful to defer debugging storage issues since we can run queries to determine the reuse of screenshot blocks. 

You'll just need to add a new root field `"synthetics.screenshot_ref"`  with the ref

src/reporters/json.ts
```diff
@@ -71,6 +84,30 @@ type OutputFields = {
...
 
+type ScreenshotBlob = {
+  blob: string;
+  id: string;
+};
+
+type ScreenshotReference = {
+  width: number;
+  height: number;
+  blockWidth: number;
```
andrewvc, line 95: Can we remove the `blockWidth` and `blockHeight` attributes here and put them on a per block level in the array below. That way in the future we can have irregularly size blocks and do sophisticated compositing on the client side. We can have the agent to generate whatever arrangement of blocks makes sense going far beyond a simple grid.

src/reporters/json.ts
andrewvc: Compared to the size the images storing the dimensions and coordinates is in significant, especially if we don't index them which we don't have to. It's more important I think to have a future proof format here then to save on a few fields. That way we do want a more sophisticated way of slicing up images the client will not have to change. I left a similar note in the type definition elsewhere

src/reporters/json.ts
```diff
@@ -219,6 +256,81 @@ function stepInfo(
...
+  for (let row = 0; row < divisions; row++) {
+    const top = row * blockHeight;
+    for (let col = 0; col < divisions; col++) {
+      const left = col * blockWidth;
+      const buf = await img
+        .extract({ top, left, width: blockWidth, height: blockHeight })
+        .webp({ quality: 80 })
+        .toBuffer();
+
+      const hash = createHash('sha1').update(buf).digest('hex');
```
andrewvc, line 291: Are we certain that creating the webP image is deterministic? I'm wondering if there's any metadata encoded in there that would throw the hash off. It feels like it might be safer to hash the raw pixel buffer. However that might be slower because we might need to extract that buffer before encoding.

I think the way we can defend against this is to have a test where we validate that encoding the same screenshot twice yields identical blocks with identical hashes. It looks like the default behavior is no metadata, but it'd be good to confirm this: https://sharp.pixelplumbing.com/api-output#withmetadata

src/reporters/json.ts
```diff
@@ -252,26 +364,7 @@ export default class JSONReporter extends BaseReporter {
...
-          this.writeJSON({
-            type: 'step/screenshot',
-            journey,
-            step,
-            blob: screenshot,
-            blob_mime: 'image/jpeg',
-          });
-        }
+      ({ journey, step, start, end, error, url, status, metrics }) => {
         this.writeJSON({
```
andrewvc, line 368: I'm wondering if we should add a new field to the payload `screenshot_type: "ref"` to make it easier on the client to know how to handle the screenshot. WDYT @justinkambic ?

andrewvc (changes requested):
Love the improvements here left some comments on some ways it could work even better

src/reporters/json.ts
vigneshshanmugam: I thought about a bit around future proofing the code on the client side. If we are not going to use `canvas` on the Kibana UI for composing and can use `sharp` or image processing library, we don't even need to send these information except for the hash.

src/reporters/json.ts
vigneshshanmugam: Sounds good, will do. We don't want it to be in `root_fields`?

src/reporters/json.ts
vigneshshanmugam: Good point, adding a test for the same.

vigneshshanmugam: Added the test, metadata is not added by default and test confirms it.

src/reporters/json.ts
vigneshshanmugam: I had changed as per your comment, but lets come back to it based on our solution on the UI side.

andrewvc: Yeah, this isn't implementation specific, but rather I'd like us to support non-grid layouts in the future. Rather than it being a list of blocks of a hard coded size let's think of it as a list of compositing operations. It's far more powerful, isn't extra work, and gives us more flexibility

andrewvc (changes requested):
This PR is essentially LGTM but I'm leaving it in request changes state because we can't merge it as is without breaking things for existing users. If we were to merge and release this today screenshots will not work with current versions of kibana. We really need a system of feature flags or something similar to heartbeat to determine which features are enabled. Heartbeat at least follows the stack versioning, and as a result can inform the code here of how to act.

I think we will also need to support the old way of screenshots working based on whatever system of flags or capabilities we come up with. I've opened https://github.com/elastic/synthetics/issues/292 to track

andrewvc (changes requested):
There's a bug somewhere in this PR that affects handling of erroring test-cases. Running the following: `cat t.js | node dist/cli.js  --network --json --inline --screenshots | jq .type` will fail early, not outputting `journey/end` and other events, where the contents of `t.js` are:

```javascript
        step("load homepage", async () => {
        });
        step("hover over products menu", async () => {
            await page.hover('css=[data-nav-item=products]');
        });
        step("failme", async () => {
            await page.hhover('css=[data-nav-item=products]');
        });
```

vigneshshanmugam:
@andrewvc Could be this https://github.com/elastic/synthetics/pull/290#discussion_r640300037, Will check tomorrow.

andrewvc:
Not sure as to the root cause

vigneshshanmugam:
Fixed the issue - We do async processing of the events in the microtasks which are being cut of by the CLI node process. Now we wait for the ack from the reporter once its written to the file descriptor before closing the process - https://github.com/elastic/synthetics/pull/290/commits/36a42e54ab81603287435bbe7f522cd94a6757fd

src/reporters/json.ts
```diff
@@ -226,6 +263,84 @@ function stepInfo(
...
+  };
+  const blocks: Array<ScreenshotBlob> = [];
+
+  for (let row = 0; row < divisions; row++) {
+    const top = row * blockHeight;
+    for (let col = 0; col < divisions; col++) {
+      const left = col * blockWidth;
+      const buf = await img
+        .extract({ top, left, width: blockWidth, height: blockHeight })
+        .jpeg()
```
vigneshshanmugam, line 293: changed to jpeg for now, we can play with `webp` once we decide a solution in the Kibana.

src/common_types.ts
```diff
@@ -119,6 +119,7 @@ export type CliArgs = {
   outfd?: number;
   headless?: boolean;
   screenshots?: boolean;
+  ssblocks?: boolean;
```
vigneshshanmugam, line 122: @andrewvc If you have any better name for enabling the screenshot block capability.

src/reporters/json.ts
vigneshshanmugam: @justinkambic FYI

__tests__/reporters/json.test.ts
```diff
@@ -207,22 +211,70 @@ describe('json reporter', () => {
...
-    const data = 'aaaaaaaaaaa';
-    runner.emit('step:end', {
+  it('return empty when dir doesnt exists', async () => {
+    const nonExistDir = join(FIXTURES_DIR, 'blah');
+    const callback = jest.fn();
+    await gatherScreenshots(nonExistDir, callback);
+    expect(callback).not.toHaveBeenCalled();
+  });
+
+  it('idempotent on constructing screenshots blocks', async () => {
```
andrewvc, line 221: Love this!

andrewvc (approved):
LGTM

PR #291, opened 2021-05-27
chore: inline example checks the cookie banner

jahtalab:


paulb-elastic:
@jahtalab this is also in https://github.com/elastic/synthetics/blob/master/examples/docker/heartbeat.docker.yml#L14.

If this solution works (worth a quick check by folks in the US where it seems there is no cookie banner), then it should also be applied in the above example.

jahtalab:
Thanks @paulb-elastic , I'll apply the the fix in the docker file as well.

Re. when there's no banner, I've tested it locally and it works (but a quick test from US folks is helpful too)

vigneshshanmugam (approved):
Tested locally, LGTM

PR #294, opened 2021-06-01
feat: add support for rich events flag

### LINKED ISSUE #289: Add `--rich-metadata` flag to enable all rich metrics
We currently have a number of flags that are used by heartbeat in conjunction with the `– – json` output, such as `--screenshots` and `--network`. As we add new flags we are creating a compatibility issue between heartbeat and the synthetics agent. Newer versions of heartbeat must add additional flags to gain from their functionality, thus breaking compatibility with older suites using older versions of the synthetics agent which do not support that flag.

This issue proposes adding a new meta-tag `--rich-metrics` that can stably be used by heartbeat which turns on the relevant flags in the synthetics agent. An additional benefit here is that adding new flags to the synthetics agent won't require adding them to the beats code base.

vigneshshanmugam:
+ fix #289 
+ Mimics the current heartbeat behaviour by introducing the `--rich-events` flag that will enable `Screenshots, Network, Default to JSON reporter`.

### DIFF
--- a/__tests__/cli.test.ts
+++ b/__tests__/cli.test.ts
@@ -52,6 +52,20 @@ describe('CLI', () => {
     expect(await cli.exitCode).toBe(0);
   });
 
+  it('enables rich events on `--rich-events` flag', async () => {
+    const cli = new CLIMock([
+      join(FIXTURES_DIR, 'fake.journey.ts'),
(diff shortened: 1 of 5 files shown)


andrewvc (approved):
LGTM! Very simple / clean

PR #295, opened 2021-06-02
feat: add support for capabilities flag

### LINKED ISSUE #292: Add flags/capabilities system
We need a system whereby heartbeat can let the synthetics library know how wants to behave at a feature level. As we develop new features they may require new functionality in kibana or elasticsearch or heartbeat to function correctly. The problem is that the synthetics library versioning is completely independent of the rest of the stack. We could solve a number of ways, but one way that does not work is adding additional CLI flags. The main reason that is not a good idea is that a newer version of heartbeat would not work with an older version of the synthetics library that didn't support a newer flag.

I can think of two ways we can do this and I'm not certain which one is the best:

1. Add a `--stack-version=X.Y.Z` flag to synthetics, which can be used to determine which features are on/off
2. Add a `--capability screenshot_refs --capability somenewcthing` flag, to enable new features on a one-by-one basis

The advantage of the first approach is that no updates to the heartbeat code have to happen to add a new capability. The advantage of the second approach is that it's quite a bit cleaner and it is more related CLI testing.

We could also do a combination of approaches where setting the stack version sets a certain set of capabilities.

vigneshshanmugam:
+ fix #292 
+ Allows for a way to pass an variadic args to `--capability` flag that controls the features inside the agent. 
```
npx @elastic/synthetics examples/todos --capability metrics trace --capability filmstrips
```
+ Also deprecated the previous `--trace, --metrics, --filmstrips` flag which are not required anymore and not used by heartbeat currently.

### DIFF
--- a/src/parse_args.ts
+++ b/src/parse_args.ts
@@ -29,7 +29,6 @@ import { reporters } from './reporters';
 
 /* eslint-disable-next-line @typescript-eslint/no-var-requires */
 const { name, version } = require('../package.json');
-const allowedCapabilities = ['trace', 'filmstrips', 'metrics'];
 
 program
   .name(`npx ${name}`)
@@ -56,11 +55,9 @@ program
   .option('--no-headless', 'run browser in headful mode')
   .option('--sandbox', 'enable chromium sandboxing')
   .option('--rich-events', 'Mimics a heartbeat run')
-  .addOption(
-    new Option(
-      '--capabilities <features...>',
-      'Enable capabiltiites through feature flags'
-    ).choices(allowedCapabilities)
+  .option(
+    '--capabilities <features...>',
+    'Enable capabiltiites through feature flags'
   )
   .option('--screenshots', 'take screenshot for each step')
   .option('--network', 'capture network information for all journeys')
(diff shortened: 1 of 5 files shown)


src/parse_args.ts
```diff
@@ -23,16 +23,14 @@
...
 import { CliArgs } from './common_types';
 import { reporters } from './reporters';
 
-const availableReporters = Object.keys(reporters)
-  .map(r => String(r))
-  .join();
-
 /* eslint-disable-next-line @typescript-eslint/no-var-requires */
 const { name, version } = require('../package.json');
+const allowedCapabilities = ['trace', 'filmstrips', 'metrics'];
```
andrewvc, line 32: I'm not sure it makes sense to have a strict list of capabilities. If we do that we may as well just have the individual flags which are easier to understand.

The point of capabilities was that these were capabilities you could ask for but if not present would not cause an error.

This is less relevant given our plan to have the `--rich-events` flag, which solves the same problem from heartbeat'- perspective. However, I still think the capabilities flag is a nice thing to have so long as we reserve it for things that are not strict requirements. We could consider adding an info or warning message for missing capabilities

andrewvc (changes requested):

src/parse_args.ts
vigneshshanmugam: ++, My mistake and feature flags shouldn't be a restrictive list. 

> We could consider adding an info or warning message for missing capabilities

I like that idea. Let me incorporate your feedback.

src/parse_args.ts
```diff
@@ -58,18 +56,11 @@ program
   .option('--sandbox', 'enable chromium sandboxing')
   .option('--rich-events', 'Mimics a heartbeat run')
   .option(
-    '--ws-endpoint <endpoint>',
-    'Browser WebSocket endpoint to connect to'
-  )
-  .option(
-    '--pause-on-error',
-    'pause on error until a keypress is made in the console. Useful during development'
+    '--capabilities <features...>',
```
andrewvc, line 59: One small change, we should name this `--capability` I think, because as a plural it sounds like it maybe takes a comma separated list rather than multiple uses

andrewvc (changes requested):
Needs one small tweak

src/parse_args.ts
vigneshshanmugam: Good call

andrewvc (approved):
LGTM

PR #296, opened 2021-06-03
chore: use different selector to pick banner

vigneshshanmugam:
+ When the element is null, the returned collection will not be iterable. So we use the single element instead of using the list.


paulb-elastic (commented):
This has fixed the issue I was seeing with `TypeError: (intermediate value) is not iterable`
(I'll leave it for review by others)

vigneshshanmugam:
Thanks for testing Paul. I ll go ahead and merge it then.

PR #298, opened 2021-06-03
fix: prioritize suite params over config

### LINKED ISSUE #297: Config files break `--suite-params`
When a config file is in use passing sweet parameters does not override any of the values. This is a bug in should be fixed suite parameter should always take precedence over config's

vigneshshanmugam:
+ fix #297 
+ We now deep merge the params passed from both config and suite params and always prioritize suite params over config params when a key is specified in both.

### DIFF
--- a/__tests__/cli.test.ts
+++ b/__tests__/cli.test.ts
@@ -83,6 +83,23 @@ describe('CLI', () => {
     });
   });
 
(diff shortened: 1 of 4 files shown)


andrewvc (approved):
LGTM

PR #300, opened 2021-06-05
feat: add filtering by tags and names

### LINKED ISSUE #26: Add tags, filtering

vigneshshanmugam:
+ fix #26 
+ Added support for glob pattern matching for both journey name and tags. 
+ Names and tags can be matched with multiple journeys that has the same name and tags spawned across multiple files. 
```
elastic-synthetics examples/todos/* --match 'check*' 
// will match all the journey that has `check` in their name and tags. 
```
This aligns with `RegExp` or glob matching found in major libraries like jest, folio, mocha etc and we don't support variadiac options for name matching as majority of the cases can be solved with a mix of both.

+ Tags matching which is a variadic option meaning you can pass multiple options in the CLI to match a group of journeys.
```ts
// will run the journeys that are tagged with the respective teams in the kibana repo. 
elastic-synthetics <dir> --tags 'kibana:uptime*'  'kibana:apm*'

journey({ name: "elastic uptime app", tags: ["kibana:uptime", "kibana:apm" }, () => {})
```

**Note: `match` pattern that matches both name and tags, `tags` pattern that matches only tags**

### DIFF
--- a/src/dsl/journey.ts
+++ b/src/dsl/journey.ts
@@ -24,14 +24,14 @@
  */
 
 import { Browser, Page, BrowserContext, CDPSession } from 'playwright-chromium';
-import { contains, isMatch } from 'micromatch';
+import micromatch, { isMatch } from 'micromatch';
 import { Step } from './step';
 import { VoidCallback, HooksCallback, Params } from '../common_types';
 
 export type JourneyOptions = {
   name: string;
   id?: string;
-  tag?: string;
+  tags?: string[];
 };
 
 type HookType = 'before' | 'after';
@@ -47,7 +47,7 @@ export type JourneyCallback = (options: {
 export class Journey {
   name: string;
   id?: string;
-  tag?: string;
+  tags?: string[];
   callback: JourneyCallback;
   steps: Step[] = [];
   hooks: Hooks = { before: [], after: [] };
@@ -55,7 +55,7 @@ export class Journey {
   constructor(options: JourneyOptions, callback: JourneyCallback) {
     this.name = options.name;
     this.id = options.id;
-    this.tag = options.tag;
+    this.tags = options.tags;
     this.callback = callback;
   }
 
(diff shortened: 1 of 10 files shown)


vigneshshanmugam:
@spalger Curios to hear if these options would work at scale for Kibana apps?

src/parse_args.ts
```diff
@@ -65,7 +65,14 @@ program
     '--dry-run',
     "don't actually execute anything, report only registered journeys"
   )
-  .option('--journey-name <name>', 'only run the journey with the given name')
+  .option(
+    '--match <name>',
```
vigneshshanmugam, line 69: `--grep`, `--match` i like both

spalger: I feel like maybe we should match names and tags with `--match`, a flag like `--tags` or `--name` feels more like a limited version of `--match`...

spalger: `--grep` has more historical meaning in my mind, and would be used to grep test names specifically.

src/parse_args.ts
```diff
@@ -65,7 +65,14 @@ program
...
+    'run only journeys with a name that matches the glob'
+  )
+  .option(
+    '--tags <name...>',
```
spalger, line 73: Since we're using globs do we support negation too?

src/dsl/journey.ts
```diff
@@ -45,13 +47,15 @@ export type JourneyCallback = (options: {
 export class Journey {
   name: string;
   id?: string;
+  tag?: string;
```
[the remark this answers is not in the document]
vigneshshanmugam: Hmmm, I was thinking if it's worth supporting. Do you think tests from multiple team coexists? I dont have a strong opposition either to support it. Went for a singular as it was more cleaner.

src/parse_args.ts
vigneshshanmugam: yes it should, Let me add a test here.

src/parse_args.ts
vigneshshanmugam: Hmmm, I like that idea. We can stick to match and pass the glob to do global name and tag matching. @andrewvc WDYT?

src/dsl/journey.ts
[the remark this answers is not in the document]
spalger: I don't expect tests from multiple teams to coexist but we use tags for several purposes currently, like labeling tests that don't work correctly with certain backends (`skipCloud`, `skipFirefox`) and to organize tests into groups (which we don't expect to bring forward). I think the idea that there are multiple dimensions a single journey could be tagged with makes sense to me.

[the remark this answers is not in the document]
vigneshshanmugam: Hmmm the more I think about it now, Yeah having multiple tags makes sense. Will add it as well.

src/parse_args.ts
andrewvc: I'm OK with matching both. I guess regexps would be for more precise name-only matching?

src/dsl/journey.ts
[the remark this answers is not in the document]
andrewvc: Big +1 on multiple tags

src/parse_args.ts
vigneshshanmugam: Had a offline discussion, We will go with this. 
```
- match pattern that matches both name and tags
- tags pattern that matches only tags
```

PR #304, opened 2021-06-08
feat: add step duration to step/end

### LINKED ISSUE #302: Add step duration as an indexed field
We currently index the step duration of the field on the payload object. Users cannot search this data or make custom kibana visualizations with it. It would be useful to make step timing charts not only in uptime, but also in kibana dashboards.

This issue proposes we create a new field `synthetics.step.duration.us` which measures the step execution time in milliseconds. We can use the `root_fields`  JSON to add this field.

I had discussed with @vigneshshanmugam the possibility of making a new field for journey durations as well, however the existing `monitor.duration.us` field covers this. I'd rather not introduce a second field with the same information in milliseconds at this point since that may be confusing.

vigneshshanmugam:
+ fix #302 
+ Adds the step duration info in `us` for all `step/end` event under step object.
+ Gets rid of start/end in payload as we rely on step info for querying all network events.


src/reporters/json.ts
```diff
@@ -380,11 +380,14 @@ export default class JSONReporter extends BaseReporter {
           journey,
           step,
           url,
+          root_fields: {
+            duration: {
```
andrewvc, line 384: Can we nest this one level deeper in `step.duration.us`? That will make it more clear to users.

andrewvc (changes requested):

src/reporters/json.ts
vigneshshanmugam: Hmmm thinking a bit more, can we keep it in top level and not add it under root fields if thats the case.

andrewvc: Sure, we'll need to add it to `synthtypes.go` in heartbeat, but that's not a big deal.

andrewvc (changes requested):
I'm pretty sure the reported numbers are not micros. In one test run I got  0.05044099999940954 as a value.

Invoked with ` npm run build; node dist/cli.js examples/todos/ --rich-events | grep step/end | jq .step` 

I'm pretty sure the values are actually fractional seconds

vigneshshanmugam:
@andrewvc 🤦🏽 moment here, Yes it is indeed denoting seconds - https://github.com/elastic/synthetics/blob/e0f7638c6da7e4ed44bbe913ff354ac8bdb145a7/src/reporters/base.ts#L86.

andrewvc (approved):
LGTM

PR #305, opened 2021-06-09
feat: add expect assertion library

### LINKED ISSUE #201: Add improved assertion library from folio
At the moment we're unopinionated about assertions, but, this is something we should probably bake in. One thing to consider would be Microsoft's [folio/expect](https://github.com/microsoft/folio/blob/master/src/expect.ts). 

We would alias it, so users could import it like `import { expect } from '@elastic/synthetics'`

As-is, we rely on `assert` in examples, which is too minimal. While users can use any library, it's good to have one sanctioned one, especially for inline examples (where we should just provide `expect`.

@vigneshshanmugam @jahtalab @paulb-elastic would love your thoughts here

vigneshshanmugam:
+ fix #201 
+ Uses the expect library with custom Matcher interface as some of the expect types are implemented by `jest` like Spy and Snapshot matcher. 
+ Types are namespaced under `synthetics` in global to allow `expect.extend` functionality work with Typescript without throwing any typings error.  Look at https://github.com/elastic/synthetics/blob/fa18f519d4d8488d49f5883b04d601f281aabdcc/__tests__/fixtures/expect.journey.ts for more details. 
+ Test coverage is done for both expect library and also for types with extensions.

### DIFF
--- a/src/core/expect.ts
+++ b/src/core/expect.ts
@@ -23,7 +23,7 @@
  *
  */
 
-import expectLibrary from 'expect';
+import expectLib from 'expect';
 import type {
   AsymmetricMatcher,
   MatcherState,
@@ -184,4 +184,4 @@ declare global {
   }
 }
 
-export const expect: Expect = expectLibrary;
+export const expect: Expect = expectLib;
--- a/src/core/expect.ts
+++ b/src/core/expect.ts
@@ -23,17 +23,16 @@
  *
  */
 
-import expectLib from 'expect';
-import type expectTypes from 'expect';
+import expectLibrary from 'expect';
 import type {
-  ExpectedAssertionsErrors,
   AsymmetricMatcher,
   MatcherState,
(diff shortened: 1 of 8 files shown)


src/core/expect.ts
```diff
@@ -0,0 +1,61 @@
...
+  stringContaining(expected: string): AsymmetricMatcher;
+  stringMatching(expected: string | RegExp): AsymmetricMatcher;
+};
+
+/**
+ * Override matchers from expect in our synthetic namespace to
+ * reduuce support for spy and snapshot matchers
+ */
+export interface Matchers<R> extends expectTypes.Matchers<R> {
+  not: Matchers<R>;
```
vigneshshanmugam, line 56: Need to pick only the list that doesn't have spy matchers and snapshot matchers and override the `expect namespace`.

andrewvc (approved):
LGTM

Issue #301, opened 2021-06-07
core vitals: update CLS to new definition


vigneshshanmugam:
+ Update CLS to the new variant - https://web.dev/evolving-cls/ which uses the windowed approach of `max-session-gap1s-limit5s `. 

+ As the UI does not utilize the previous CLS values, we will go ahead with not suporting the previous value which only calculated the CLS for main frames and instead report only the windowed CLS value for all frames under `cumulativeLayoutshift` name.

PR #306, opened 2021-06-10
fix: perf relative traces and values

### LINKED ISSUE #303: Improve mapping for additional metrics
For newly added metrics in https://github.com/elastic/synthetics/pull/194

If you want to build visualisation for example on LCP.

First you will have to do filter on `browser.experience.name` and then use value from `browser.experience.name` , so they metric type is stored as value instead of field. 

I think we should add each metric as new field. Since this will make querying es data easier.

vigneshshanmugam:
+ fix #303 
+ Relative traces and metrics are grouped under two keys now to provide clear distinction. 
+ Grouped all traces under `browser.relative_trace` to keep the mapping similar and these traces are mark or measure events with `start` and `end` time denoting the time when particular event (LCP, FCP, etc) occurred during the waterfall. 
+ All distinct values like FCP, LCP, CLS, etc are grouped under `browser.experience.<metric-name>` to keep the querying easier in the UI and useful for displaying values in step detail page. 

### Relative trace events
```json
{
  "type": "journey/metrics",
  "@timestamp": 1623300251603855.2,
  "journey": {
    "name": "check if title is present",
    "id": "check if title is present"
  },
  "root_fields": {
    "browser": {
      "relative_trace": {
        "name": "firstContentfulPaint",
        "type": "mark",
        "start": {
          "us": 142089104934
        }
      }
    },
  }
}

{
  "type": "journey/metrics",
  "journey": {
    "name": "inline",
    "id": "inline"
  },
  "root_fields": {
    "browser": {
      "relative_trace": {
        "name": "widget_core:widget_loaded",
        "type": "measure",
        "start": {
          "us": 142095075042
        },
        "end": {
          "us": 142095312098
        }
      }
    },
  }
}

```

### Experience metrics
```json
{
  "type": "journey/metrics",
  "@timestamp": 1623300251603949.5,
  "journey": {
    "name": "check if title is present",
    "id": "check if title is present"
  },
  "root_fields": {
    "browser": {
      "experience": {
        "cls": 0,
        "fcp": {
          "us": 189382
        },
        "lcp": {
          "us": 189382
        },
        "dcl": {
          "us": 209052
        },
        "load": {
          "us": 209115
        }
      }
    },
}
```


examples/todos/basic.journey.ts
```diff
@@ -14,17 +14,17 @@ journey('check if title is present', ({ page }) => {
   });
 });
 
-journey('check if input placeholder is correct', ({ page }) => {
-  step('launch app', async () => {
-    const path = 'file://' + join(__dirname, 'app', 'index.html');
-    await page.goto(path);
-  });
+// journey('check if input placeholder is correct', ({ page }) => {
```
vigneshshanmugam, line 17: need to revert.

src/sdk/trace-metrics.ts
```diff
@@ -147,19 +170,12 @@ export class CumulativeLayoutShift {
     return { score: clsScore, event: finalLayoutShiftEvent };
   }
 
-  static compute(trace: LHTrace): LayoutShift {
+  static compute(trace: LHTrace) {
     const events = trace.mainThreadEvents;
-    const { score, event } = this.computeCLSValue(events);
```
vigneshshanmugam, line 152: all CLS relative trace events will be added as part of the work in #301

src/common_types.ts
```diff
@@ -38,29 +38,26 @@ export type StatusValue = 'succeeded' | 'failed' | 'skipped';
...
+  type: string;
   end?: number;
 };
 
-export type UserTiming = TraceOutput & {
-  name: string;
-  type: string;
-  duration?: number;
+export type PerfMetrics = {
+  fcp: number;
```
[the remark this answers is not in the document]
vigneshshanmugam: I thought about that, but keeping it sync with RUM agent intake `transaction.experience.fcp` vs `browser.experience.fcp`. Thought it would be cleaner in the UI.

src/sdk/trace-metrics.ts
```diff
@@ -78,11 +78,20 @@ export class UserTimings {
...
+
+  static buildMetric(event: TraceEvent, shortName?: string, name?: string) {
     if (!event) {
       return;
     }
-
+    /**
+     * Calculate metrics relative to the origin event
+     */
+    if (this.navigationStart) {
```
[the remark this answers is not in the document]
vigneshshanmugam: Good feedback, will address it.

src/common_types.ts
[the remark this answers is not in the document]
andrewvc: I think that now would be an OK time to add the units in. Maybe this is a case where synthetics leads and RUM follows? Elsewhere in synthetics we use `.us` so it's a question of where we keep consistency.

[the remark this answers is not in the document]
justinkambic: I'm used to having a `*.us` field from beats; there isn't going to be any annotation when I'm looking at documents in an Elasticsearch response, so that explicitness leaves no doubt.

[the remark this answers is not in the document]
shahzad31: I think it's always helpful to have units

src/helpers.ts
```diff
@@ -66,12 +65,12 @@ export function getMonotonicTime() {
 /**
  * Converts the trace events timestamp field from the
  * format -  hrTime[0] * 1e6 + Math.round(hrTime[1] / 1000) to
- * the internal timestamp similar to other event types (journey, step, etc)
+ * the our internal monotonic clock timestamp similar to other event types (journey, step, etc)
```
andrewvc, line 68: Can we be clear about the unit here? Is it Micros? I'd add it to the function name `convertToMonotonicTimeMicros`

I will say, one thing I love about golang is the `time.Duration` type (everything stored as nanos). Makes things clear!

src/reporters/json.ts
```diff
@@ -472,9 +472,9 @@ export default class JSONReporter extends BaseReporter {
             this.writeJSON({
               type: 'journey/filmstrips',
               journey,
-              payload: {
-                index,
-                start: strip.start,
+              payload: { index },
+              root_fields: {
+                browser: { relative_trace: { start: strip.start } },
```
andrewvc, line 477: Can we add units to `start` as well?

andrewvc (changes requested):
Found some more missing units, otherwise got. Looking better!

src/reporters/json.ts
vigneshshanmugam: They are already relative right? Do you have any suggestion?

vigneshshanmugam: start.time.<unit>?

andrewvc: Yeah, that's what I was thinking with the `us`

vigneshshanmugam: Done.

src/helpers.ts
vigneshshanmugam: I have updated the comments and also changed the method signatures where required and also kept everything in us without converting them back.

andrewvc (approved):
LGTM, so much cleaner with easy to read units :)

PR #307, opened 2021-06-10
feat: target to build a standalone package

kuisathaverat:
In some cases, it is needed a standalone package to install in offline systems, this new target `npm run standalone-package` would generate an `elastic-synthetics-VERSION.tgz` tar file that contains the synthetics package and all its dependencies. This package can be installed in offline systems with the following command `npm install -g --offline elastic-synthetics-VERSION.tgz`


cachedout (approved):

package.json
```diff
@@ -11,6 +11,7 @@
     "watch": "tsc -w",
     "lint": "eslint . --rulesdir utils/eslint-rules",
     "lint:fix": "npm run lint -- --fix",
+    "standalone-package": "npm run clean && npm install && npm run build && npm-pack-all",
```
vigneshshanmugam, line 14: nit: Also, always good to keep npm install out of the npm scripts 

```suggestion
    "pkg": "npm run clean && npm run build && npm-pack-all",
```

vigneshshanmugam (approved):

PR #309, opened 2021-06-14
feat: update CLS to new windowed approach

### LINKED ISSUE #301: core vitals: update CLS to new definition

vigneshshanmugam:
+ fix #301 
+ This PR updates the CLS to new definition which captures CLS based on max session window of 5 seconds with 1 sec gaps. More details - https://web.dev/evolving-cls/
+ Keeps in sync with Lighthouse, changes includes only the output format and relative traces with updated score for each metric to include in the waterfall. 
+ We capture only the `cumulative layout shift for all frames` and do not capture the old metric which was the total cumulative layout shift which we dont surface in the UI before.


src/sdk/trace-metrics.ts
```diff
@@ -153,36 +153,85 @@ export class ExperienceMetrics {
...
+    return layoutShiftEvents;
+  }
+
+  /**
+   * Calculate cumulative layout shift per sesion window where each session
+   * windows lasts for 5 seconds since the last layoutShift event and has 1 second
+   * gap between them. Return the maximum score between all windows.
+   * More details - https://web.dev/evolving-cls/
+   */
+  static calculateScore(layoutShiftEvents) {
```
andrewvc, line 195: Could use a good set of unit tests for different window scenarios

src/sdk/trace-metrics.ts
```diff
@@ -153,36 +153,85 @@ export class ExperienceMetrics {
...
+    const gapMicroseconds = 1_000_000;
+    const limitMicroseconds = 5_000_000;
+    let maxScore = 0;
+    let currentScore = 0;
+    let firstTimestamp = Number.NEGATIVE_INFINITY;
+    let prevTimestamp = Number.NEGATIVE_INFINITY;
+
+    for (const event of layoutShiftEvents) {
+      if (
+        event.ts - firstTimestamp > limitMicroseconds ||
```
andrewvc, line 205: The logic here is tricky, and comments always help I think. It's doing two things in one pass, calculating the current window, and figuring out which window is to be used.  The code would be clearer IMHO if this was broken down more. Maybe even using map/reduce (esp. since this isn't perf critical code). It might be clearer to have two distinct phases, creating the windows, then selecting one, rather than doing it all-in-one.

vigneshshanmugam: Kept in sync with lighthouse so we can use keep it more or less same and get rid of our code at some point. I can add more comments if that helps.

src/sdk/trace-metrics.ts
vigneshshanmugam: Will add some simulated tests.

src/sdk/trace-metrics.ts
vigneshshanmugam: Added detailed comments, Should be good to reason about it now. I don't think we need to create session windows manually and use map-reduce. The current code feels cleaner to me and can accomplish it in a single pass.

__tests__/sdk/trace-metrics.test.ts
```diff
@@ -174,6 +174,32 @@ describe('Trace metrics', () => {
     });
   });
 
+  it('calculate cls score with simulated sessions', () => {
+    // events with single session
```
andrewvc, line 178: Love these new tests

andrewvc (approved):
LGTM

PR #311, opened 2021-06-17
feat: add support for screenshot toggle

### LINKED ISSUE #308: Add screenshot options
`--screenshots` should take an option of: `off`, `on`, `only-on-failure` or `space-efficient`.

`space-efficient` be the default in heartbeat and would retain just the last step's screenshot on successful runs, and retain all steps on errors. This is important because screenshots, even with the efficiency gains we've recently gotten, just take up too much space and cost too much. Even a handful of journeys can takeup ~1g/day with those optimizations. This should cut down disk usage dramatically. 

Counterpart of https://github.com/elastic/beats/issues/25201

@liciavale  @drewpost @paulb-elastic @vigneshshanmugam thoughts on this approach? For the `space-efficient` option it'd mean that on the steps page in Uptime we'd usually have missing screenshots.

vigneshshanmugam:
+ part of #308 
+ Extends the previous `--screenshot` option with `on|off|only-on-failure` and defaulting to `on`. 
```
on - capture screenshots for all step in a journey
off - do not capture any screenshots
only-on-failure - capture screenshots for all steps when anyone of the journey fails (any step in a journey fails). 
```
+ This is a non-breaking change and backwards compatible with previous boolean flag.


src/core/runner.ts
```diff
@@ -207,25 +207,29 @@ export default class Runner extends EventEmitter {
       data.error = error;
     } finally {
       data.url ??= driver.page.url();
-      if (screenshots) {
+      if (screenshots !== 'off') {
```
andrewvc, line 210: Could we clean up this function and extract the screenshot portion to a separate function for readability?  `runStep` is getting quite large.

src/core/runner.ts
```diff
@@ -292,15 +296,15 @@ export default class Runner extends EventEmitter {
...
+      options,
       ...pluginOutput,
       browserconsole: status == 'failed' ? pluginOutput.browserconsole : [],
     });
     /**
      * Wait for the all the reported events to be consumed aschronously
      * by reporter.
      */
-    if (options.reporter == 'json') {
+    if (options.reporter === 'json') {
```
andrewvc, line 307: Weird that the linter missed this

src/parse_args.ts
```diff
@@ -59,7 +59,12 @@ program
     '--capability <features...>',
     'Enable capabilities through feature flags'
   )
-  .option('--screenshots', 'take screenshot for each step')
+  .addOption(
```
andrewvc, line 62: Could use a comment here explaining why we have this declared twice. I take it we need the second option for the `--screenshots` option to work with no setting, yes?

src/reporters/json.ts
```diff
@@ -415,42 +415,48 @@ export default class JSONReporter extends BaseReporter {
...
       }) => {
-        await gatherScreenshots(
-          join(CACHE_PATH, 'screenshots'),
-          async (step, data) => {
-            if (ssblocks) {
-              const { blob_mime, blocks, reference } =
-                await getScreenshotBlocks(Buffer.from(data, 'base64'));
-              for (let i = 0; i < blocks.length; i++) {
-                const block = blocks[i];
+        const { ssblocks, screenshots } = options;
```
andrewvc, line 420: Could we break this up as well? It's an unruly size. I can picture this being two functions, one for ssblocks, one for regular screenshots. Or, at a minimum, just extracting the ssblocks logic to its own function.

andrewvc (commented):
Left some thoughts on code quality improvements, have yet to run it, but will do so soon!

src/parse_args.ts
vigneshshanmugam: It was my mistake, i did not remove the next line, it should work without two options.

src/reporters/json.ts
vigneshshanmugam: Sure thing 👍🏽

src/core/runner.ts
vigneshshanmugam: Done.

vigneshshanmugam:
E2E failures is unrelated, have to look in to that separately.

__tests__/reporters/json.test.ts
```diff
@@ -268,49 +272,83 @@ describe('json reporter', () => {
...
+        status,
+        options,
+      });
+
+    it('write whole blobs data ', async () => {
+      emitEnd({
+        screenshots: 'on',
+        ssblocks: false,
+      });
+      const stepEnd = (await readAndCloseStreamJson()).find(
```
andrewvc, line 303: Can we rename these `stepEnd` variables? These look to be copy/pasted from earlier tests where they were indeed `step/end` docs, but they now should be named `stepScreenshot` or `screenshot` etc.

andrewvc (approved):
LGTM pending a minor fix on var naming

__tests__/reporters/json.test.ts
vigneshshanmugam: Good catch, Fixed!

PR #316, opened 2021-06-22
fix: capture beforeAll hook errors

### LINKED ISSUE #280: propagate errors from `beforeAll` and `afterAll` hooks
+ Now error in the `beforeAll` and `afterAll` hooks would be captured and reported as error, but they will not be associated with the reporters in the correct way. 
+ Without associating these errors, the Uptime UI will have no information about what happened during the context of a single execution. 

We have to figure out a way to solve this problem.

vigneshshanmugam:
+ fix #280 
+ We capture the beforeAll hook errors and any error that happens in any one of the hooks would be captured and all journeys that run on the current invocation will report that error event with `failed` status to be able to captured as part of the Uptime UI. 
+ `afterAll` errors behaves the same - we report them in the stderror logs which will be captured by Heartbeat.

### DIFF
--- a/__tests__/core/runner.test.ts
+++ b/__tests__/core/runner.test.ts
@@ -153,12 +153,14 @@ describe('runner', () => {
       throw error;
     });
     runner.addJourney(new Journey({ name: 'j1' }, () => step('step1', noop)));
+    runner.addJourney(new Journey({ name: 'j2' }, () => step('step1', noop)));
     const result = await runner.run({
       wsEndpoint,
       outfd: fs.openSync(dest, 'w'),
     });
     expect(result).toEqual({
       j1: { status: 'failed', error },
+      j2: { status: 'failed', error },
     });
   });
 
--- a/__tests__/core/runner.test.ts
+++ b/__tests__/core/runner.test.ts
@@ -147,6 +147,21 @@ describe('runner', () => {
     });
   });
 
+  it('run journey - failed on beforeAll', async () => {
+    const error = new Error('Broken beforeAll hook');
+    runner.addHook('beforeAll', () => {
+      throw error;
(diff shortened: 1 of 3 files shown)


__tests__/core/runner.test.ts
```diff
@@ -147,6 +147,21 @@ describe('runner', () => {
...
+    const error = new Error('Broken beforeAll hook');
+    runner.addHook('beforeAll', () => {
+      throw error;
+    });
+    runner.addJourney(new Journey({ name: 'j1' }, () => step('step1', noop)));
+    const result = await runner.run({
+      wsEndpoint,
+      outfd: fs.openSync(dest, 'w'),
+    });
+    expect(result).toEqual({
```
andrewvc, line 160: Can we add a second journey to this test? The hook should cause all journeys to fail, so we should assert it for > 1 journey

src/reporters/json.ts
```diff
@@ -425,6 +423,9 @@ export default class JSONReporter extends BaseReporter {
           await gatherScreenshots(
             join(CACHE_PATH, 'screenshots'),
             async (step, data) => {
+              if (!data) {
```
andrewvc, line 426: Can we add a note here mentioning that this is needed for fake journeys?

andrewvc (approved):
LGTM pending the minor changes mentioned

src/reporters/json.ts
vigneshshanmugam: I added this just incase, not really related to the commit.

__tests__/core/runner.test.ts
vigneshshanmugam: Yes good point, will add it.

PR #317, opened 2021-06-23
feat: add browser and context options support

### LINKED ISSUE #169: Add slowMo flag for playwright
Playwright supports a [`slowMo`](https://github.com/microsoft/playwright/blob/master/docs/api.md) flag, which slows down every operation it performs. This is useful for debugging with `--no-headless` as it lets you actually see what's going on.

### LINKED ISSUE #171: Add `ignoreHTTPSError` option on journeys and CLI
Users sometimes need to test sites that have known bad certs, we should probably  set this on a per-journey and CLI options level, both probably have their uses. See https://github.com/microsoft/playwright/issues/2814 for more info on implementation, it's a browser level option.

### LINKED ISSUE #198: Support extra HTTP headers for pages
We should allow people to set custom headers, which must be done when initializing the page (via playwright's `extraHTTPHeaders` option on the page constructor). I propose we add this ability to `before` hooks, letting them either invoke a special method to set this, or set a special variable. This issue is a stub and could use more detail.

### LINKED ISSUE #209: Emulate device type (UA String and Viewport)
As a user of Elastic Synthetic Monitoring
I want to be able to emulate a different device types
So that I can see how well my site performs from when loading the site that is optimised for that device (e.g. a mobile site)

ACs:
* The ability to configure the UA String used by a specific monitor (the same UA String will be used throughout all steps of the user journey)
* The ability to configure the viewport (width and height) properties used by a specific monitor
  * The UI reporting will be able to represent these different viewports, both for the thumbnail and full size screenshots
* The ability to configure the orientation of the screen (if this is possible in Playwright)
* It is possible to determine the device type/emulation used when looking at a test result

We need to consider where these properties are set, for example as parameters to the `journey` object? When integrated with Fleet with a curated UI for configuring monitors, perhaps this could be extended to have a list of common device types to choose from, which sets both the UA string and viewport accordingly (whilst manual configuration of the monitor still allows for these parameters to be set independently).

Whilst most monitors will have the same viewport throughout the whole user journey (and as such, is convenient to set as a parameter for the `journey` object), there may be a future requirement to be able to set this and the orientation on a step by step level (e.g. where a user journey may need to flip between portrait and horizontal for some kind of behaviour). In these cases, it may also be useful to have these defined as `step` parameters (which may be a future enhancement, not necessarily required for this initial version). This seems to be possible in the Playwright [implementation](https://playwright.dev/docs/api/class-page/#pagesetviewportsizeviewportsize).

As in the ACs, outside of any Fleet/Integration curated UI to configure these, there is a @elastic/observability-design  requirement to consider where this is shown in the results. Although the UA String can be determined from the request headers, that’s quite deep in the report hierarchy to discover. The viewport/orientation isn’t shown anywhere right now. It should be possible to see what emulation was set against each monitor (bearing in mind it could change over time).

### LINKED ISSUE #299: Add support for browser & context options via config
There are lot of ways we can configure the Playwright Browser and the context that is used to creating page objects in Playwright. 

1. Browser launch options - `chromium.launch(options)` -  https://playwright.dev/docs/api/class-browsertype?_highlight=launch#browsertypelaunchoptions
    - This would let us control the browser with `timeouts, headless, sandbox, slowMo` and also `browserName` in the future - elastic/synthetics#169 
 
2. Browser Context options - `browser.newContext(options)` - https://playwright.dev/docs/api/class-browser#browsernewcontextoptions
     - These options allow for `ignoreHttpsError, Throttling, Offline emulation, device emulation etc` - elastic/synthetics#171 elastic/synthetics#209 

To make it easier for users, We can introduce new flags in the `synthetics.config.ts` instead of adding CLI option for each and every config out there in the playwright. 

This Proposal is as allow configuring `browser` and `context` options to the synthetic config in addition to the existing `params` support. 

```ts
{
import {devices} from "playwright-chromium"

export default env => {
  return {
    params: {
      url: 'non-dev',
    },
    browser: {
      ...// browser options
    },
    context: {
      ...devices['Pixel 5'] // uses chromium
    }
  };
};
}
```
We could also merge them under `playwrightOptions` instead of separating them out and provide a validation that allowss only a restrictive list of options to keep things cleaner as well. 

```ts
import { devices } from 'playwright-chromium';

export default env => {
  return {
    params: {
      url: 'non-dev',
    },
    playwrightOptions: {
      ignoreHttpsError: true,
      slowMo: true,
      timeout: 1000,
      ...devices['Pixel 5'],
      screenshots: true, // validate and throw error as we capture it ourselves
    },
  };
};
```

@andrewvc @paulb-elastic Thoughts?

vigneshshanmugam:
+ fixes #171 
+ fixes #169 
+ fixes #299 
+ fixes #198 
+ fixes #209 
+ Exposes all playwright `BrowserLaunch` options and `ContextOptions` through `playwrightOptions` using the `synthetics.config.ts` file. 
+ CLI params that exist today like `--headless` and `--sandbox` takes precedence over the playwrightOptions when launch in non headless mode and such. 
+ There is no validation right now as its mostly an escape hatch and will be added in the future to avoid `recordVideo` and `screenshot` capture as synthetics provides them out of the box. 

### Usage
**synthetics.config.ts**
```ts
import type { SyntheticsConfig } from '@elastic/synthetics';

  playwrightOptions: {
    ignoreHTTPSErrors: true,
    extraHTTPHeaders: {
      "foo": "bar"
    },
    ...devices['Galaxy S9+'],
  },
}
export default config;
```

### DIFF
--- a/examples/todos/synthetics.config.ts
+++ b/examples/todos/synthetics.config.ts
@@ -1,0 +1,28 @@
+// note that `env` is the `env` argument passed to the synthetics program
+// the default being `development`
+export default env => {
+  const cfg = {
+    params: {
+      url: 'http://localhost:8080',
+    },
+    playwrightOptions: {
+      // Note, this is mostly equivalent to `...devices["iPhone 6"]` with import { devices } from 'playwright-chromium';
+      // Just expanded for illustration
+      userAgent:
+        'Mozilla/5.0 (iPhone; CPU iPhone OS 11_0 like Mac OS X) AppleWebKit/604.1.38 (KHTML, like Gecko) Version/11.0 Mobile/15A372 Safari/604.1',
+      viewport: {
+        width: 375,
+        height: 667,
+      },
+      deviceScaleFactor: 2,
+      isMobile: true,
+      hasTouch: true,
+    },
+  };
+
+  if (env === 'production') {
+    cfg.params.url = 'https://elastic.github.io/synthetics-demo/';
+  }
+
+  return cfg;
+};
--- a/__tests__/cli.test.ts
+++ b/__tests__/cli.test.ts
@@ -76,10 +76,25 @@ describe('CLI', () => {
       'testing',
     ]);
     await cli.waitFor('journey/start');
-    expect(await cli.exitCode).toBe(0);
     const output = cli.output();
+    expect(await cli.exitCode).toBe(0);
     expect(JSON.parse(output).payload).toMatchObject({
       params: { url: 'non-dev' },
+    });
+  });
+
+  it('pass playwright options to runner', async () => {
+    const cli = new CLIMock([
+      join(FIXTURES_DIR, 'pwoptions.journey.ts'),
+      '--json',
+      '--config',
+      join(FIXTURES_DIR, 'synthetics.config.ts'),
+    ]);
+    await cli.waitFor('step/end');
+    const output = cli.output();
+    expect(await cli.exitCode).toBe(0);
+    expect(JSON.parse(output).step).toMatchObject({
+      status: 'succeeded',
     });
   });
 
--- a/__tests__/config.test.ts
+++ b/__tests__/config.test.ts
@@ -24,6 +24,7 @@
  */
 
 import { join } from 'path';
+import { devices } from 'playwright-chromium';
 import { generateTempPath } from '../src/helpers';
 import { readConfig } from '../src/config';
 
@@ -36,8 +37,11 @@ describe('Config', () => {
       params: {
         url: 'dev',
       },
+      playwrightOptions: {
+        ...devices['Galaxy S9+'],
+      },
     });
-    expect(readConfig('testing', configPath)).toEqual({
+    expect(readConfig('testing', configPath)).toMatchObject({
       params: {
         url: 'non-dev',
       },
--- a/__tests__/fixtures/pwoptions.journey.ts
+++ b/__tests__/fixtures/pwoptions.journey.ts
@@ -23,20 +23,14 @@
  *
  */
 
-import { devices } from 'playwright-chromium';
-import type { SyntheticsConfig } from '../../src';
+import { journey, step, expect } from '../../dist';
 
-module.exports = env => {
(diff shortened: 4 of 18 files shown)


examples/todos/synthetics.config.ts
```diff
@@ -0,0 +1,28 @@
...
+  const cfg = {
+    params: {
+      url: 'http://localhost:8080',
+    },
+    playwrightOptions: {
+      // Note, this is mostly equivalent to `...devices["iPhone 6"]` with import { devices } from 'playwright-chromium';
+      // Just expanded for illustration
+      userAgent:
+        'Mozilla/5.0 (iPhone; CPU iPhone OS 11_0 like Mac OS X) AppleWebKit/604.1.38 (KHTML, like Gecko) Version/11.0 Mobile/15A372 Safari/604.1',
+      viewport: {
```
andrewvc, line 13: I don't see these options reflected with `--no-headless`, are we sure they're applied correctly? Could be an error on my part.

andrewvc (commented):
This generally looks good, but I think there may be a bug present. I tried to modify the `todos` example (pushed to your branch, hope that's not a problem), and don't see the added options in effect.

One other note, you must invoke `elastic-synthetics` from inside the `todos` folder, not via `elastic-synthetics examples/todos`  for the `synthetics.config.ts` to be applied (I'll open a separate issue for that).

examples/todos/synthetics.config.ts
vigneshshanmugam: It does work properly for me, can you pass me the command used to invoke these?

vigneshshanmugam:
> One other note, you must invoke elastic-synthetics from inside the todos folder, not via elastic-synthetics examples/todos for the `synthetics.config.ts` to be applied (I'll open a separate issue for that).

This is intended, we traverse up the root and not down the root path. This is keeping in mind how heartbeat invokes the underlying suites test and matches the behaviour of finding package.json files.

andrewvc (approved):
LGTM

PR #318, opened 2021-06-24
BREAKING CHANGE: remove -e in favor of NODE_ENV

vigneshshanmugam:
+ Remove -e flag in favor of NODE_ENV which will behave the same way as the -e flag. 
+ Dynamic config still works the same way and we default to `development` when running suites.


andrewvc (approved):
LGTM

PR #321, opened 2021-06-24
fix: avoid writing duplicate screenshot docs

### LINKED ISSUE #320: screenshots docs are duplicated across journeys
+ As we utilize the global cache path, on each journey end we end up writing the same docs that was already written on the previous journey resulting in N*M screenshot docs where N is the number of journeys and M is the number of steps.

vigneshshanmugam:
+ fix #320 
+ We clear the global screenshot cache once each journey end. We do not need to worry about the parallel execution problem here as our screenshot path contains the process id which would be different for each worker if we end up running journeys in parallel. 
+ This feels much simpler than creating a hash for every journey where we need to pass that hash across every journey context. 
+ Also cleaned up some deprecated API and moved to promise based FS API.

### DIFF
--- a/__tests__/core/runner.test.ts
+++ b/__tests__/core/runner.test.ts
@@ -515,4 +515,49 @@ describe('runner', () => {
       fd: expect.any(Number),
     });
   });
+
+  const readAndCloseStreamJson = () => {
+    const fd = fs.openSync(dest, 'r');
+    const buffer = fs.readFileSync(fd, 'utf-8');
+    const out = [];
+    buffer.split('\n').forEach(l => {
+      try {
+        out.push(JSON.parse(l));
+      } catch (e) {
+        return; // ignore empty lines
+      }
+    });
+    return out;
+  };
+
+  it('run api - verify screenshots', async () => {
+    const j1 = new Journey({ name: 'j1' }, noop);
+    const s1 = j1.addStep('j1s1', noop);
+    const j2 = new Journey({ name: 'j2' }, noop);
+    const s2 = j2.addStep('j2s2', noop);
+    runner.addJourney(j1);
+    runner.addJourney(j2);
+
+    await runner.run({
+      wsEndpoint,
+      reporter: 'json',
+      screenshots: 'on',
+      outfd: fs.openSync(dest, 'w'),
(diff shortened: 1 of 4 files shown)


__tests__/core/runner.test.ts
```diff
@@ -515,4 +515,49 @@ describe('runner', () => {
...
+      wsEndpoint,
+      reporter: 'json',
+      screenshots: 'on',
+      outfd: fs.openSync(dest, 'w'),
+    });
+
+    const screenshotJson = readAndCloseStreamJson().filter(
+      ({ type }) => type === 'step/screenshot'
+    );
+    expect(screenshotJson.length).toEqual(2);
```
vigneshshanmugam, line 551: running this test on master reveals the bug where there will be 3 docs in total.

src/core/runner.ts
```diff
@@ -428,16 +435,16 @@ export default class Runner extends EventEmitter {
...
     return result;
   }
 
-  reset() {
+  async reset() {
     /**
      * Clear all cache data stored for post processing by
      * the current synthetic agent run
      */
-    rmdirSync(CACHE_PATH, { recursive: true });
```
vigneshshanmugam, line 440: rmdirSync is deprecated from Node 14, so cleaned it up and moved to promise based FS API's

__tests__/core/runner.test.ts
andrewvc: Thanks for the note!

src/helpers.ts
```diff
@@ -48,8 +48,12 @@ export const symbols = {
   failed: red('✖'),
 };
 
+export function generateHash() {
```
andrewvc, line 51: Can we rename this? It's not a hash, but a unique value. Maybe `generateUniqueValue`. Could we also either add to or replace this without the output of `Math.random()` to add more entropy? I'm always suspicious of timers as sources of uniqueness.

andrewvc (approved):
LGTM pending that minor change in the naming of the uniqueness function

src/helpers.ts
vigneshshanmugam: Done.

PR #322, opened 2021-06-25
feat: default to ssblocks for richevents

### LINKED ISSUE #319: `--screenshots` option is overridden by `--rich-events`
It seems that if using `--rich-events` in conjunction with `--screenshots` the latter option is fully ignored, meaning https://github.com/elastic/beats/pull/26455#pullrequestreview-692291735 doesn't actually have any effect. 

I propose we simply remove `--screenshots` from `--rich-events` as a simple solution.

vigneshshanmugam:
+ We now default to spitting out screenshot blocks whenever `--rich-events` flag is enabled which mimicks the heartbeat mode. 
+ fixes #319 - Allow screenshots option to be overridden per heartbeat monitor config. 
+ remove `-j ` flag in favor of using `--reporter`
+ `network` is now moved inside `capability` to match with other tracing, metrics and filmstrip capability.

### DIFF
--- a/__tests__/cli.test.ts
+++ b/__tests__/cli.test.ts
@@ -42,20 +42,10 @@ describe('CLI', () => {
   });
 
   it('produce json output via --json flag', async () => {
-    const cli = new CLIMock([join(FIXTURES_DIR, 'fake.journey.ts'), '--json']);
-    await cli.waitFor('fake journey');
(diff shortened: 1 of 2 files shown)


andrewvc (approved):
LGTM, love the test coverage

PR #327, opened 2021-06-29
[npm] limit to node@12 and npm@6

spalger:
In order to help devs working with the code locally this adds a `.nvmrc` and `"engines"` configuration to package.json which indicates that Node @ v12 should be used along with an npm @ v6 version.


vigneshshanmugam (commented):
We use node:12 in the Heartbeat image which uses `npm@6.x`, So I am not sure if its a good idea to move to npm 7 locally. We would then have to update heartbeat to use npm7 and such incase there are any issues. 

WDYT?

spalger:
Interesting, I'm open to using `node@12` and `npm@6` for local dev, happy to add `.npmrc` and `"engines"` stuff to package.json.

vigneshshanmugam:
> happy to add .npmrc and "engines" stuff to package.json.

Sounds good 👍🏽, yes lets do it.

vigneshshanmugam (approved):

vigneshshanmugam (approved):
Thanks @spalger

PR #328, opened 2021-06-29
wrap fs methods with promisify for node@12 compat

spalger:
While working on limiting the node version in CI to match the version installed by Heartbeat (#327) we realized that the use of the `fs/promises` API is not compatible with this node version.

This PR removes the `fs/promises` usage and instead exposes part of the fs API wrapped with `util.promisify()` from `src/helpers` and uses that in `src/core/runner`.

### DIFF
--- a/src/core/runner.ts
+++ b/src/core/runner.ts
@@ -25,7 +25,6 @@
 
 import { once, EventEmitter } from 'events';
 import { join } from 'path';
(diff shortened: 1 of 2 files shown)


vigneshshanmugam (approved):
Thanks @spalger for the quick fix, 

I will merge this once I fix the broken tests in master.

PR #343, opened 2021-07-07
fix: remove the node version restriction

vigneshshanmugam:
+ fix errors that will be thrown when using different version of node on the local environment. 

From Kibana
```
info [bazel] error @elastic/synthetics@1.0.0-beta.8: The engine "node" is incompatible with this module. Expected version "12.18.4". Got "14.17.2"
```
+ CI still runs with the version of `12.18.4` as per `.nvmrc` file.

### DIFF
--- a/package.json
+++ b/package.json
@@ -69,9 +69,8 @@
     "ts-jest": "^26.4.4",
     "rimraf": "^3.0.2"
   },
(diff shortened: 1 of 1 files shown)


shahzad31 (approved):
LGTM !!

PR #346, opened 2021-07-13
fix: default to name for journey id

vigneshshanmugam:
+ Id and Name are always required for the monitor to show correctly in the Uptime UI.


justinkambic (approved):
LGTM

Issue #253, opened 2021-04-12
Synthetics Doesn’t Handle Popups (new tab or new window)


paulb-elastic:
The Synthetic Agent is capturing the performance data and screenshots for the main tab/page of the browser.

If a user journey causes a new tab or window (popup) to be opened, the metrics and screen shots are still taken from the initial (main) page, so there are no new metrics for the new tab/page, and the screenshot still reflects the original page.

This can be recreated by extending the todos folder/app

Create `./examples/todos/app/popup1.html`
```
<html>
<head>
</head>
<body>
Test popup page 1
<p><a href="popup2.html" onclick="return !window.open(this.href, 'Test Page', 'width=500,height=500')" target="_blank" id="linkWithPopup">Click</a> here (id=linkWithPopup) to open - Test Page - popup window</p>
<p><a href="popup2.html" target="_blank" id="linkWithPopupNoNewWindow">Click</a> here (id=linkWithPopupNoNewWindow) to open - Test Page - new tab (not new window)</p>
</body>
</html>
```
Create `./examples/todos/app/popup2.html`
```
<html>
<head>
</head>
<body>
Test popup page 2
</body>
</html>
```
Create `./examples/todos/popup.journey.ts`
```
import { journey, step } from '@elastic/synthetics';
import { deepStrictEqual } from 'assert';
import { join } from 'path';

journey('Popup Test', async ({ page }) => {
  step('Popup Landing Page', async () => {
    const path = 'file://' + join(__dirname, 'app', 'popup1.html');
    await page.goto(path);
    await page.waitForSelector('text=Test popup page 1');
  });

  step('Open Popup', async () => {
    // Test 1 - new window (popup)
    await page.click('#linkWithPopup');

    // Test 2 - new tab (not new popup window)
    //await page.click('#linkWithPopupNoNewWindow');

    const [popup] = await Promise.all([
      page.waitForEvent('popup'),
      page.click('#linkWithPopup')
    ]);
    await popup.waitForSelector('text=Test popup page 2');
  });
});
```

**Note** that there are two test cases (both will fail in the same way)
* Test 1 opens a new popup window
* Test 2 (commented out) opens a new tab (not a separate popup window)

We can act on the new/popup page, for example with `popup.waitForSelector
` as shown in [this example](https://playwright.dev/docs/api/class-page/#pagewaitforloadstatestate-options), however the Synthetic Agent is still connected to the CDP from the original window.

When running this, you will notice:
[...]

PR #352, opened 2021-07-21
fix: capture correct url for navigations

vigneshshanmugam:
+ When there are iframes and new windows that exists as part of the page lifecycle, we capture the wrong page URL as we are using the page object from previous session. 
+ part of #253


andrewvc:
LGTM

PR #353, opened 2021-07-22
fix: record correct screenshots for popups

vigneshshanmugam:
+ part of #253 
+ captures proper screenshots for new navigations done via popups and new tabs.
+ Listens for `page` events in each step and captures the screenshot of last navigation that happened in the step. 

### Tested using inline script for popups and navigation
```js
 step('Popup Landing Page', async () => {
  await page.goto(`http://localhost:8080/popup1.html`);
  await page.waitForSelector('text=Test popup page 1');
});
step('Open Popup', async () => {
  const [popup] = await Promise.all([
    page.waitForEvent('popup'),
    page.click('#linkWithPopup'),
  ]);
  // Test 2 - new tab (not new popup window)
  await Promise.all([
    context.waitForEvent('page'),
    page.click('#linkWithPopupNoNewWindow'),
  ]);
});

```

### Screenshots

Screen Shot 2021-07-21 at 17 00 12
Screen Shot 2021-07-21 at 17 00 02

### DIFF
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -51,9 +51,16 @@ if (process.env.DEBUG === namespace || Boolean(options.debug)) {
 }
 
 const loadInlineScript = source => {
-  const scriptFn = new Function('step', 'page', 'browser', 'params', source);
-  journey('inline', async ({ page, browser, params }) => {
-    scriptFn.apply(null, [step, page, browser, params]);
+  const scriptFn = new Function(
+    'step',
+    'page',
+    'context',
+    'browser',
+    'params',
+    source
+  );
+  journey('inline', async ({ page, context, browser, params }) => {
+    scriptFn.apply(null, [step, page, context, browser, params]);
   });
 };
 
--- a/src/core/runner.ts
+++ b/src/core/runner.ts
@@ -233,13 +233,7 @@ export default class Runner extends EventEmitter {
       driver.page.off('request', captureUrl);
     };
     driver.page.on('request', captureUrl);
-    /**
-     * Capture screenshot for the newly created pages
-     * via popup or new windows/tabs
-     */
-    let page = driver.page;
-    const capturePage = newPage => (page = newPage);
-    driver.context.on('page', capturePage);
+
     try {
       pluginManager.onStep(step);
       await step.callback();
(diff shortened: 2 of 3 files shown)


src/cli.ts
```diff
@@ -51,9 +51,16 @@ if (process.env.DEBUG === namespace || Boolean(options.debug)) {
...
+  const scriptFn = new Function(
+    'step',
+    'page',
+    'context',
+    'browser',
+    'params',
+    source
+  );
+  journey('inline', async ({ page, context, browser, params }) => {
+    scriptFn.apply(null, [step, page, context, browser, params]);
```
vigneshshanmugam, line 63: allows the use of `context` in inline journeys, was missing before.

src/core/runner.ts
```diff
@@ -243,9 +244,19 @@ export default class Runner extends EventEmitter {
...
-      data.url ??= driver.page.url();
-      if (screenshots && screenshots !== 'off') {
-        await this.captureScreenshot(driver.page, step);
+      /**
+       * Capture screenshot for the newly created pages
+       * via popup or new windows/tabs
+       *
+       * Last open page will get us the correct screenshot
+       */
+      const pages = driver.context.pages();
```
vigneshshanmugam, line 253: Note: This handles if any popups/tabs are closed during the step. 

Example

```
// Test 1 - new window (popup)
const [popup] = await Promise.all([
  page.waitForEvent('popup'),
  page.click('#linkWithPopup'),
]);
await popup.close();
```

andrewvc (commented):
This needs some tests. Also, after the PR is merged we should add this to the docs. It's not entirely obvious how this would work to a user. 

Also, if a user wants to switch the focus back to the original page how does that work? Looking at the code it looks like you'd need to close the popup. Is that right? If so, we should document that.

vigneshshanmugam:
> Also, if a user wants to switch the focus back to the original page how does that work? Looking at the code it looks like you'd need to close the popup. Is that right? If so, we should document that.

Not sure if that should be documented, the active window in the step should be screenshoted, otherwise we need to come up with a plan of having multiple screenshots per step if there are more than one active window/pageg.

andrewvc (approved):
LGTM

Issue #356, opened 2021-07-28
Rich events should not exit with status > 0 when tests fail


andrewvc:
Synthetics currently exits with a > 0 status code when any test fails. Heartbeat logs errors for any > 0 status code. This was primarily done to make CI services work right since the look at exit codes. 

For heartbeat, however, it's hard to tell the difference between a failed test > 0 exit code and something seriously wrong (e.g. node couldn't be found), this leads to alarming error logs as below:

```
2021-07-28T03:08:14.573Z        INFO    synthexec/synthexec.go:172      Command has completed(1): /usr/share/heartbeat/.node/node/bin/elastic-synthetics --screenshots on --inline --rich-events --outfd 3
2021-07-28T03:08:14.573Z        WARN    synthexec/synthexec.go:179      Error executing command '/usr/share/heartbeat/.node/node/bin/elastic-synthetics --screenshots on --inline --rich-events --outfd 3': exit status 1
```

We really only should record the first `INFO` log line, and omit the second `WARN` line unless something terrible actually happened.

There's no need to have a non-zero exit code if the suite runs successfully with `--rich-events` since the JSON has all the data needed to understand that.

PR #357, opened 2021-07-28
Add --quiet-exit-code and make it default --rich-events behavior

andrewvc:
Fixes https://github.com/elastic/synthetics/issues/356

Accomplishes the fix by making --rich-events always exit with a code of 0, even if tests fail

### DIFF
--- a/__tests__/cli.test.ts
+++ b/__tests__/cli.test.ts
@@ -41,15 +41,13 @@ describe('CLI', () => {
     expect(await cli.exitCode).toBe(1);
   });
 
-  describe('with --quiet-exit-code', () => {
-    it('runs the suites, exiting with 0', async () => {
-      const cli = new CLIMock([
-        join(FIXTURES_DIR, 'error.journey.ts'),
-        '--quiet-exit-code',
-      ]);
-      await cli.waitFor('boom');
-      expect(await cli.exitCode).toBe(0);
-    });
+  it('runs the suites with --quiet-exit-code, always exiting with 0', async () => {
+    const cli = new CLIMock([
+      join(FIXTURES_DIR, 'error.journey.ts'),
+      '--quiet-exit-code',
+    ]);
+    await cli.waitFor('boom');
+    expect(await cli.exitCode).toBe(0);
   });
 
   it('produce json output  --json and reporter=json flag', async () => {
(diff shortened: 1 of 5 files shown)


vigneshshanmugam (commented):
The code looks good, But I would like to propose another solution. Instead of adding a flag specific for HB, we could exit with non zero code for CI only environments? 

https://www.npmjs.com/package/is-ci, we can use this. Thoughts?

andrewvc:
I'd rather avoid the dependency, and while I did say `ci` in the description, I think the current exit code behavior makes sense on the CLI as well, it only doesn't make sense in heartbeat. 

I should mention, the flag is only there for completeness, since it's turned on by default with `--rich-events` heartbeat won't need to add any other flags.

__tests__/cli.test.ts
```diff
@@ -41,6 +41,17 @@ describe('CLI', () => {
     expect(await cli.exitCode).toBe(1);
   });
 
+  describe('with --quiet-exit-code', () => {
+    it('runs the suites, exiting with 0', async () => {
```
vigneshshanmugam, line 45: nit: would move it out of `describe` block to keep in same level as other test.

vigneshshanmugam (approved):

__tests__/cli.test.ts
```diff
@@ -77,6 +88,7 @@ describe('CLI', () => {
   it('mimick new heartbeat with `--rich-events` flag', async () => {
     const cli = new CLIMock([
       join(FIXTURES_DIR, 'fake.journey.ts'),
+      join(FIXTURES_DIR, 'error.journey.ts'),
```
vigneshshanmugam, line 91: We need to increase the timeout for this specific test.

it ('asda', callback, timeout: number), Increasing with 10 secs should do it as we now test for 2 files and also screenshots which takes more than 5 seconds.

__tests__/cli.test.ts
```diff
@@ -94,7 +104,7 @@ describe('CLI', () => {
     });
 
     expect(await cli.exitCode).toBe(0);
-  });
+  }, 30);
```
vigneshshanmugam, line 107: nit: this is in milliseconds.

PR #361, opened 2021-08-10
Add --ignore-https-errors flag

andrewvc:
Adds the ability to ignore HTTPS errors via a flag. Checks one box in https://github.com/elastic/beats/issues/27202

This is adding a flag, not a capability since if a user specifies the new option we want it to fail, rather than silently behave as though it were not set.


examples/todos/package-lock.json
```diff
@@ -0,0 +1,1315 @@
+{
```
vigneshshanmugam, line 1: Need to remove this file. Not sure how it showed up as we have a `.npmrc` file with `package-lock=false
`

__tests__/cli.test.ts
```diff
@@ -238,6 +249,34 @@ describe('CLI', () => {
     expect(await cli.exitCode).toBe(0);
     process.env['TS_NODE_TYPE_CHECK'] = 'false';
   });
+
+  describe('testing a TLS site with self-signed cert', () => {
```
vigneshshanmugam, line 253: nit: Would keep it smaller
```suggestion
  describe('TLS site with self-signed cert', () => {
```

__tests__/cli.test.ts
```diff
@@ -238,6 +249,34 @@ describe('CLI', () => {
...
+    afterAll(async () => await tlsServer.close());
+
+    it('fails by default', async () => {
+      const cli = new CLIMock(cliArgs);
+      expect(await cli.exitCode).toBe(1);
+    });
+
+    it('succeeds succeeds with --ignoreHTTPSErrors', async () => {
+      const cli = new CLIMock([...cliArgs, '--ignoreHTTPSErrors']);
+      expect(await cli.exitCode).toBe(0);
```
vigneshshanmugam, line 277: would look for some events to exist, just for double checking.

__tests__/cli.test.ts
```diff
@@ -259,13 +298,16 @@ class CLIMock {
...
+      // Uncomment the line below if the process is blocked and you need to see its output
+      // console.warn(this.data);
       this.chunks.push(...this.data.split('\n').filter(Boolean));
       if (this.waitForPromise && this.data.includes(this.waitForText)) {
         this.process.stdout.off('data', dataListener);
         this.waitForPromise();
       }
     };
     this.process.stdout.on('data', dataListener);
+    this.process.stderr.on('data', dataListener);
```
vigneshshanmugam, line 310: This is inccorrect, Any data in stderr, should exit the process which is handled here - https://github.com/elastic/synthetics/blob/6cf145200e73d6c127cf12e34be36294504e76df/__tests__/cli.test.ts#L313

__tests__/fixtures/selfsigned.cert
```diff
@@ -0,0 +1,23 @@
+-----BEGIN CERTIFICATE-----
```
vigneshshanmugam, line 1: Would move the `cert` and `key` files under new folder `fixtures/ca/*`

src/parse_args.ts
```diff
@@ -91,6 +91,10 @@ program
     '--pause-on-error',
     'pause on error until a keypress is made in the console. Useful during development'
   )
+  .option(
+    '--ignoreHTTPSErrors',
```
vigneshshanmugam, line 95: should be `--ignore-https-errors`

src/parse_args.ts
```diff
@@ -126,6 +130,7 @@ if (options.capability) {
    * filmstrips - record detailed filmstrips for all journeys
    * metrics - capture performance metrics (DOM Nodes, Heap size, etc.) for each step
    * ssblocks - Dedupes the screenshots in to blocks to save storage space
+   * ignoreHTTPsErrors - Sets playwright option to ignore HTTPS errors
```
vigneshshanmugam, line 133: nit: remove from capability comments.

andrewvc:
I removed the promise rejection on stderr. We should generally be able to consider exit status for determining error. If we want to watch for `stderr` messages we should make that explicit in the tests IMHO since `stderr` != a failed test where we want to short circuit the CLI Mock.

examples/todos/package-lock.json
vigneshshanmugam: @andrewvc Can you remove this file?

vigneshshanmugam:
@andrewvc Sure, Agree with your thoughts. We can watch out for exitCode which should generally work, but it was easier to debug when something was written on stderr either from CLI or errors thrown from Playwright which we haven't caught. But for now I would resolve and reject the promise if exitCode is != 0.

andrewvc:
@vigneshshanmugam if you're OK with the current behavior can I get an LGTM?

__tests__/cli.test.ts
andrewvc: Done

__tests__/cli.test.ts
```diff
@@ -116,7 +127,13 @@ describe('CLI', () => {
       join(FIXTURES_DIR, 'synthetics.config.ts'),
     ]);
     await cli.waitFor('journey/end');
-    const data = cli.buffer().map(data => JSON.parse(data));
+    const data = cli.buffer().map(data => {
+      try {
+        return JSON.parse(data);
+      } catch (e) {
+        throw `Error ${e} could not parse data '${data}'`;
```
vigneshshanmugam, line 134: Glad you did this 👍🏽

vigneshshanmugam (approved):

PR #364, opened 2021-08-12
Bump to node 14

andrewvc:
This pairs with https://github.com/elastic/beats/pull/27345 to bump the node to the latest LTS version.


vigneshshanmugam (approved):
The E2E right now works on the older version as the heartbeat image is old. We need to do manual testing to ensure everything works once new HB image is ready,

PR #369, opened 2021-08-27
feat: introduce step level tracing

### LINKED ISSUE #358: Introduce step level tracing for perf metrics
#### Problem:

Currently, it's hard to associate the performance metrics like FCP, LCP, etc. that gets measured as part of the journey as there could be multiple navigations inside steps of a given journey. We currently only capture the last navigation which means we lose the tracing metrics for all previous navigations. 

Navigation Metrics - DOM Load, FCP, LCP, etc.
Non navigational metrics - CLS, Long tasks, User timing metrics etc. 

####  Solution:

1. Start tracer before each step.
    - If step contains navigation - hard/sot - Report all navigation metrics like DOM load, Vitals etc and include them as `step/metrics`
    - If step does not contain any navigation metrics and report them under `step/metrics`. 
2. Stop tracer at end of each step. 

**Filmstrips would have to work the same way as they are part of tracing architecture.**

In order to reduce the CPU load when tracing, we need to increase the sampling interval.

vigneshshanmugam:
+ fix #358 
+ Updates the lighthouse processor with new changes that accounts for different modes - navigation, snapshot etc
+ Trace computation has been updated to handle the navigation changes inside steps
+ Plugin manager is cleaned up to  add support for step level and journey level plugins with new API such as register, unregister etc. 
+ Each Plugin follows the same interface to keep things simpler in the PluginManager. 
    - new Plugin(driver)
    - plugin.start
    - plugin.stop
+ JSON reporter now writes Metrics and Traces for each step if there is a navigation event of interest.


src/core/runner.ts
```diff
@@ -83,13 +84,18 @@ type JourneyContext = BaseContext & {
 type StepResult = {
   status: StatusValue;
   url?: string;
-  metrics?: Metrics;
   error?: Error;
+  pagemetrics?: PageMetrics;
+  filmstrips?: PluginOutput['filmstrips'];
```
andrewvc, line 89: I didn't know about this syntax with the brackets

src/core/runner.ts
```diff
@@ -233,10 +238,26 @@ export default class Runner extends EventEmitter {
     driver.context.on('request', captureUrl);
 
     try {
+      /**
+       * Set up plugin maanger context and also register
```
andrewvc, line 242:
```suggestion
       * Set up plugin manager context and also register
```

src/core/runner.ts
```diff
@@ -233,10 +238,26 @@ export default class Runner extends EventEmitter {
...
+       */
       if (metrics) {
-        data.metrics = await pluginManager.get(PerformanceManager).getMetrics();
+        data.pagemetrics = await (
+          pluginManager.get('performance') as PerformanceManager
+        ).getMetrics();
+      }
+      if (traceEnabled) {
+        const traceOutput = await pluginManager.stop('trace');
+        Object.assign(data, { ...traceOutput });
```
andrewvc, line 260: What is the difference between this and `Object.assign(data, traceOutput);` ? They are the same no?

src/plugins/network.ts
```diff
@@ -207,159 +209,150 @@ export class NetworkManager {
...
+  }
+  if (startTime > responseReceivedTime) {
+    responseReceivedTime = startTime;
+  }
+  return responseReceivedTime;
+}
+
+/**
+ * The timing calculations are based on the chrome devtools frontend
+ * https://github.com/ChromeDevTools/devtools-frontend/blob/7f5478d8ceb7586f23f4073ab5c2085dac1ec26a/front_end/network/RequestTimingView.js#L98-L193
```
andrewvc, line 238: Can we add the license for this linked file to `NOTICE.txt` and reference that here?

src/reporters/json.ts
```diff
@@ -379,7 +379,37 @@ export default class JSONReporter extends BaseReporter {
...
+        pagemetrics,
+        traces,
+        metrics,
+        filmstrips,
+      }) => {
+        this.writeMetrics(journey, step, 'relative_trace', traces);
+        this.writeMetrics(journey, step, 'experience', metrics);
+        if (filmstrips) {
+          // Write each filmstrip separately so that we don't get documents that are too large
+          filmstrips.forEach((strip, index) => {
```
andrewvc, line 399: Does this affect timing? I thought we saved it for the end on purpose so as not to affect journey timing as much

andrewvc (changes requested):
Looks great aside from cleaning up the licensing a bit

src/core/runner.ts
vigneshshanmugam: Yes fixed it.

src/reporters/json.ts
vigneshshanmugam: We left it to the end to make sure to run the tracer only once per journey, but we are already running trace for each step anyway so there is no harm in also capturing filmstrips.

src/plugins/network.ts
vigneshshanmugam: Done.

src/plugins/plugin-manager.ts
```diff
@@ -23,75 +23,93 @@
...
         break;
       case 'performance':
-        instance = new PerformanceManager(this.driver.client);
-        instance.start();
+        instance = new PerformanceManager(this.driver);
         break;
       case 'browserconsole':
-        instance = new BrowserConsole(this.driver.page);
-        instance.start();
+        instance = new BrowserConsole(this.driver);
```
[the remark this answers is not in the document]
vigneshshanmugam: Not really, only to be consistent across all plugins. Makes things easier.

src/plugins/performance.ts
```diff
@@ -41,7 +41,7 @@ const supportedMetrics = new Set<string>([
   'JSHeapTotalSize',
 ]);
 
-export interface Metrics {
+export interface PageMetrics {
```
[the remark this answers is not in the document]
vigneshshanmugam: Its per browser tab which alias to `Page` in playwright context.

src/plugins/plugin-manager.ts
```diff
@@ -23,75 +23,93 @@
...
-  get<T extends Plugin>(Type: new (...args: any[]) => T): T {
-    return this.plugins.get(Type.name) as T;
+  get(type: PluginType) {
+    return this.plugins.get(type);
   }
 
   onStep(step: Step) {
-    this.get(BrowserConsole) && (this.get(BrowserConsole)._currentStep = step);
-    this.get(NetworkManager) && (this.get(NetworkManager)._currentStep = step);
+    (this.get('browserconsole') as BrowserConsole)._currentStep = step;
```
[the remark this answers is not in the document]
vigneshshanmugam: Good point, But we don't want to call this for each plugin instead abstract away from runner and handle it in single place in PluginManger which handles the logic of assigning to respective plugins.

andrewvc (approved):
LGTM Pending @dominiqueclarke +1

dominiqueclarke (approved):
LGTM

PR #372, opened 2021-09-08
feat: use events from PW for network data

### LINKED ISSUE #253: Synthetics Doesn’t Handle Popups (new tab or new window)

### LINKED ISSUE #287: Use network events from PW instead of using CDPSession
Currently, we use own internal CDP session for monitoring network info and construct the network waterfall. It works good, but requires a heavy work when we want to capture network events from popups and other tabs. 

We uses this approach as playwright previous did not have any HAR tracer and used page level network events which is pretty much same as ours. Now PW exposes all these events from the browser context https://github.com/microsoft/playwright/pull/6370 which means we can extract all the network information for cross pages, popups and windows relying on the same. 

However, we need to do this after a new release of PW is published including the network events changes and also the security details.

vigneshshanmugam:
+ fix #287 
+ fix #253 

TODO
- [ ] Response/Request size calculation
- [ ] HTTP Version calculation missing from PW


__tests__/plugins/network.test.ts
```diff
@@ -149,114 +154,34 @@ describe('network', () => {
...
     expect(netinfo[1]).toMatchObject({
       url: `${server.PREFIX}/chunked`,
       status: 200,
       response: expect.any(Object),
       timings: expect.any(Object),
     });
-    expect(netinfo[1].timings.total).toBeGreaterThan(delayTime);
-  });
-
-  describe('waterfall timing calculation', () => {
```
andrewvc, line 180: Is there a good reason to remove this? It seems helpful if for no other reason than to ensure that the data we send to heartbeat/uptime is consistent. Is that checked elsewhere?

__tests__/reporters/__snapshots__/json.test.ts.snap
```diff
@@ -10,64 +10,34 @@ Object {
...
-        "referrer": undefined,
-        "referrer_policy": "strict-origin-when-cross-origin",
         "url": "https://vigneshh.in/",
       },
       "response": Object {
-        "body": Object {
-          "bytes": 3329,
-        },
-        "connection_id": 12,
-        "connection_reused": false,
```
andrewvc, line 36: I take it all these fields are unused currently?

src/reporters/json.ts
```diff
@@ -207,12 +179,11 @@ export function formatNetworkFields(network: NetworkInfo) {
     user_agent: {
       name: browser.name,
       version: browser.version,
-      original: request.headers?.['User-Agent'],
+      original: request.headers?.['user-agent'],
     },
     http: {
-      version: formatVersion(response?.protocol),
```
andrewvc, line 213: Why did we remove the version?

src/reporters/json.ts
```diff
@@ -132,9 +132,10 @@ function getMetadata() {
   };
 }
 
+/* eslint-disable-next-line @typescript-eslint/no-unused-vars */
```
andrewvc, line 135: Why are we keeping this if unused? Why not use it for ECS later as we used to?

__tests__/plugins/network.test.ts
```diff
@@ -105,11 +106,11 @@ describe('network', () => {
     });
     await driver.page.goto(server.PREFIX + '/route1');
     const netinfo = await network.stop();
+    await Gatherer.stop();
     expect(netinfo[0]).toMatchObject({
-      resourceSize: 10,
-      transferSize: expect.any(Number),
+      resourceSize: -1,
```
andrewvc, line 111: What's the reason for the change here?

__tests__/plugins/network.test.ts
vigneshshanmugam: We don't do calculation in the Synthetics code anymore, It's as part of Playwright code.

__tests__/reporters/__snapshots__/json.test.ts.snap
vigneshshanmugam: Yes thats correct, we were directly sending all the fields that were coming out of chrome protocol. https://chromedevtools.github.io/devtools-protocol/tot/Network/#type-Response

Now we only send the strict data that is being given out from PW.

src/reporters/json.ts
vigneshshanmugam: In the PR description already

> http.version data is not exposed, but we can add that as a follow up once PW adds support for it and we are not using it on the waterfall UI.

src/reporters/json.ts
vigneshshanmugam: We might reuse it once we do a PR to the playwright to expose the version.

__tests__/plugins/network.test.ts
vigneshshanmugam: As per the PR description

> Bare uncompressed resource size is not exposed for each request instead we have header and body sizes for both request and responses.

shahzad31:
Aren't we displaying queuing time in waterfal chart, if so, this seems like a breaking change.

src/plugins/network.ts
```diff
@@ -23,336 +23,217 @@
...
+  }
+
+  private async _onRequestCompleted(request: Request) {
+    const networkEntry = this._findNetworkEntry(request);
+    if (!networkEntry) return;
+
+    networkEntry.loadEndTime = epochTimeInMills();
+
+    // For aborted/failed requests sizes does not exist
+    const sizes = await request.sizes().catch(() => {});
```
dominiqueclarke, line 141: Could we add an `isFailedRequest` boolean to the NetworkInfo type to more easily identify failed/aborted requests on the front end?

src/plugins/network.ts
```diff
@@ -23,336 +23,217 @@
...
+    const firstPositive = (numbers: number[]) => {
+      for (let i = 0; i < numbers.length; ++i) {
+        if (numbers[i] > 0) {
+          return numbers[i];
+        }
+      }
+      return null;
+    };
+    const roundMilliSecs = (value: number): number => {
+      return ((value * 1000) | 0) / 1000;
```
dominiqueclarke, line 178: Is there a more readable alternative to use bitwise operators here?

dominiqueclarke (commented):
Looks great. Do you have any suggestions for testing the feature? Can I run it with heartbeat to check for compatibility with waterfall?

src/plugins/network.ts
```diff
@@ -23,336 +23,217 @@
...
-    record.resourceSize += dataLength;
-    if (encodedDataLength >= 0) record.transferSize += encodedDataLength;
-  }
-
-  _onResponseReceived(event: Protocol.Network.responseReceivedPayload) {
-    const { requestId, response, timestamp } = event;
-    const record = this._findNetworkRecord(requestId);
-    if (!record) {
-      return;
+    (request as any)[this._networkEntrySymbol] = networkEntry;
```
dominiqueclarke, line 100: Could you extend the request type to prevent having to cast it to `any`?
I'm playing around with this on my local copy of your PR, and it seems to be working okay. 

```
export const networkEntrySymbol = Symbol.for('networkEntry');

export type NetworkRequest = Request & {
  [networkEntrySymbol]?: NetworkInfo
}
```

vigneshshanmugam:
> Looks great. Do you have any suggestions for testing the feature? Can I run it with heartbeat to check for compatibility with waterfall?

Yeah we need to run it via HB to test this for compatibility
1. Switch to this branch and install it globally `npm i -g .`
2. Run only inline monitors via HB which will use this version.

src/plugins/network.ts
vigneshshanmugam: Math.floor, I can switch to that.

src/plugins/network.ts
vigneshshanmugam: `networkEntry.response` would be null for failed reqs.

vigneshshanmugam:
> Aren't we displaying queuing time in waterfal chart, if so, this seems like a breaking change.

I wouldnt call it as breaking change, if its not sent to UI, it shouldn't show those timings.

dominiqueclarke (commented):
We are expecting `requestStartTime` and `loadEndTime` to be chrome's special baseline in seconds value. It appears to have now been changed to epoch millis in this PR. This will require a change on the Kibana side to be compatible with this version of synthetics, unless you are able to convert those two values back to the way chrome was providing it.

vigneshshanmugam:
> We are expecting requestStartTime and loadEndTime to be chrome's special baseline in seconds value

I remember we were using this on the UI only for calculating the offset, Why does this have to be in a special format? I can surely workaround to not break the UI, but would like to know the reason so we can solve this in a better way to never break on the future iteration.

dominiqueclarke:
> > We are expecting requestStartTime and loadEndTime to be chrome's special baseline in seconds value
> 
> I remember we were using this on the UI only for calculating the offset, Why does this have to be in a special format? I can surely workaround to not break the UI, but would like to know the reason so we can solve this in a better way to never break on the future iteration.

We are using it for the offset, but before we use it we actually covert it to milliseconds. This allows accurately place the start of each request on the chart in milliseconds. So the code is expecting seconds to convert to milliseconds.

vigneshshanmugam:
@dominiqueclarke So if I can change the epochTime in Seconds then it should be good from the UI side? We dont need it to be in the chrome devtools monotonic time format?

dominiqueclarke:
> monotonic time format?

Yes, I believe that should work just fine.

dominiqueclarke (approved):
LGTM. Timing changes look good with Kibana waterfall chart. 

Screen Shot 2021-10-05 at 1 00 29 PM
Screen Shot 2021-10-05 at 1 00 17 PM

dominiqueclarke:
Timings and headers look good as well. 
Screen Shot 2021-10-06 at 10 42 25 AM

src/plugins/network.ts
```diff
@@ -23,336 +23,234 @@
...
+        mimeType: 'x-unknown',
+        headers: {},
+        redirectURL: '',
+      },
+      isNavigationRequest: request.isNavigationRequest(),
+      status: -1,
       loadEndTime: -1,
       responseReceivedTime: -1,
-      response: null,
-      resourceSize: 0,
```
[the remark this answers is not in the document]
vigneshshanmugam: Setting it to 0 would solve it? I had a thought since 0 is a actual size and could be for cached resource.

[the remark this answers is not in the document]
vigneshshanmugam: Gotcha, will revert it to zero. May be we can add a fix on the UI side in parallel. Do you mind creating an issue?

Issue #226, opened 2021-02-23
capture unhandled exception on tested pages


vigneshshanmugam:
+ we only capture JS errors console error and warnings which are captured via `console.error` and `console.warn`.

+ Need to capture uncaught exceptions that gets thrown which would be useful for debugging and associate it with each steps. 

We can use `page.on('pageerror')`  events - https://playwright.dev/docs/api/class-page/#pageonpageerror

dominiqueclarke:
Seeing this implemented all the way through to Kibana for Synthetics agent `beta.16`, when tested against Kibana branch 7.16 and commit `0c347055e667851383a7415dfeaaa985bc01fe07`
<img width="1501" alt="Screen Shot 2021-11-01 at 4 27 12 PM" src="https://user-images.githubusercontent.com/11356435/139738056-b2a749ea-7111-4bf5-a85f-e30767a3a2a2.png">

PR #374, opened 2021-09-13
Capture and log page errors

shahzad31:
Fixes https://github.com/elastic/synthetics/issues/226

Add an event handler in browser console for "pageerror".

Capturing `error.message` and `error.stack.toString()`


src/plugins/browser-console.ts
```diff
@@ -54,12 +54,31 @@ export class BrowserConsole {
...
+    if (!this._currentStep) {
+      return;
+    }
+    const { name, index } = this._currentStep;
+    this.messages.push({
+      timestamp: getTimestamp(),
+      text: error.message,
+      type: 'error',
+      step: { name, index },
+      stacktrace: error.stack.toString(),
```
vigneshshanmugam, line 67: pass it as error, we have a global formatError function that handles them properly in the console event - https://github.com/elastic/synthetics/blob/e67a91dacc4c981aeb6b0f9d715f07b059da8f80/src/reporters/json.ts#L486-L495

src/plugins/browser-console.ts
```diff
@@ -54,12 +54,31 @@ export class BrowserConsole {
...
+    });
+    if (this.messages.length > defaultMessageLimit) {
```
vigneshshanmugam, line 69: Move this logic to stop to avoid doing at multiple places.

src/common_types.ts
```diff
@@ -119,6 +119,7 @@ export type NetworkInfo = {
 export type BrowserMessage = {
   text: string;
   type: string;
+  stacktrace?: string;
```
vigneshshanmugam, line 122: would become error, see the formatError comment.

vigneshshanmugam (commented):
Thanks for working on this, tests can be written in `browser-console.test.ts`, where you can throw some promise errors inside the evaluate function. 
```
    await page.evaluate(() => Promise.reject(new Error("Boom")));
    // make sure to add some stack here and do the needful. 
```

Let me know if you need any help.

shahzad31:
@vigneshshanmugam i have added a unit test

src/reporters/json.ts
```diff
@@ -483,13 +483,17 @@ export default class JSONReporter extends BaseReporter {
...
         if (browserconsole) {
-          browserconsole.forEach(({ timestamp, text, type, step }) => {
+          browserconsole.forEach(({ timestamp, text, type, step, error }) => {
             this.writeJSON({
               type: 'journey/browserconsole',
               journey,
               timestamp,
               step,
-              payload: { text, type } as Payload,
+              payload: {
```
vigneshshanmugam, line 492: you can pass error as a standalone field here. will be formatted correctly in write handler.

__tests__/plugins/browser-console.test.ts
```diff
@@ -55,4 +55,49 @@ describe('BrowserConsole', () => {
...
+  it('should capture browser page errors', async () => {
+    const driver = await Gatherer.setupDriver({ wsEndpoint });
+    const browserConsole = new BrowserConsole(driver);
+    const { page } = driver;
+    browserConsole.start();
+    await page.goto(server.TEST_PAGE);
+    browserConsole._currentStep = { name: 'step-name', index: 0 };
+    const bodyHandle = await page.$('body');
+    try {
+      await page.evaluate(
```
vigneshshanmugam, line 68: You can do await page.setContent to make this simpler.

__tests__/plugins/browser-console.test.ts
```diff
@@ -55,4 +55,49 @@ describe('BrowserConsole', () => {
...
+    const messages = browserConsole.stop();
+    await Gatherer.stop();
+    const notFoundMessage = messages.find(
+      m => m.text.indexOf('Failed to load resource:') >= 0
+    );
+    expect(notFoundMessage.text).toEqual(
+      `Failed to load resource: the server responded with a status of 404 (Not Found)`
+    );
+    expect(notFoundMessage.type).toEqual('error');
+    expect(notFoundMessage.timestamp).toBeDefined();
```
vigneshshanmugam, line 90: nit: we can remove these timestamp checks and keep only required ones.

vigneshshanmugam (commented):
Looking good, can you add a test for promise rejection as well.

src/reporters/json.ts
shahzad31: write handler will get only text from the error, no? if we want the stack trace implicitely we will have to get it like it's done here...

shahzad31:
> Looking good, can you add a test for promise rejection as well.

@vigneshshanmugam promoise rejections doesnt let the test run. I think current test cover enough scenerios.

vigneshshanmugam:
> @vigneshshanmugam promoise rejections doesnt let the test run. I think current test cover enough scenerios.

Probably its a problem then, The intention of the original issue was to capture unhandled exceptions. Will merge once i test it out.

src/reporters/json.ts
vigneshshanmugam: The payload would contain `type and text` as it was previously and you would write `errror` in top level which would be picked up here - https://github.com/elastic/synthetics/blob/8c502cfa027781f4031829d97763ac3ad3496cf4/src/reporters/json.ts#L589

__tests__/plugins/browser-console.test.ts
shahzad31: done

src/reporters/json.ts
shahzad31: got it, i have pushed the update

vigneshshanmugam (approved):

PR #379, opened 2021-09-16
Re-add support for suite params

andrewvc:
Heartbeat 7.15/7.x did not get https://github.com/elastic/beats/pull/26674 backported to them. Since they will be released soon we need to work on re-supporting the `--suite-params` flag and do a new release once this is merged.

### DIFF
--- a/src/common_types.ts
+++ b/src/common_types.ts
@@ -157,6 +157,9 @@ export type CliArgs = {
   debug?: boolean;
   ignoreHttpsErrors?: boolean;
   params?: Params;
+  /**
+   * @deprecated use params instead
+   */
   suiteParams?: Params;
   richEvents?: boolean;
 };
--- a/src/common_types.ts
+++ b/src/common_types.ts
@@ -157,6 +157,7 @@ export type CliArgs = {
   debug?: boolean;
   ignoreHttpsErrors?: boolean;
   params?: Params;
+  suiteParams?: Params;
   richEvents?: boolean;
 };
 
--- a/src/parse_args.ts
+++ b/src/parse_args.ts
@@ -38,8 +38,13 @@ program
     'configuration path (default: synthetics.config.js)'
   )
   .option(
-    '-p, --params <jsonstring>',
+    '-p, --params <jsonstring>, --suite-params <jsonstring>',
     'JSON object that gets injected to all journeys',
+    JSON.parse
+  )
+  .option(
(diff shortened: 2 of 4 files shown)


dominiqueclarke (approved):
code LGTM.

src/parse_args.ts
```diff
@@ -38,10 +38,15 @@ program
     'configuration path (default: synthetics.config.js)'
   )
   .option(
-    '-p, --params <jsonstring>',
+    '-p, --params <jsonstring>, --suite-params <jsonstring>',
```
vigneshshanmugam, line 41: I am confused with this notation, What does this do?

src/parse_args.ts
```diff
@@ -144,5 +149,9 @@ if (options.capability) {
   }
 }
 
+if (options.suiteParams && !options.params) {
```
vigneshshanmugam, line 152: We can revert to the previous way - https://github.com/vigneshshanmugam/synthetics/blob/b1ba055bee6074ff96f22a4aebda9cfeb22bc362/src/cli.ts#L168-L172

src/parse_args.ts
andrewvc: Oops, let me remove it. It does nothing. You can't add a third form of the command, I needed to add a new option below.

vigneshshanmugam (commented):
Overall I feel like we can take the changes from the previous PR https://github.com/elastic/synthetics/pull/331/files and add the test we have here.

src/common_types.ts
```diff
@@ -157,6 +157,7 @@ export type CliArgs = {
   debug?: boolean;
   ignoreHttpsErrors?: boolean;
   params?: Params;
+  suiteParams?: Params;
```
vigneshshanmugam, line 160: nit: We can add `@deprecated notation`

```
 /**
   * @deprecated use `params` instead
   */
```

src/parse_args.ts
```diff
@@ -42,6 +42,11 @@ program
     'JSON object that gets injected to all journeys',
     JSON.parse
   )
+  .option(
+    '-s, --suite-params <jsonstring>, --suite-params <jsonstring>',
```
vigneshshanmugam, line 46: nit: forgot to remove third args.

vigneshshanmugam (approved):
LGTM

Issue #232, opened 2021-03-01
Elastic Synthetic Identifier In UA String


paulb-elastic:
As a Synthetics user
I want to be able to easily identify Elastic Synthetic traffic in my analytics
So that I can exclude such traffic from my visitor volumes as this monitoring skews real customer figures

ACs:
* The user agent string should include some kind of addition to identify monitors running from Elastic Synthetics (appending `Elastic/Synthetics` to the actual UA string)
* It must still be possible to identify the actual browser making the requests (i.e. it won't remove or overwrite the *natural* UA string completely)
* There must be a configuration parameter (boolean) associated with the monitor to control this behaviour, defaulted to enabled (needed for the rare cases where a non standard UA string may cause problems with UA sniffing)

Referencing somewhat related tickets:
* https://github.com/elastic/synthetics/issues/151
* https://github.com/elastic/uptime/issues/302

dominiqueclarke:
Seeing `Elastic/Synthetics` appended in the user agent when run in headful mode on version `beta-16`. 

<img width="1281" alt="Screen Shot 2021-11-01 at 4 10 08 PM" src="https://user-images.githubusercontent.com/11356435/139735523-41e4db82-2fa8-4a45-8847-3534e4df76e6.png">

PR #380, opened 2021-09-20
Append Elastic/synthetics to UA string

shahzad31:
Fixes https://github.com/elastic/synthetics/issues/232

Appends Elastic/Synthetics to UA string. 

This PR uses a pretty ugly way to determin current user agent string and append the desired string into it.

Not sure if there is a way better way to do this. Struggling with playwright docs on how to get the current user agent string without opening a browser context.

### DIFF
--- a/__tests__/core/gatherer.test.ts
+++ b/__tests__/core/gatherer.test.ts
@@ -28,13 +28,11 @@ import { PluginManager } from '../../src/plugins';
 import { wsEndpoint } from '../utils/test-config';
 import { devices } from 'playwright-chromium';
 import { Server } from '../utils/server';
-import exp = require('constants');
 
 jest.mock('../../src/plugins/network');
 
 describe('Gatherer', () => {
   let server: Server;
-
   beforeAll(async () => {
     server = await Server.create();
   });
@@ -67,11 +65,9 @@ describe('Gatherer', () => {
   describe('Append Elastic/Synthetics to UA', () => {
     it('works on a single page', async () => {
       const driver = await Gatherer.setupDriver({ wsEndpoint });
-
       expect(await driver.page.evaluate(() => navigator.userAgent)).toContain(
         ' Elastic/Synthetics'
       );
-
       await Gatherer.dispose(driver);
       await Gatherer.stop();
     });
@@ -81,11 +77,9 @@ describe('Gatherer', () => {
         wsEndpoint,
         playwrightOptions: { ...devices['Galaxy S9+'] },
       });
-
       expect(await driver.page.evaluate(() => navigator.userAgent)).toContain(
         ' Elastic/Synthetics'
       );
-
       await Gatherer.dispose(driver);
       await Gatherer.stop();
     });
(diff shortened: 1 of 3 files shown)


vigneshshanmugam (changes requested):
Even though this works, This would be the last approach we would need to do as it slows down the startup time of our runner. Please look in to - https://chromedevtools.github.io/devtools-protocol/tot/Browser/#method-getVersion

which can be used to get the version and can be used for setting the global user agent context. Let me know if you need any help.

vigneshshanmugam:
We discussed offline and decided to go with using browser based session approach to get the user agent. 

Benchmarks if anyone is curious
```
Session: 3.505ms
Context: 245.072ms
```

shahzad31:
i tested manualy, it seems to work with iframe and new window as well.

will try to write some tests for those use cases.

shahzad31:
Added tests for iframe and popup

src/core/gatherer.ts
```diff
@@ -46,12 +48,26 @@ export class Gatherer {
...
     }
-    const context = await Gatherer.browser.newContext(playwrightOptions);
+
+    const context = await Gatherer.browser.newContext({
+      ...playwrightOptions,
+      userAgent: await Gatherer.getUserAgent(),
+    });
+
     const page = await context.newPage();
+
```
vigneshshanmugam, line 58: nit: can we get rid of these empty spaces? both in tests and code? would like to keep it similar to existing code.

__tests__/core/gatherer.test.ts
```diff
@@ -54,4 +63,84 @@ describe('Gatherer', () => {
...
+        ' Elastic/Synthetics'
+      );
+
+      expect(await page1.textContent('body')).toEqual('Not found');
+
+      await Gatherer.dispose(driver);
+      await Gatherer.stop();
+    });
+
+    it('works with iframe content', async () => {
```
vigneshshanmugam, line 122: I would remove this iframe test in favor of request test. Or merge it on the previous popup test to look for request to `popup.html` and check if request.headers.userAgent contains the Elastic identifier. 

Iframes tests are unnecessary as they inherit the parent document.

__tests__/core/gatherer.test.ts
```diff
@@ -26,10 +26,19 @@
...
+import { devices } from 'playwright-chromium';
+import { Server } from '../utils/server';
+import exp = require('constants');
 
 jest.mock('../../src/plugins/network');
 
 describe('Gatherer', () => {
+  let server: Server;
+
+  beforeAll(async () => {
```
vigneshshanmugam, line 38: Close the server on afterAll.

__tests__/core/gatherer.test.ts
```diff
@@ -26,10 +26,19 @@
 import { Gatherer } from '../../src/core/gatherer';
 import { PluginManager } from '../../src/plugins';
 import { wsEndpoint } from '../utils/test-config';
+import { devices } from 'playwright-chromium';
+import { Server } from '../utils/server';
+import exp = require('constants');
```
vigneshshanmugam, line 31: bad import.

src/core/gatherer.ts
```diff
@@ -27,7 +27,9 @@ import { chromium, ChromiumBrowser } from 'playwright-chromium';
 import { PluginManager } from '../plugins';
 import { RunOptions } from './runner';
 import { log } from './logger';
-import { Driver } from '../common_types';
+import { Driver, PlaywrightOptions } from '../common_types';
+import { Protocol } from 'playwright-chromium/types/protocol';
```
vigneshshanmugam, line 31: need to remove below two imports

__tests__/core/gatherer.test.ts
```diff
@@ -54,4 +63,84 @@ describe('Gatherer', () => {
     expect(network.start).toHaveBeenCalled();
     await Gatherer.stop();
   });
+
+  describe('Append Elastic/Synthetics to UA', () => {
```
vigneshshanmugam, line 67: nit: Elastic UA identifier

vigneshshanmugam (commented):
Generally looks good, good to go after removing some bad import statements.

src/core/gatherer.ts
shahzad31: i really like spaced out code :D

vigneshshanmugam: I see 😄

vigneshshanmugam (approved):
LGTM 🎉

PR #383, opened 2021-09-23
feat: Add network conditions cli args with sensible defaults

### LINKED ISSUE #284: Sensible defaults for network throttling
Although the more rounded solution for network throttling is defined in https://github.com/elastic/synthetics/issues/210, this is to at least provide sensible defaults for all monitors.

As a Synthetic user
I want all monitors to run with a sensible network throttling configuration
So that they don’t run at line speed and result in variable performance metrics based on how busy the network is

ACs:
* Determine a reasonable default network speed (upload speed, download speed and latency)
* Apply these to all monitors that run in Synthetics (inline, suites, `heartbeat.yml`, Fleet Integration)
* It is possible to override these defaults for specific monitors (e.g. via the script as shown in https://github.com/elastic/synthetics/issues/210#issuecomment-790502321)

Heartbeat issue here: https://github.com/elastic/beats/issues/27952

dominiqueclarke:
Adds network conditions cli args with sensible defaults
* fix #284 
* `--upload-throughput`: default 5Mbps
* `--download-throughput`: default 3Mbps
* `--latency`: default 20ms
* `--offline`: default `false`
* `--no-throttling`: disables all throttling, including sensible defaults :

### DIFF
--- a/__tests__/cli.test.ts
+++ b/__tests__/cli.test.ts
@@ -26,6 +26,7 @@
 import { ChildProcess, spawn } from 'child_process';
 import { join } from 'path';
 import { Server } from './utils/server';
+import { networkConditionDefaults, megabytesToBytes } from '../src/helpers';
 
 describe('CLI', () => {
   let server: Server;
@@ -353,6 +354,12 @@ describe('CLI', () => {
 
   describe('throttling', () => {
     let cliArgs: Array<string>;
+    const defaultNetworkConditions = {
+      offline: false,
+      uploadThroughput: megabytesToBytes(networkConditionDefaults.upload),
+      downloadThroughput: megabytesToBytes(networkConditionDefaults.download),
+      latency: networkConditionDefaults.latency,
+    }
 
     beforeAll(async () => {
       cliArgs = [
@@ -381,26 +388,17 @@ describe('CLI', () => {
       await cli.waitFor('synthetics/metadata');
       const journeyStartOutput = JSON.parse(cli.output());
       expect(await cli.exitCode).toBe(0);
-      expect(journeyStartOutput.payload).toHaveProperty('networkConditions', {
-        downloadThroughput: 1024 * 1024 * 5, 
-        latency: 20, 
-        offline: false, 
-        uploadThroughput: 1024 * 1024 * 3
-      });
+      expect(journeyStartOutput.payload).toHaveProperty('networkConditions', defaultNetworkConditions);
     });
 
     it('applies custom throttling', async () => {
-      const downloadThroughput = 1024 * 1024 * 3;
-      const uploadThroughput = 1024 * 1024 * 1;
+      const downloadThroughput = megabytesToBytes(3);
+      const uploadThroughput = megabytesToBytes(1);
       const latency = 30;
       const cli = new CLIMock()
         .args(cliArgs.concat([
-          '--download-throughput',
-          String(downloadThroughput),
-          '--upload-throughput',
-          String(uploadThroughput),
-          '--latency',
-          String(latency),
+          '--throttle',
+          '3d,1u,30l',
         ]))
         .run();
       await cli.waitFor('synthetics/metadata');
@@ -414,16 +412,42 @@ describe('CLI', () => {
       });
     });
 
-    it('allows offline', async () => {
+    it('applies custom throttling order agnostic', async () => {
+      const downloadThroughput = megabytesToBytes(3);
+      const uploadThroughput = megabytesToBytes(1);
+      const latency = 30;
       const cli = new CLIMock()
         .args(cliArgs.concat([
-          '--offline'
+          '--throttle',
+          '1u,30l,3d',
         ]))
         .run();
-      expect(await cli.exitCode).toBe(1);
-      expect(JSON.parse(cli.output()).error.message).toEqual(
-        expect.stringContaining('net::ERR_INTERNET_DISCONNECTED')
-      );
+      await cli.waitFor('synthetics/metadata');
+      const journeyStartOutput = JSON.parse(cli.output());
+      expect(await cli.exitCode).toBe(0);
+      expect(journeyStartOutput.payload).toHaveProperty('networkConditions', {
+        ...defaultNetworkConditions,
+        downloadThroughput, 
+        latency, 
+        uploadThroughput
+      });
+    });
+
+    it('uses default throttling when specific params are not provided', async () => {
+      const downloadThroughput = 1024 * 1024 * 2;
+      const cli = new CLIMock()
+        .args(cliArgs.concat([
+          '--throttle',
+          '2d',
+        ]))
+        .run();
+      await cli.waitFor('synthetics/metadata');
+      const journeyStartOutput = JSON.parse(cli.output());
+      expect(await cli.exitCode).toBe(0);
+      expect(journeyStartOutput.payload).toHaveProperty('networkConditions', {
+        ...defaultNetworkConditions,
+        downloadThroughput,
+      });
     });
   });
 });
--- a/__tests__/cli.test.ts
+++ b/__tests__/cli.test.ts
@@ -368,7 +368,7 @@ describe('CLI', () => {
       const cli = new CLIMock()
         .args(cliArgs.concat(['--no-throttling']))
         .run();
-      await cli.waitFor('journey/start');
+      await cli.waitFor('synthetics/metadata');
       const journeyStartOutput = JSON.parse(cli.output());
       expect(await cli.exitCode).toBe(0);
       expect(journeyStartOutput.payload).not.toHaveProperty('networkConditions');
@@ -378,7 +378,7 @@ describe('CLI', () => {
       const cli = new CLIMock()
         .args(cliArgs)
         .run();
-      await cli.waitFor('journey/start');
+      await cli.waitFor('synthetics/metadata');
       const journeyStartOutput = JSON.parse(cli.output());
       expect(await cli.exitCode).toBe(0);
       expect(journeyStartOutput.payload).toHaveProperty('networkConditions', {
@@ -403,7 +403,7 @@ describe('CLI', () => {
           String(latency),
         ]))
         .run();
-      await cli.waitFor('journey/start');
+      await cli.waitFor('synthetics/metadata');
       const journeyStartOutput = JSON.parse(cli.output());
       expect(await cli.exitCode).toBe(0);
       expect(journeyStartOutput.payload).toHaveProperty('networkConditions', {
--- a/__tests__/cli.test.ts
+++ b/__tests__/cli.test.ts
@@ -378,7 +378,7 @@ describe('CLI', () => {
       await cli.waitFor('synthetics/metadata');
       const journeyStartOutput = JSON.parse(cli.output());
       expect(await cli.exitCode).toBe(0);
-      expect(journeyStartOutput.payload).not.toHaveProperty('networkConditions');
+      expect(journeyStartOutput.payload).not.toHaveProperty('network_conditions');
     });
 
     it('applies default throttling', async () => {
@@ -388,7 +388,7 @@ describe('CLI', () => {
       await cli.waitFor('synthetics/metadata');
       const journeyStartOutput = JSON.parse(cli.output());
       expect(await cli.exitCode).toBe(0);
-      expect(journeyStartOutput.payload).toHaveProperty('networkConditions', defaultNetworkConditions);
+      expect(journeyStartOutput.payload).toHaveProperty('network_conditions', defaultNetworkConditions);
     });
 
     it('applies custom throttling', async () => {
@@ -404,7 +404,7 @@ describe('CLI', () => {
       await cli.waitFor('synthetics/metadata');
       const journeyStartOutput = JSON.parse(cli.output());
       expect(await cli.exitCode).toBe(0);
-      expect(journeyStartOutput.payload).toHaveProperty('networkConditions', {
+      expect(journeyStartOutput.payload).toHaveProperty('network_conditions', {
         downloadThroughput, 
         latency, 
         offline: false, 
@@ -425,7 +425,7 @@ describe('CLI', () => {
       await cli.waitFor('synthetics/metadata');
       const journeyStartOutput = JSON.parse(cli.output());
       expect(await cli.exitCode).toBe(0);
-      expect(journeyStartOutput.payload).toHaveProperty('networkConditions', {
+      expect(journeyStartOutput.payload).toHaveProperty('network_conditions', {
         ...defaultNetworkConditions,
         downloadThroughput, 
         latency, 
@@ -444,7 +444,7 @@ describe('CLI', () => {
       await cli.waitFor('synthetics/metadata');
       const journeyStartOutput = JSON.parse(cli.output());
       expect(await cli.exitCode).toBe(0);
-      expect(journeyStartOutput.payload).toHaveProperty('networkConditions', {
+      expect(journeyStartOutput.payload).toHaveProperty('network_conditions', {
         ...defaultNetworkConditions,
         downloadThroughput,
       });
--- a/__tests__/cli.test.ts
+++ b/__tests__/cli.test.ts
@@ -350,6 +350,82 @@ describe('CLI', () => {
       );
     });
   });
+
+  describe('throttling', () => {
+    let cliArgs: Array<string>;
+
+    beforeAll(async () => {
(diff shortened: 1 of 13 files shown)


__tests__/core/gatherer.test.ts
```diff
@@ -69,6 +69,7 @@ describe('Gatherer', () => {
   describe('Elastic UA identifier', () => {
     it('works on a single page', async () => {
       const driver = await Gatherer.setupDriver({ wsEndpoint });
+              // @ts-ignore
```
vigneshshanmugam, line 72: any clue why ts error shows up for you?

src/core/gatherer.ts
```diff
@@ -52,6 +57,9 @@ export class Gatherer {
     });
     const page = await context.newPage();
     const client = await context.newCDPSession(page);
+    if (networkConditions) {
+      await client.send('Network.emulateNetworkConditions', networkConditions);
```
vigneshshanmugam, line 61: This session is only for page level and wont work for new tabs/windows. You need to use something like this 
https://github.com/elastic/synthetics/blob/a621b530a9559538c5d52eefbaabb8c9c9e62c8e/src/core/gatherer.ts#L67

I would move that code inside and reuse it for useragent and network throttling.

src/core/runner.ts
```diff
@@ -320,8 +325,8 @@ export default class Runner extends EventEmitter {
   registerJourney(journey: Journey, context: JourneyContext) {
     this.currentJourney = journey;
     const timestamp = getTimestamp();
-    const { params } = context;
-    this.emit('journey:start', { journey, timestamp, params });
+    const { params, networkConditions } = context;
+    this.emit('journey:start', { journey, timestamp, params, networkConditions });
```
vigneshshanmugam, line 329: I don't think we need it for each journey as we dont allow users to modify it per journey, instead it should be only for top level event `start` which is for every run. https://github.com/elastic/synthetics/blob/a621b530a9559538c5d52eefbaabb8c9c9e62c8e/src/core/runner.ts#L437

and we write it as part of `synthetics/metadata` event.

src/parse_args.ts
```diff
@@ -104,6 +111,20 @@ program
...
+  .addOption(
+    new Option('--upload-throughput <throughput>', 'Maximal aggregated upload throughput (bytes/sec). -1 disables upload throttling.')
+      .default(defaultNetworkConditions.uploadThroughput)
+  )
+  .addOption(
+    new Option('--latency <latency>', 'Minimum latency from request sent to response headers received (ms).')
+      .default(defaultNetworkConditions.latency)
+  )
+  .option('--offline', 'True to emulate internet disconnection.', defaultNetworkConditions.offline)
+  .option('--no-throttling', 'Turns off default throttling.')
```
vigneshshanmugam, line 127: except for this option `no-throttling`, I would remove every other arg from CLI and only allow them to be configured only via synthetics config file if users needs customization.

src/reporters/json.ts
```diff
@@ -368,12 +370,12 @@ export default class JSONReporter extends BaseReporter {
       });
     });
 
-    this.runner.on('journey:start', ({ journey, timestamp, params }) => {
+    this.runner.on('journey:start', ({ journey, timestamp, params, networkConditions }) => {
```
vigneshshanmugam, line 373: moves to `synthetics/metadata` event.

__tests__/cli.test.ts
```diff
@@ -350,6 +350,82 @@ describe('CLI', () => {
...
+      expect(journeyStartOutput.payload).toHaveProperty('networkConditions', {
+        downloadThroughput: 1024 * 1024 * 5, 
+        latency: 20, 
+        offline: false, 
+        uploadThroughput: 1024 * 1024 * 3
+      });
+    });
+
+    it('applies custom throttling', async () => {
+      const downloadThroughput = 1024 * 1024 * 3;
```
vigneshshanmugam, line 393: nit: String(number)

__tests__/core/gatherer.test.ts
```diff
@@ -116,4 +117,25 @@ describe('Gatherer', () => {
       await Gatherer.stop();
     });
   });
+
+  describe('Network emulation', () => {
+    it('applies network throttling', async () => {
```
vigneshshanmugam, line 122: would add one more test to check the network condition on subsequent window or popup.

vigneshshanmugam (commented):
Added some feedback, but the overall approach looks good. Thanks for picking it up 🎉

src/parse_args.ts
dominiqueclarke: @vigneshshanmugam What about configuring via heartbeat, especially for inline scripts. Do we want to make this feature available for inline scripts?

src/core/gatherer.ts
dominiqueclarke: `Network.emulateNetworkConditions` is only available on the browser session, not the page session. So I cannot set the the network conditions on the browser CDP session. To mitigate this and ensure it works on popups, I've added https://github.com/elastic/synthetics/pull/383/commits/ec3816fde03c78b328ffdd4efb34b5a2d901e17a which ensures that network conditions are set on new pages and adds a test for popups.

vigneshshanmugam: Interesting, Sure that page event workaround seems to be fine.

src/parse_args.ts
vigneshshanmugam: We need a way to solve this going forward with synthetics config file from HB for inline suites as we cannot keep adding CLI for each individual feature needing to be exposed via heartbeat flag. I would remove this from here and add a feature to HB to invoke them via a temp synthetics config file. 

@andrewvc Thoughts?

src/parse_args.ts
```diff
@@ -104,6 +111,20 @@ program
     '--quiet-exit-code',
     'always return 0 as an exit code status, regardless of test pass / fail. Only return > 0 exit codes on internal errors where the suite could not be run'
   )
+  .addOption(
```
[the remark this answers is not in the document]
andrewvc: ++ on keeping them separate. JSON makes sense for `--params` because it involves nesting and structure we can't anticipate, but it's a very weird thing to do. Really like the args we have here.

src/parse_args.ts
```diff
@@ -104,6 +111,20 @@ program
...
+  )
+  .addOption(
+    new Option('--upload-throughput <throughput>', 'Maximal aggregated upload throughput (bytes/sec). -1 disables upload throttling.')
+      .default(defaultNetworkConditions.uploadThroughput)
+  )
+  .addOption(
+    new Option('--latency <latency>', 'Minimum latency from request sent to response headers received (ms).')
+      .default(defaultNetworkConditions.latency)
+  )
+  .option('--offline', 'True to emulate internet disconnection.', defaultNetworkConditions.offline)
```
andrewvc, line 126: Is this to disable the internet or to disable all network connections? The way this is written it sounds like the LAN would still work. I'm pretty sure that's not the case, so we should clarify and `s/internet/network`

src/parse_args.ts
andrewvc: It's an interesting question, I was actually thinking we could keep adding flags. I don't see why not? What's the downside? We still don't have nearly as many flags as, say, [cURL](https://linux.die.net/man/1/curl). 

One nice property of flags is that they break if the version of synthetics doesn't support them. This means that if a user intends to override a default that only works if the version of synthetics supports it. If it does not it fails loudly. 

A config file would probably fail silently, which IMHO is less desirable.

src/parse_args.ts
dominiqueclarke: You're right, I've updated it. Interesting enough, I pulled the descriptions straight from the devtools docs

https://chromedevtools.github.io/devtools-protocol/tot/Network/#method-emulateNetworkConditions

src/parse_args.ts
vigneshshanmugam: There is nothing wrong in adding extra flags, But this nifty feature which is related to a small change results in 5 flags is not a desirable thing to do IMO. We should stick with a strict number of flags and allow fine tuning via config files as they allow more customization. 

> One nice property of flags is that they break if the version of synthetics doesn't support them.

We can do the same with config files. Its just a implementation detail as of now.

andrewvc: One sort of middle path would be to combine the throttling options into a mini syntax like: `--throttle="5d/3u/2l"`.

I think you have a good point @v re: too many options being a bit unwieldy UX-wise in terms of having to wade through them, so this may strike a good balance.

Re: the idea of using a file for options, one really important property of CLI args is that replicating the commands heartbeat invokes is as simple as copy/pasting the heartbeat log line which shows the full command invoked by heartbeat. Having to replicate the config file too would make debugging tricky in support situations.

andrewvc: As another note, we could parse the throttle config in any order, just split on `/` and look at the trailing letter to see what it's affecting.

src/parse_args.ts
vigneshshanmugam: Can we kill this offline flag for now and add this later if we have a new feature request?

src/parse_args.ts
dominiqueclarke: @vigneshshanmugam I know we discussed parsing it based on order, but I think making it order agnostic makes sense for cases in which you want to override one value while leaving other defaults. Otherwise if order mattered we'd have to have some token for skipping a value or the users would have to know what the defaults are and always provide every value.

andrewvc: ++ to order agnostic, it's more clear and not much more work

src/parse_args.ts
andrewvc: @vigneshshanmugam can you share your reasoning for removing this? I'm ++ on removing it myself because it's beyond the scope of the original issue, and I think we shouldn't add options unless we have a clear use-case behind it.

The only use case I can think of is testing PWAs, but then maybe that should be on a per-journey basis? Either way, I think it could use some refinement

vigneshshanmugam: Apologies for not posting about the case, Yes its a pretty nifty use-case when it comes to testing web apps and only applies to Web apps that uses Service workers (PWA). We can introduce this flag if we ever need to support that use-case in the future and also its not in the scope of this PR.

src/parse_args.ts
vigneshshanmugam: ++ Sounds good to me.

src/parse_args.ts
```diff
@@ -121,6 +130,10 @@ if (options.richEvents) {
   options.quietExitCode = true;
 }
 
+if (options.throttling) {
```
vigneshshanmugam, line 133: not relevant as its controlled in `cli.ts` already.

src/helpers.ts
```diff
@@ -257,3 +257,50 @@ export const CACHE_PATH = join(cwd, '.synthetics', process.pid.toString());
...
+      case uploadToken:
+        networkConditions.uploadThroughput = Number(value);
+        break;
+      case downloadToken:
+        networkConditions.downloadThroughput = Number(value);
+        break;
+      case latencyToken: 
+        networkConditions.latency = Number(value);
+        break;
+      default:
```
vigneshshanmugam, line 293: dead code. We can remove the default break.

src/helpers.ts
```diff
@@ -257,3 +257,50 @@ export const CACHE_PATH = join(cwd, '.synthetics', process.pid.toString());
 export function getDurationInUs(duration: number) {
   return Math.trunc(duration * 1e6);
 }
+
+export function megabytesToBytes(megabytes: number) {
+  return megabytes * 1024 * 1024;
+}
+
+export const networkConditionDefaults = {
```
vigneshshanmugam, line 265: nit: DEFAULT_NETWORK_CONDITIONS capital case to reason about that its a constant

src/helpers.ts
```diff
@@ -257,3 +257,50 @@ export const CACHE_PATH = join(cwd, '.synthetics', process.pid.toString());
...
+  download: 5, // megabytes/second
+  upload: 3, // megabytes/second
+  latency: 20, // milliseconds
+}
+
+export function parseNetworkConditions(args: string): NetworkConditions {
```
vigneshshanmugam, line 271: The code works correctly, But we can make some improvements to make it easier to understand

1. We could reuse the `DEFAULT_NETWORK_CONDITIONS` as the default value for networkConditions
```
 const networkConditions = { ...DEFAULT_NETWORK_CONDITIONS };
```
2. if we find the  token defined, then use the provided value or keep using the default. 

Advantage of this is that, we dont need to populate empty object and change object shape as we go along, Instead keep the structure same and then `return networkConditions;` only if user specified any options instead of recreating another object like we do now https://github.com/elastic/synthetics/pull/383/files#diff-ca5696b367d474e785317cb7a0d9853fef5729387ab8d0fab4c46268c03aae99R298-R304

src/core/runner.ts
```diff
@@ -104,7 +107,10 @@ type HookType = 'beforeAll' | 'afterAll';
 export type SuiteHooks = Record<HookType, Array<HooksCallback>>;
 
 interface Events {
-  start: { numJourneys: number };
+  start: { 
+    numJourneys: number, 
+    networkConditions: NetworkConditions;
```
vigneshshanmugam, line 112: must be optional as users can use --no-throttling

src/parse_args.ts
```diff
@@ -104,6 +108,11 @@ program
     '--quiet-exit-code',
     'always return 0 as an exit code status, regardless of test pass / fail. Only return > 0 exit codes on internal errors where the suite could not be run'
   )
+  .addOption(
+    new Option('--throttle <throttle>', 'Comma separated list of options to throttle network conditions for download throughput (d) in megabytes/second, upload throughput (u) in megabytes/second and latency (l) in milliseconds. Ex: --throttle "3d,2u,10l"')
```
vigneshshanmugam, line 112: nit: How about `throttling` `no-throttling` for naming? https://github.com/tj/commander.js#other-option-types-negatable-boolean-and-booleanvalue

__tests__/reporters/base.test.ts
```diff
@@ -58,7 +64,10 @@ describe('base reporter', () => {
 
   it('writes each step to the FD', async () => {
     const { stream } = new BaseReporter(runner, { fd: fs.openSync(dest, 'w') });
-    runner.emit('start', { numJourneys: 1 });
+    runner.emit('start', { 
+      numJourneys: 1,
+      networkConditions,
```
vigneshshanmugam, line 69: this goes away once we make it optional.

vigneshshanmugam (commented):
Added some suggestions, LGTM otherwise.

src/helpers.ts
```diff
@@ -257,3 +257,57 @@ export const CACHE_PATH = join(cwd, '.synthetics', process.pid.toString());
...
+
+export function megabytesToBytes(megabytes: number) {
+  return megabytes * 1024 * 1024;
+}
+
+export function bytesToMegabytes(bytes: number) {
+  return bytes / 1024 / 1024;
+}
+
+export const DEFAULT_NETWORK_CONDITIONS = {
```
vigneshshanmugam, line 269: nit: DEFAULT_NETWORK_CONDITIONS: NetworkConditions

src/helpers.ts
```diff
@@ -257,3 +257,57 @@ export const CACHE_PATH = join(cwd, '.synthetics', process.pid.toString());
...
+
+export const DEFAULT_NETWORK_CONDITIONS_ARG = formatNetworkConditionsArgs(DEFAULT_NETWORK_CONDITIONS);
+
+export function parseNetworkConditions(args: string): NetworkConditions {
+  const uploadToken = 'u';
+  const downloadToken = 'd';
+  const latencyToken = 'l';
+  const networkConditions = {
+    ...DEFAULT_NETWORK_CONDITIONS,
+  } as NetworkConditions;
```
vigneshshanmugam, line 291: type casting can be removed with above comment.

src/reporters/json.ts
```diff
@@ -352,12 +354,15 @@ export default class JSONReporter extends BaseReporter {
...
      * could be used for better sharding
      */
-    this.runner.on('start', ({ numJourneys }) => {
+    this.runner.on('start', ({ numJourneys, networkConditions }) => {
       this.writeJSON({
         type: 'synthetics/metadata',
         root_fields: {
           num_journeys: numJourneys,
         },
+        payload: {
```
vigneshshanmugam, line 363: can add a check to undefined to avoid sending empty payload object.

vigneshshanmugam (approved):
LGTM 🎉

PR #385, opened 2021-09-27
feat: introduce more commands to formatter

vigneshshanmugam:
+ Necessary for the recorder assertions
+ Does a regex match instead of a strict match for text content and inner text.
+ Added more assertions like isEditable, isChecked extra that is missing.


justinkambic (approved):
LGTM

PR #387, opened 2021-09-28
feat: enable trace as part of rich-events

### LINKED ISSUE #386: enable trace collection as part of rich events flag
Starting 7.16, we are adding Ui capability to show the step level trace metrics in the Waterfall UI. We need to enable the capability `trace` as part of this effort in the agent side.

vigneshshanmugam:
+ fix #386 
+ As part of `--rich-events` flag, all trace level data will be sent as `step/metrics` events to HB.

### DIFF
--- a/__tests__/cli.test.ts
+++ b/__tests__/cli.test.ts
@@ -113,7 +113,6 @@ describe('CLI', () => {
     const cli = new CLIMock()
       .args([
         join(FIXTURES_DIR, 'example.journey.ts'),
-        join(FIXTURES_DIR, 'error.journey.ts'),
         '--rich-events',
         '--params',
         JSON.stringify(serverParams),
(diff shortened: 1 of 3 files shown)


__tests__/cli.test.ts
```diff
@@ -125,20 +124,19 @@ describe('CLI', () => {
...
       .map(data => JSON.parse(data))
       .find(({ type }) => type === 'step/screenshot_ref');
-    expect(screenshotRef).toMatchObject({
-      journey: {
-        id: 'example journey',
-        name: 'example journey',
-      },
-      root_fields: expect.any(Object),
-    });
+    expect(screenshotRef).toBeDefined();
```
dominiqueclarke, line 127: Why was this removed?

PR #391, opened 2021-10-01
chore: use action title if available in context

vigneshshanmugam:
+ For the recorded use case, when step action is edited and IR is sent over to the formatter, we need to preserve the edited title instead of generating a new one.

### DIFF
--- a/__tests__/formatter/javascript.test.ts
+++ b/__tests__/formatter/javascript.test.ts
@@ -171,4 +171,15 @@ describe('Synthetics JavaScript formatter', () => {
     const formatter = new SyntheticsGenerator(true);
     expect(formatter.generateText(recorderActions)).toMatchSnapshot();
   });
+
+  it('use modified title if available', async () => {
+    const formatter = new SyntheticsGenerator(false);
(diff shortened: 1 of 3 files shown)


__tests__/formatter/javascript.test.ts
```diff
@@ -171,4 +171,15 @@ describe('Synthetics JavaScript formatter', () => {
...
+
+  it('use modified title if available', async () => {
+    const formatter = new SyntheticsGenerator(false);
+    const actions = [
+      {
+        ...recorderActions[1],
+        title: 'Visiting profile',
+      },
+    ];
+    expect(formatter.generateText(actions)).toMatchSnapshot();
```
justinkambic, line 183: I like inline snapshots for smaller outputs like this, but this does keep the file cleaner.

justinkambic (approved):
LGTM

__tests__/formatter/javascript.test.ts
vigneshshanmugam: I don't have a strong preference though.

PR #398, opened 2021-10-13
fix: handle page close during response event

vigneshshanmugam:
+ Trying to access some of the API's like headers, sizes would trigger unhandled error as the underlying page would be closed before the response event is finished. 

**Note: This technique is used by Hartracer in Playwright and we use the same here, we might move to PW HarTracer once its extendable.**

### DIFF
--- a/src/plugins/network.ts
+++ b/src/plugins/network.ts
@@ -23,7 +23,7 @@
  *
  */
 
-import { Request, Response } from 'playwright-chromium';
+import { Page, Request, Response } from 'playwright-chromium';
 import { NetworkInfo, BrowserInfo, Driver } from '../common_types';
 import { Step } from '../dsl';
 import { getTimestamp } from '../helpers';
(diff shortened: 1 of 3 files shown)


src/plugins/network.ts
```diff
@@ -49,11 +49,25 @@ type RequestWithEntry = Request & {
 
 export class NetworkManager {
   private _browser: BrowserInfo;
+  private _barrierPromises = new Set<Promise<void>>();
   results: Array<NetworkInfo> = [];
   _currentStep: Partial<Step> = null;
 
   constructor(private driver: Driver) {}
 
+  private _addBarrier(page: Page, promise: Promise<void>) {
```
dominiqueclarke, line 58: Are we trying to ensure that we collect header information before the page closes, or just prevent the unhandled exceptions in this case, or both?

Reading through the logic, it's clear to me how this would prevent unhandled exceptions from the page closing before the main promise promise is fulfilled. I'm having more trouble understanding the logic for preventing the page from closing before the data is collected, if that's part of this.

vigneshshanmugam: Any async information that we are extracting from the response object or request object should happen during the page lifecycle, 

if the page is closed, the associated information is lost and this barrier enforces that we capture the information on the network entry only during the existence of the page.

dominiqueclarke (approved):
LGTM!

PR #406, opened 2021-11-03
deprecate suite-params option over params

vigneshshanmugam:
+ We have reverted this change - https://github.com/elastic/synthetics/pull/379 for the time being to make 7.15 work, But we added patch to the 7.x branch to add params support in HB and now that we are at the 7.16 release timeline, we can safely remove this from Synthetics side.


dominiqueclarke (approved):
LGTM. We can't test the integration between heartbeat and synthetics for suite tests locally, but the patch has been put in for heartbeat 7.16 so this  should be good to go.

PR #410, opened 2021-11-05
fix: end timestamp should reflect duration

### LINKED ISSUE #409: Mysterious overhead being added between journey start/stop and step start/stop
The `monitor.duration` in the heartbeat summary is calculated based purely on the differences of times between the `journey/start` and `journey/end` events. See https://github.com/elastic/beats/blob/master/x-pack/heartbeat/monitors/browser/synthexec/enrich.go#L167 for the implementation.

In the real world we extra time accounted for between journey start/stop and step start/end events. We should figure out why.

Running a simple journey locally I see:

```
# elastic-synthetics . --match "check if title is present" --reporter json --screenshots off | jq '{t: .type, ts: ."@timestamp"}'
{
  "t": "synthetics/metadata",
  "ts": 1636138571717791.8 // T=-0.282s
}
{
  "t": "journey/start",
  "ts": 1636138572000137.5 // T=0s
}
{
  "t": "step/end",
  "ts": 1636138572231909.8 // T=0.23s
}
{
  "t": "step/end",
  "ts": 1636138572265338.5 // T=0.26s
}
{
  "t": "journey/end",
  "ts": 1636138572265777.2 // T=0.265s
}
```

So, I'm seeing a total of `0.235s` time added as overhead.

vigneshshanmugam:
+ fix #409 

### Problem
Heartbeat monitor calculation is based on the event.timestamp field for journey start and journey end event. Synthetic agent journey/end timestamp is calculated after post processing of the screenshots and network events being written to the JSON output. As a result, Step duration and Monitor duration timings would be hugely different as we are accounting for all the CPU heavy work that comes to extracting and chopping screenshots. 

### Solution

We use the timestamp during journey end and don't account for the time taken for the post processing of other work that synthetics agent is doing under the hood.

### DIFF
--- a/src/core/runner.ts
+++ b/src/core/runner.ts
@@ -123,6 +123,7 @@ interface Events {
     JourneyResult & {
       journey: Journey;
       options: RunOptions;
+      timestamp: number;
     };
   'journey:end:reported': unknown;
   'step:start': { journey: Journey; step: Step };
@@ -149,8 +150,12 @@ export default class Runner extends EventEmitter {
   static screenshotPath = join(CACHE_PATH, 'screenshots');
 
   static async createContext(options: RunOptions): Promise<JourneyContext> {
-    const start = monotonicTimeInSeconds();
     const driver = await Gatherer.setupDriver(options);
+    /**
+     * Do not include browser launch/context creation duration
+     * as part of journey duration
+     */
+    const start = monotonicTimeInSeconds();
     const pluginManager = await Gatherer.beginRecording(driver, options);
     /**
      * For each journey we create the screenshots folder for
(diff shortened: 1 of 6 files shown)


src/core/runner.ts
```diff
@@ -149,8 +150,12 @@ export default class Runner extends EventEmitter {
   static screenshotPath = join(CACHE_PATH, 'screenshots');
 
   static async createContext(options: RunOptions): Promise<JourneyContext> {
-    const start = monotonicTimeInSeconds();
     const driver = await Gatherer.setupDriver(options);
+    /**
```
andrewvc, line 154: Thanks for adding the comment

andrewvc (approved):
LGTM

PR #413, opened 2021-11-10
fix: ensure timestamps of events are inorder

### LINKED ISSUE #411: rewrite screenshots blocks timestamp with creation time
+ Currently screenshot block `timestamp` is based on screenshot blob construction time instead of when the screenshot was taken. As a result, the screenshot timestamp would be greater than the journey/end timestamp which is not true. 

### Wrong order
```
-- step/end timestamp
-- journey/end timestamp

--- screenshot/block timestamp
--- screenshot/block timestamp

-- end timestamp
```

#### Correct order
```
-- screenshot/block timestamp
-- screenshot/block timestamp

-- step/end timestamp
-- journey/end timestamp
-- end timestamp
```

vigneshshanmugam:
+ fix #411


src/reporters/json.ts
```diff
@@ -515,23 +518,25 @@ export default class JSONReporter extends BaseReporter {
...
         type: 'screenshot/block',
+        timestamp: screenshot.timestamp,
         _id: block.id,
         blob: block.blob,
         blob_mime,
       });
     }
     this.writeJSON({
       type: 'step/screenshot_ref',
+      timestamp: screenshot.timestamp,
```
dominiqueclarke, line 537: We are using the document timestamp in the UI. I'm assuming we should move to using this timestamp.

dominiqueclarke (approved):
LGTM

src/reporters/json.ts
vigneshshanmugam: We rewrite it to document timestamp here - https://github.com/elastic/synthetics/blob/926e7d812be37f99e3af60bcc5d946bc3d9e2209/src/reporters/json.ts#L584

PR #420, opened 2021-11-29
Export `ActionInContext` type

justinkambic:
As part of https://github.com/elastic/synthetics-recorder/issues/85, it would be good to depend on the upstream action type instead of maintaining parallel types that might be out of sync. It will take some of the maintenance burden off of the recorder as well.

With this change, when the next release is published we should be able to reference this type and delete our existing action type.


vigneshshanmugam (approved):
LGTM

PR #426, opened 2021-12-23
Fix do not read configuration files for inline journeys unless a --config parameter is passed

### LINKED ISSUE #407: Running --inline inside of a directory with synthetics.config.ts breaks
Spotted by @awahab07  catting a `--inline` script while the `$CWD` was https://github.com/elastic/synthetics-demo/tree/main/todos/synthetics-tests

```
abdul.zahid@Abduls-MacBook-Pro synthetics-tests % cat ../../../beats/x-pack/heartbeat/monitors/test-script.js | elastic-synthetics --inline
Error: Unable to read synthetics config: /Users/abdul.zahid/Desktop/Development/synthetics-demo/todos/synthetics.config.ts
    at readAndParseConfig (/Users/abdul.zahid/.nvm/versions/node/v14.17.0/lib/node_modules/@elastic/synthetics/dist/config.js:74:15)
    at Object.readConfig (/Users/abdul.zahid/.nvm/versions/node/v14.17.0/lib/node_modules/@elastic/synthetics/dist/config.js:47:34)
    at /Users/abdul.zahid/.nvm/versions/node/v14.17.0/lib/node_modules/@elastic/synthetics/dist/cli.js:179:29
    at processTicksAndRejections (internal/process/task_queues.js:95:5)
```

AC:
* Stop reading the `synthetics.config.ts` when using inline monitors

lucasfcosta:
# Summary

Fixes #407.

After this PR, the runner will _not_ try to read `synthetics.config.[t|j]s` files unless it's given a `--config` parameter.

Furthermore, it allows `config` files which use TypeScript to be read when running `inline` tests. Before, it wasn't possible to read `.ts` configuration files for inline tests, which is what initially caused #407.

**⭐  Please read [this comment](https://github.com/elastic/synthetics/issues/407) for an explanation about why this bug happened, why I've allowed the configuration parameter to be passed for `inline` tests, and about how I've fixed reading `.ts` configs in inline tests.**

# How to test this PR

Please follow all the test cases below.

## Not reading `.ts` configuration files with inline tests

1. Checkout this branch and build it (`npm run build`).
2. Create an inline journey like the one below. Save it to `/tmp/example/journey.js`, for example.
    ```js
    step('google', () => {
      page.goto('https://www.google.com');
    });

    step('BBC', () => {
      page.goto('https://www.bbc.co.uk');
    });
    ```
3. Create a configuration file next to your `journey.js` example file and name it `synthetics.config.ts`. This file should have the following content:
    ```js
    export default () => {
      console.log("Config loaded!");
      return {};
    };
    ```
4. Navigate to `/tmp/example` and run your jurney from within there:
    ```
    # Make sure to update the path to `cli.js` according to where your clone of this repo is
    cat /tmp/example.js | node ~/Repositories/synthetics/dist/cli.js --inline
    ```
    N.b. It's important that you run this command from within that particular folder, as the `cwd` matters in this test.
5. You'll see that the configuration is not read because `Config loaded!` (the string in the config file) is not logged to the console.

<br>

## Reading `.ts` configuration files with inline tests when passing the `--config` flag

1. Checkout this branch and build it (`npm run build`).
2. Create an inline journey like the one below. Save it to `/tmp/example/journey.js`, for example.
    ```js
    step('google', () => {
      page.goto('https://www.google.com');
    });

    step('BBC', () => {
      page.goto('https://www.bbc.co.uk');
    });
    ```
3. Create a configuration file next to your `journey.js` example file and name it `synthetics.config.ts`. This file should have the following content:
    ```js
    export default () => {
      console.log("Config loaded!");
      return {};
    };
    ```
4. Navigate to `/tmp/example` and run your jurney from within there **and pass the `--config` parameter pointing to your `synthetics.config.ts` file:
    ```
    # Make sure to update the path to `cli.js` according to where your clone of this repo is
    cat /tmp/example.js | node ~/Repositories/synthetics/dist/cli.js --inline --config /tmp/example/synthetics.config.ts
    ```
    N.b. It's important that you run this command from within that particular folder, as the `cwd` matters in this test.
5. You'll see that the configuration is not read because `Config loaded!` (the string in the config file) _is_ logged to the console.

### DIFF
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -126,6 +126,25 @@ async function prepareSuites(inputs: string[]) {
 }
 
 (async () => {
+  /**
+   * Transform `.ts` files out of the box by invoking
+   * the `ts-node` via `transpile-only` mode that compiles
+   * TS files without doing any extensive type checks.
+   *
+   * We must register `ts-node` _before_ loading inline
+   * scripts too because otherwise we will not be able to
+   * require `.ts` configuration files.
+   */
+  /* eslint-disable-next-line @typescript-eslint/no-var-requires */
+  require('ts-node').register({
+    transpileOnly: true,
+    compilerOptions: {
+      esModuleInterop: true,
+      allowJs: true,
+      target: 'es2018',
+    },
+  });
+
   if (options.inline) {
     const source = await readStdin();
     loadInlineScript(source);
@@ -141,20 +160,6 @@ async function prepareSuites(inputs: string[]) {
         throw new Error(`cannot find module '${name}'`);
       }
     }
-    /**
-     * Transform `.ts` files out of the box by invoking
-     * the `ts-node` via `transpile-only` mode that compiles
-     * TS files without doing any extensive type checks
-     */
-    /* eslint-disable-next-line @typescript-eslint/no-var-requires */
-    require('ts-node').register({
-      transpileOnly: true,
-      compilerOptions: {
-        esModuleInterop: true,
-        allowJs: true,
-        target: 'es2018',
-      },
-    });
     /**
      * Handle piped files by reading the STDIN
      * ex: ls example/suites/*.js | npx @elastic/synthetics
@@ -174,7 +179,10 @@ async function prepareSuites(inputs: string[]) {
   /**
    * Validate and handle configs
    */
-  const config = readConfig(environment, options.config);
+  const config =
+    options.config || !options.inline
+      ? readConfig(environment, options.config)
+      : {};
   const params = merge(config.params, options.params || {});
 
   /**
@@ -187,7 +195,7 @@ async function prepareSuites(inputs: string[]) {
       headless: options.headless,
       chromiumSandbox: options.sandbox,
       ignoreHTTPSErrors: options.ignoreHttpsErrors,
-    }
+    },
   ]);
 
   const results = await run({
--- a/__tests__/cli.test.ts
+++ b/__tests__/cli.test.ts
@@ -40,19 +40,73 @@ describe('CLI', () => {
 
   const FIXTURES_DIR = join(__dirname, 'fixtures');
 
-  it('runs inline tests', async () => {
-    const cli = new CLIMock()
-      .stdin(
-        `step('check h2', async () => {
-        await page.goto(params.url, { timeout: 1500 });
-        const sel = await page.waitForSelector('h2.synthetics', { timeout: 1500 });
-        expect(await sel.textContent()).toBe("Synthetics test page");
-      })`
-      )
-      .args(['--inline', '--params', JSON.stringify(serverParams)])
-      .run();
-    await cli.waitFor('Journey: inline');
-    expect(await cli.exitCode).toBe(0);
+  describe('for inline tests', () => {
+    it('runs inline tests', async () => {
+      const cli = new CLIMock()
+        .stdin(
+          `step('check h2', async () => {
+          await page.goto(params.url, { timeout: 1500 });
+          const sel = await page.waitForSelector('h2.synthetics', { timeout: 1500 });
+          expect(await sel.textContent()).toBe("Synthetics test page");
+        })`
+        )
+        .args(['--inline', '--params', JSON.stringify(serverParams)])
+        .run();
+      await cli.waitFor('Journey: inline');
+      expect(await cli.exitCode).toBe(0);
+    });
+
+    it('does not load a configuration file without a config param', async () => {
+      // jest by default sets NODE_ENV to `test`
+      const original = process.env['NODE_ENV'];
+      const output = async () => {
+        const cli = new CLIMock()
+          .stdin(`step('fake step', async () => {})`)
+          .args(['--reporter', 'json', '--inline'])
+          .run({ cwd: FIXTURES_DIR });
+        await cli.waitFor('journey/start');
+        expect(await cli.exitCode).toBe(0);
+        return cli.output();
+      };
+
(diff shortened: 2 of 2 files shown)


src/cli.ts
```diff
@@ -174,7 +179,10 @@ async function prepareSuites(inputs: string[]) {
   /**
    * Validate and handle configs
    */
-  const config = readConfig(environment, options.config);
+  const config =
+    options.config || !options.inline
```
vigneshshanmugam, line 183: This is changing the behavior of our current runner, by default synthetics looks for config files recursively till root directory https://github.com/lucasfcosta/synthetics/blob/f9d4eba2cc337175b6e97137b1d5cd7349b4f863/src/config.ts#L45-L47

src/cli.ts
```diff
@@ -187,7 +195,7 @@ async function prepareSuites(inputs: string[]) {
       headless: options.headless,
       chromiumSandbox: options.sandbox,
       ignoreHTTPSErrors: options.ignoreHttpsErrors,
-    }
+    },
```
vigneshshanmugam, line 198: DId we enable trailing comma in prettier?

src/cli.ts
lucasfcosta: Hey @vigneshshanmugam, indeed, this does change the default current behaviour of the runner. That's because when running inline we would use the `cwd` to find the configuration.

That can cause a few problems which, in my view, are very significant:
* IMO this design doesn't comply with the "least surprising behaviour" principle.
    Configuration would be read from `cwd`, not the files directory, which could be confusing for users as they could simply run synthetics from another directory and thus have totally different configurations applied.
* When running tests inline, you're piping a journey's contents into the executable for testing. Therefore, the options you'd usually use for suites/directories wouldn't apply most of the time (you can still pass a configuration file path though). This is the same behaviour as Jest, and thus is more compliant with the JS testing environment "standards" (or at least their usual behaviour).

Btw that's as per the AC in https://github.com/elastic/synthetics/issues/407.

Do you think we should change it? I'm sorry but I didn't understand exactly whether you suggested a change here, and, if so, what that change is.

src/cli.ts
lucasfcosta: It seems to be the default (`es5`) in prettier 🤔 

https://prettier.io/docs/en/options.html#trailing-commas

src/cli.ts
vigneshshanmugam: Apologies for the lack of detail here. I agree with you on the inline journeys here, we dont need to use config unless specified. 

I was going for this. Assume this is the folder structure
```
 - example/todos/
    - a.journeys.ts
    - b.journeys.ts
    - synthetics.config.ts
    - package.json
```

Previous to this PR, when the runner is invoked within `example/todos`, it picks up the config from `synthetics.config.ts`.

With the current PR, the runner does not run with the config. Is this intended change?

src/cli.ts
vigneshshanmugam: Hmmm, not sure how its picking it now.

src/cli.ts
lucasfcosta: Thanks for taking the time to explain 💖 

In the case you mentioned, the configuration _will_ be picked up, though. The configuration will only be ignored if you're using the inline flag, meaning you're piping a script into the runner's `stdin`.

This line:

```js
const config = options.config || !options.inline
      ? readConfig(environment, options.config)
      : {};
```

Will ensure that if you're _not_ running an inline script, `config` will be truthy and thus it will be read from the passed directory.

I've just tested this exact case by doing the following:

1. I created a new directory with two files: a journey, and a configuration file. The configuration file in that folder will write to the console when it's loaded. For the sake of this example, let's imagine these files' folder is `/synth-example`.
    ```js
    // example.journey.ts
    const { journey, step } = require('@elastic/synthetics');

    journey('my journey', ({ page }) => {
      step('google', async () => {
        await page.goto('https://www.google.com');
      });
    });
    ```
    
    ```js
    // synthetics.config.js
    module.exports = (() => {
      console.log('CONFIG LOADED');
      return {};
    })();
    ```
2. I checked out this branch and built the runner with `npm run build`
3. I then `cd`ed into `/synthe-example` and invoked the built runner using the built `cli` for the package `/path_to_synthetics_package/synthetics/dist/cli.js .`. As you can see below, the config was loaded.
    ```
    $ ../dist/cli.js .
    CONFIG LOADED

    Journey: my journey
       ✓  Step: 'google' succeeded (726 ms)

     1 passed (2373 ms)
    ```
4. When piping the contents of `example.journey.js` into the built `cli.js`, I don't see the `CONFIG LOADED` string written to `stdout`, as configuration wasn't loaded.
    > Please notice that you need to convert your journey to an inline journey to run it inline (that means using only the globally provided `step`).
    ```js
    // example-inline.journey.js
    step('google', async () => {
      await page.goto('https://www.google.com');
    });
    ```
    Results:
    ```
    ➜ cat example-inline.journey.js | ../dist/cli.js --inline
    Journey: inline
       ✓  Step: 'google' succeeded (1155 ms)
     1 passed (2541 ms)
    ```


@vigneshshanmugam if you're happy with the explanation above just lmk and I can go ahead and merge. If you think we should change this behaviour please let me know, but I do want to reassure you that in the case you mentioned, the configuration will be read from the passed directory. It's only _not_ read when using the `--inline` flag.

vigneshshanmugam: My bad, I somehow thought we have changed the behaviour of config loading for suites. Thanks for the detailed explanation and apologies for not looking through the code correctly. PR looks good.

vigneshshanmugam (approved):
LGTM

PR #436, opened 2022-01-04
Update dependencies

lucasfcosta:
# Summary

This PR updates the following dependencies:

* `jest` and `ts-jest` (for downstream `json-schema`)
* `sharp` for downstream `ansi-regex`.

These were required by our dependency related automations.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5,35 +5,34 @@
   "requires": true,
   "dependencies": {
     "@babel/code-frame": {
-      "version": "7.12.11",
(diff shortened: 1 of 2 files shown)


vigneshshanmugam (approved):

justinkambic (approved):
I've noticed that when running `npm install` we end up with a dirty git status. This is happening on `main` as well, so it's not introduced by this patch. We should add an additional step to CI to check for this so we can avoid having unstaged changes when doing a fresh install.

PR #437, opened 2022-01-06
chore: fix flaky tests in CLI mock

### LINKED ISSUE #430: Address synthetic flaky tests
# Context

It has been detected several flaky tests and the goal of this issue is to make them more stable. 

The tests encountered so far are:

- **network › timings for aborted requests**. [Pipeline](https://apm-ci.elastic.co/blue/organizations/jenkins/apm-agent-rum%2Felastic-synthetics%2FPR-429/detail/PR-429/1/pipeline/309)
- **CLI › TLS site with self-signed cert › succeeds succeeds with --ignore-https-errors**. [Pipeline](https://apm-ci.elastic.co/blue/organizations/jenkins/apm-agent-rum%2Felastic-synthetics%2FPR-428/detail/PR-428/1/pipeline/)
- **CLI › screenshots with device emulation**. [Pipeline](https://apm-ci.elastic.co/blue/organizations/jenkins/apm-agent-rum%2Felastic-synthetics%2FPR-428/detail/PR-428/2/pipeline/)

# Findings

@emilioalvap have done some findings:

"some of the unit tests for synthetics are relying on the cli pipe to emit message individually but under certains circumstances two or more can be passed in the same event, making the out not-Json-parseable"

vigneshshanmugam:
+ fix #430 

Synthetics runner uses streams in object mode to write the data to the corresponding File descriptor, in our case it's process.stdout. So the runner writes the data on each new line for every event type - `journey/start`, `step/end` and so on.

The data written to the underlying stream will be flushed once their internal buffer length exceeds, We can control this [buffering](https://nodejs.org/api/stream.html#buffering) behavior of streams by using `highWaterMark` option. In the tests case, some of the data is interleaved as screenshot blob data events chunk exceeds that limit and written as interleaved chunks to the file descriptor. 

Our current code does not take this in to account and instead pushes the broken data chunk to the final buffer. The solution was simply to move the splitting logic after merging all the chunks which would take care of interleaved chunks.

### DIFF
--- a/__tests__/cli.test.ts
+++ b/__tests__/cli.test.ts
@@ -187,6 +187,7 @@ describe('CLI', () => {
       ])
       .run();
     await cli.waitFor('journey/end');
+    expect(await cli.exitCode).toBe(0);
 
     const data = safeParse(cli.buffer());
     const screenshotRef = data.find(
@@ -201,9 +202,7 @@ describe('CLI', () => {
 
     const traceData = data.find(({ type }) => type === 'step/metrics');
     expect(traceData).toBeDefined();
-
-    expect(await cli.exitCode).toBe(0);
-  }, 30000);
+  });
 
   it('override screenshots with `--rich-events` flag', async () => {
     const cli = new CLIMock()
@@ -215,10 +214,9 @@ describe('CLI', () => {
       ])
       .run();
     await cli.waitFor('journey/end');
-    const screenshots = cli
-      .buffer()
-      .map(data => JSON.parse(data))
-      .find(({ type }) => type === 'step/screenshot_ref');
+    const screenshots = safeParse(cli.buffer()).find(
+      ({ type }) => type === 'step/screenshot_ref'
+    );
     expect(screenshots).not.toBeDefined();
     expect(await cli.exitCode).toBe(0);
   });
--- a/__tests__/cli.test.ts
+++ b/__tests__/cli.test.ts
@@ -559,7 +559,7 @@ class CLIMock {
       this.data = data.toString();
       // Uncomment the line below if the process is blocked and you need to see its output
       // console.log('CLIMock.stdout:', this.data);
-      this.chunks.push(...this.data.split('\n').filter(Boolean));
+      this.chunks.push(this.data);
       if (this.waitForPromise && this.data.includes(this.waitForText)) {
         this.process.stdout.off('data', dataListener);
         this.waitForPromise();
@@ -588,6 +588,9 @@ class CLIMock {
   }
 
   buffer() {
-    return this.chunks;
+    // Merge all the interleaved chunks from stdout and
+    // split them on new line as synthetics runner writes the
+    // JSON output in separate lines for every event.
+    return this.chunks.join('').split('\n').filter(Boolean);
   }
 }
--- a/__tests__/cli.test.ts
+++ b/__tests__/cli.test.ts
@@ -364,6 +364,8 @@ describe('CLI', () => {
         JSON.stringify({ url: tlsServer.TEST_PAGE }),
         '--reporter',
         'json',
+        '--screenshots',
+        'off',
       ];
     });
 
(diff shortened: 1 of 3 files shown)


devcorpio:
LGTM, I have run the locally multiple times and they have worked successfully 🎉

devcorpio (approved):

emilioalvap:
I am personally seeing some flakiness still while running tests on WSL2-Ubuntu20.04:

```
 ● CLI › TLS site with self-signed cert › succeeds succeeds with --ignore-https-errors

    SyntaxError: Unexpected token { in JSON at position 12173
        at JSON.parse (<anonymous>)

      317 |       expect(await cli.exitCode).toBe(0);
    > 318 |       expect(JSON.parse(cli.output()).journey).toEqual(
          |                   ^
      320 |         expect.objectContaining({ status: 'succeeded' })
      321 |       );
      322 |     });
```

Steps I've taken to reproduce:
- Checkout PR
- Ran `npm run clean && npm run build`
- Ran tests with `npm run test`

Did some debugging and it seems `cli.output()` is still mapped to multiple JSON objects:

vigneshshanmugam:
@emilioalvap Interesting, Thanks for testing and confirming the bug still exists. Could you log the output and send it over as it would make it easier for me to add a fix? Thanks

emilioalvap:
Sure, I've modified the following tests to log the relevant cli output in case of error:
- `CLI › TLS site with self-signed cert › fails by default`
- `CLI › TLS site with self-signed cert › succeeds succeeds with --ignore-https-errors`

With the following:
```js
try {
     expect(JSON.parse(cli.output()).journey).toEqual(
          expect.objectContaining({ status: 'failed' })
     );
} catch(e) {
     console.log(cli.output())
     throw e;
}
```
I've attached the relevant [console.log](https://github.com/elastic/synthetics/files/7845528/console.log). It took a few attemps today to get the error and I've noticed  that other tests are also resolving flakily. I'm attaching [npm_test.log](https://github.com/elastic/synthetics/files/7845541/npm_test.log) for a run which contains all the flaky errors I've been able to notice so far.

vigneshshanmugam:
@emilioalvap Could you have a look again at this PR. Thanks

emilioalvap:
@vigneshshanmugam It has definitively solved some the issues, so I think this could be a LGTM.

I'm still getting a flaky test result on this test:
````js
 FAIL  __tests__/core/runner.test.ts (17.531 s)
  ● runner › run - timestamps must be in order

    expect(received).toEqual(expected) // deep equality

    - Expected  - 1
    + Received  + 0

      Array [
        "synthetics/metadata",
        "journey/start",
        "journey/network_info",
        "step/screenshot",
    -   "step/filmstrips",
        "step/end",
        "journey/end",
      ]

      689 |     ];
      690 |     const collectOrder = events.map(event => event.type);
    > 691 |     expect(collectOrder).toEqual(realEventsOrder);
          |                          ^
      692 |   });
      693 |
      694 |   /**

      at Object.<anonymous> (__tests__/core/runner.test.ts:691:26)
````

I've debugged around a bit, and it seems sometimes `filmstrips` array is populated and sometimes it isn't:

````js
# Error run

console.log
      >>>> DEBUG4 >>>>> { filmstrips: [] }

      at PluginManager.stop (src/plugins/plugin-manager.ts:86:15)

# Success
console.log
      >>>> DEBUG >>>>> {
        filmstrips: [
          {
            blob: '/9j/4AAQSkZJRgABA...',
            mime: 'image/jpeg',
            start: [Object]
          }
        ]
      }

      at PluginManager.stop (src/plugins/plugin-manager.ts:86:15)
````

Since this error doesn't seem to happen in CI and it only seems to happen inside WSL, I'd say we can probably merge this one and address it in another issue. What do you think?

vigneshshanmugam:
@emilioalvap Thanks for checking it, yeah let's address that in separate PR. The issue could be that there is not enough time for the tracing to kick in.

PR #447, opened 2022-01-28
fix: ensure journey/end is written to fd

### LINKED ISSUE #423: Debug logs should not go to special FD (3)
It is currently the case that if debugging is enabled, synthetics will log to the `--out-fd` value passed by heartbeat . This FD is reserved purely for JSON formatted messages, and creates ugly heartbeat logs like the following:

```
2021-12-15T17:55:22.123Z    WARN    synthexec/synthexec.go:226  error parsing line: invalid character 'a' looking for beginning of value for line: at 1822 ms register step: my-step-name
```

Heartbeat will properly pick up output to stderr/stdout and forward to ES as `synthetics.type: console/log` documents that the UI knows how to display.

We probably just need to change [this line](https://github.com/elastic/synthetics/blob/5ddb5c281276ef70084269f767a6002c6040dafa/src/index.ts#L42) to point elsewhere

### LINKED ISSUE #446: reporter stream buffer is not emptied in all cases
The issue was brought up when users were not seeing `heartbeat/summary` documents for some of the pages that has more than 200 network requests. 

Example test is
```js
step('load homepage', async () => {
});
```

When the agent is invoked from HB, the agent writes all the events in JSON format and sends `journey/end` event which is responsible for creating `summary` event in Heartbeat. 

But in some cases we were not seeing the `journey/end` event and the node process is completed before draining all the items in the stream buffer.

vigneshshanmugam:
+ fix #446 
+ The stream buffer was not emptied as the `minLength` for flushing was set to `0` by default which made the flush to return immediately without draining its contents - https://github.com/pinojs/sonic-boom/blob/21d711575e4fe4c2fbd50e3f0cbf8f52231fdbaa/index.js#L255
+ Now, we set the minLength to `1` to make sure we don't miss even the last byte when the synthetics runner ends. 

+ fixes #423 
+ Now debug logs are written to stdout and heartbeat picks them and writes them as part of `stdout` payload message.


__tests__/core/runner.test.ts
```diff
@@ -690,4 +690,21 @@ describe('runner', () => {
     const collectOrder = events.map(event => event.type);
     expect(collectOrder).toEqual(realEventsOrder);
   });
+
+  it('run - ensure journey/end is done for real world pages', async () => {
+    const j1 = journey('journey1', async ({ page }) => {
+      step('load homepage', async () => {
+        await page.goto('https://www.elastic.co');
```
vigneshshanmugam, line 697: Its really hard to test this behaviour which is why I added a real world page instead of adding a custom page which contains multiple resources.

If you remove the `minLength: 0`, this test breaks which ensures the fix is working correctly.

vigneshshanmugam:
E2E test failure is due to this current bug we have as in Heartbeat summary docs is not generated and the current 8.x snapshot image contains the previous version of Synthetics which stops the Uptime journey files from running and hanging forever. https://github.com/elastic/synthetics/blob/201f7da7db8d8ce9c13855023c8566b73ebaf91a/__tests__/e2e/uptime.journey.ts#L67-L93

__tests__/core/runner.test.ts
andrewvc: Can you add this as a comment. My $0.02 is I always try to leave a comment in the code vs. the PR except in the case of comments explaining deleted stuff.

vigneshshanmugam: Sure will do.

andrewvc (approved):
LGTM

dominiqueclarke (approved):
LGTM!

PR #450, opened 2022-02-04
fix: capture ReferenceError inside journeys

vigneshshanmugam:
+ For inline journeys, the ReferenceErrors inside script was not captured as part of the journey error context. This PR fixes that behaviour and also adds tests. 
+ The issue was due to the fact that our inline journey function was `async`, which is wrong as the JourneyCallback is expected to register steps synchronously.


lucasfcosta (approved):
LGTM, good catch on this one!

I have re-triggered CI as it seems like it didn't have any failures, it was just aborted.

PR #454, opened 2022-02-07
Export `Action` type

justinkambic:
Exports the `Action` type for downstream use alongside `ActionInContext`.


src/formatter/javascript.ts
```diff
@@ -39,7 +39,7 @@ export type ActionInContext = {
   title?: string;
 };
 
-type Action = {
+export type Action = {
```
vigneshshanmugam, line 42: might need Signal as well?

justinkambic: Done. https://github.com/elastic/synthetics/pull/454/commits/72d044442418137e876687f87628eb9f45bcd609

vigneshshanmugam (approved):

PR #457, opened 2022-02-11
feat: Make JavaScript generator auto step insertion optional

justinkambic:
Updates the JS code generator to allow for a flag that will enable calling code to generate a single step with all the passed actions contained in it.

Also added a code to ignore the check for being in a current step.

Introduced a new function that will generate code directly from a set of steps.

Moved step types from the recorder to here to avoid circular references.

~Code is WIP.~


__tests__/formatter/javascript.test.ts
```diff
@@ -25,161 +25,230 @@
...
   });
 
-  it('suite journeys', async () => {
+  it('suite journeys', () => {
     const formatter = new SyntheticsGenerator(true);
-    expect(formatter.generateText(recorderActions)).toMatchSnapshot();
+    expect(formatter.generateText(recorderStep.actions)).toMatchSnapshot();
   });
 
   it('use modified title if available', async () => {
```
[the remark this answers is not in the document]
justinkambic: https://github.com/elastic/synthetics/pull/457/commits/3066b7cd9c1799030e67e47ca25534204220a735

src/formatter/javascript.ts
```diff
@@ -73,7 +80,16 @@ export class SyntheticsGenerator extends JavaScriptLanguageGenerator {
     this.previousContext = undefined;
   }
 
-  generateAction(actionInContext: ActionInContext) {
+  generateAction(
```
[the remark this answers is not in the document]
justinkambic: https://github.com/elastic/synthetics/pull/457/commits/87a9e848ec89c6d5478b521271efd8031575b8b0

__tests__/formatter/javascript.test.ts
```diff
@@ -25,161 +25,230 @@
...
-          isAsync: true,
-        },
-      ],
-      button: 'left',
-      modifiers: 0,
-      clickCount: 1,
+    {
+      pageAlias: 'page1',
+      isMainFrame: true,
+      frameUrl: 'https://github.com/babel/minify',
```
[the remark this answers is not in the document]
justinkambic: https://github.com/elastic/synthetics/pull/457/commits/3066b7cd9c1799030e67e47ca25534204220a735

src/formatter/javascript.ts
```diff
@@ -73,7 +80,25 @@ export class SyntheticsGenerator extends JavaScriptLanguageGenerator {
...
+   * @param actionInContext The action to create code for.
+   * @param ignoreStepDefaults When true, the function will not create steps between
+   * navigation events as it normally would.
+   * @param customStepsOverride When true, the function will not throw an error if the
+   * called while the generator is already recording a step.
+   * @returns the strings generated for the action.
+   */
+  generateAction(
+    actionInContext: ActionInContext,
+    ignoreStepDefaults = false,
```
kyungeunni, line 94: Can we be more specific on the parameter name, for example, `newStepOnNavigate` or something so users don't need to remember the default behaviour? (I appreciate you've provided a great description, but I still prefer if I didn't need to read them :D)

src/formatter/javascript.ts
```diff
@@ -73,7 +80,25 @@ export class SyntheticsGenerator extends JavaScriptLanguageGenerator {
...
+    customStepsOverride = false
```
kyungeunni, line 95: I couldn't understand what this parameter does and why we need this one - could you tell me more about it please?

src/formatter/javascript.ts
```diff
@@ -211,18 +242,76 @@ export class SyntheticsGenerator extends JavaScriptLanguageGenerator {
...
     if (this.isSuite) {
       text.push(this.generateFooter());
     }
+    return text.filter(s => !!s || !filterEmptyLines);
+  }
+
+  /**
+   * Generates JavaScript code for the given list of actions.
+   * @param actions The actions to convert to code.
+   * @param overrideStepOrder When `true`, all passed actions will be nested in one step.
```
kyungeunni, line 284: Similar concerns to the comment I left for `ignoreStepDefaults`. Do you think something like `wrapIntoStep` could work?

lucasfcosta (approved):
LGTM. Also tested this PR with the recorder as per the instructions on the recorder PR.

PS: Great comments by @kyungeunni on naming.

src/formatter/javascript.ts
```diff
@@ -39,7 +46,7 @@ export type ActionInContext = {
   title?: string;
 };
 
-type Action = {
+export type Action = {
   name: string;
   selector?: string;
```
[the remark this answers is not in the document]
justinkambic: That change is already approved, so let's rebase/merge it and then incorporate it into these changes.

src/formatter/javascript.ts
justinkambic: I'm happy to accept suggestions for a more appropriate name here. Perhaps we can call it `ignoreNewStepLogic`?

__tests__/formatter/__snapshots__/javascript.test.ts.snap
```diff
@@ -1,51 +1,41 @@
...
-  ]);
-  await page1.close();
-});"
-`;
-
-exports[`Synthetics JavaScript formatter suite journeys 1`] = `
-"const { journey, step, expect } = require('@elastic/synthetics');
+exports[`Synthetics JavaScript formatter accepts custom step organization 1`] = `
+Array [
+  "const { journey, step, expect } = require('@elastic/synthetics');
```
vigneshshanmugam, line 5: the formatting if not correct.

__tests__/formatter/javascript.test.ts
```diff
@@ -24,162 +24,193 @@
...
+    expect(
+      generator
+        .generateFromSteps(
+          testSteps.map((s: Step) => {
+            s.name = 'test-name';
+            return s;
+          }),
+          false
+        )
+        .filter(line => line.length === 0)
```
vigneshshanmugam, line 205: We can move the handling to the step generation code.

vigneshshanmugam (changes requested):
There are few problems I see with the code that are worth addressing, 
- The formatting and overrides seems all over the place - We need to do only at 1 place which should be based on journey is a `Suite` or not. 
- Instead of these flags `newStepsOnNavigate` and `customStepsOverride` which are overriden anyway by the underlying step generator. It would be better to move towards a logic where we can provide a callback from `Recorder` on what it considers a `isNewStep`. So basically you would pass a `isNewStep` callback that does not generate a step on navigation and controlled by the boolean function from Recorder.

__tests__/formatter/__snapshots__/javascript.test.ts.snap
justinkambic: Overwrote this in https://github.com/elastic/synthetics/pull/457/commits/069aaba4c69b93887ba4b83f18a438031926a709.

It's now a snapshot of a single string that looks ok.

justinkambic:
After some offline conferring we decided it's the best course of action to allow the Recorder to simply specify declarative objects that contain the data needed for Synthetics to generate the step code. This means we're stripping all step insertion logic out of the `generateAction` function, and leaving that up to the `generateFromSteps` function added in this PR.

justinkambic:
@vigneshshanmugam I think we are good for a re-review now.

src/formatter/javascript.ts
```diff
@@ -213,19 +219,41 @@ export class SyntheticsGenerator extends JavaScriptLanguageGenerator {
...
+        if (actionText.length) text.push(actionText);
+      }
+
+      text.push(this.generateStepEnd());
     }
     if (this.isSuite) {
       text.push(this.generateFooter());
     }
-    return text.filter(t => Boolean(t)).join('\n');
+    return text.filter(s => !!s || !filterEmptyLines).join('\n');
```
vigneshshanmugam, line 256: does filterEmptylines still required? Asking bcoz you are already checking for actionText.length before which should eliminate empty actions.

vigneshshanmugam (approved):
LGTM, Thanks @justinkambic

src/formatter/javascript.ts
justinkambic: d5638cbd91017ef0667e394d3bd9c3d0a041f6a4

PR #462, opened 2022-02-22
fix: allow UA emulation when overriding

vigneshshanmugam:
+ When users are overriding the useragent with  `--playwright-options` the UA was being overwritten in the agent code. This PR fixes that by using the UA passed to the agent.

### DIFF
--- a/src/core/gatherer.ts
+++ b/src/core/gatherer.ts
@@ -65,7 +65,7 @@ export class Gatherer {
     return { browser: Gatherer.browser, context, page, client };
   }
 
-  static async getUserAgent(userAgent: string) {
+  static async getUserAgent(userAgent?: string) {
     const syntheticsIdentifier = ' Elastic/Synthetics';
     if (!userAgent) {
       const session = await Gatherer.browser.newBrowserCDPSession();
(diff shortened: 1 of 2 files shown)


src/core/gatherer.ts
```diff
@@ -65,10 +65,13 @@ export class Gatherer {
     return { browser: Gatherer.browser, context, page, client };
   }
 
-  static async getUserAgent() {
-    const session = await Gatherer.browser.newBrowserCDPSession();
-    const { userAgent } = await session.send('Browser.getVersion');
-    return userAgent + ' Elastic/Synthetics';
+  static async getUserAgent(userAgent: string) {
```
dominiqueclarke, line 68:
```suggestion
  static async getUserAgent(userAgent?: string) {
```

src/core/gatherer.ts
```diff
@@ -65,10 +65,13 @@ export class Gatherer {
...
+    const syntheticsIdentifier = ' Elastic/Synthetics';
+    if (!userAgent) {
+      const session = await Gatherer.browser.newBrowserCDPSession();
+      ({ userAgent } = await session.send('Browser.getVersion'));
```
dominiqueclarke, line 72: TIL

src/core/gatherer.ts
vigneshshanmugam: Good catch. Fixed.

dominiqueclarke (approved):
LGTM. I can see the correct user agent when using playwright options, appended with `Elastic Synthetics`.

Screen Shot 2022-02-22 at 4 56 18 PM

PR #465, opened 2022-03-05
fix: empty the write buffer before exit

### LINKED ISSUE #446: reporter stream buffer is not emptied in all cases

vigneshshanmugam:
+ fix #446 
+ The max buffer capacity for SonicBoom is `64 * 1024` bytes and the API `flush` and `flushSync` behaves differently when it comes to writing the last data from the streams buffer.

+ `flush` API only writes the current element in the buffer which is always the first item and drops other data and closes the FD.
+ However `flushSync` ensures all the data from the buffer is completely written.

#### Example
```
// Previous - FLUSH Async
SONICBOOM: Data-0-65189
SONICBOOM: Data-1-29862
SONICBOOM: flush-95051
SONICBOOM: Writing-65189
-- Data 1 is completetly lost
- Runner closed

// Current - FlushSync
SONICBOOM: Data-0-61310
SONICBOOM: Data-1-53290
SONICBOOM: flushSync-114600
SONICBOOM: Writing-61310
SONICBOOM: Writing-53290
- Runner closed
```

- [ ] TODO - Add tests

### DIFF
--- a/src/reporters/base.ts
+++ b/src/reporters/base.ts
@@ -70,8 +70,13 @@ export default class BaseReporter {
      * before destroying the pipe with underlying file descriptor
      */
     this.stream = new SonicBoom({ fd: this.fd, sync: true, minLength: 1 });
-    this._registerListeners();
+
+    // flushSync is used here to make sure all the data from the underlying
+    // SonicBoom stream buffer is completely written to the fd before closing
+    // the process
     this.runner.on('end', () => this.stream.flush());
+
+    this._registerListeners();
   }
 
   _registerListeners() {
--- a/src/reporters/json.ts
+++ b/src/reporters/json.ts
@@ -343,7 +343,7 @@ export async function gatherScreenshots(
 }
 
 export default class JSONReporter extends BaseReporter {
-  _registerListeners() {
+  override _registerListeners() {
     /**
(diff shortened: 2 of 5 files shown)


__tests__/core/runner.test.ts
```diff
@@ -692,13 +692,10 @@ describe('runner', () => {
...
 
   /**
-   * TODO: Move this as part of integration test
    * Its really hard to ensure the journey/end is called for a real world page
    * without actually testing on a real world webpage.
-   * 
-   * FLAKY skipped
    */
-  it.skip('run - ensure journey/end is written for real world pages', async () => {
+  it('run - ensure journey/end is written for real world pages', async () => {
```
vigneshshanmugam, line 698: We cant get a better indicator than this test, I am keeping this as it is. 

I did not want to test the sonic-boom package functionality in our agent.

PR #467, opened 2022-03-09
fix: keep only necessary fields for network info

### LINKED ISSUE #466: remove unused fields from network information
+ We need a allowlist of fields that can be set as part of the `journey/network_info` event rather than creating the JSON with all the available fields from CDP. This would also avoid the max buffer limit that gets triggered and causes issues like #446 

+ Reason being it leads of explosion of data due to the field Subject Alternative Name list. A single request from [recaptcha.net](http://recaptcha.net/) has a SAN list of 500KiB of data

+ We need this as an interim solution until we get this PR back #372 and move to newer versions of PW.

vigneshshanmugam:
+ fix #466 
+ We use only the relevant fields that are shown on the UI and drop all the other fields that increases the size of the network document. example- `san_list` from certificates. 
+ Benchmarks with before/after this PR

```
Website - https://www.elastic.co
Before - 1795494 bytes
After - 1207272 bytes


Website - https://www.cnn.com/
Before - 2492815 bytes
After - 1895130 bytes

Website - recaptcha which has huge SAN list
Before - 1101419 bytes
After - 440826 bytes
```

Yields a savings of 60kB approx. for every synthetics test run.


src/plugins/network.ts
```diff
@@ -117,15 +118,25 @@ export class NetworkManager {
...
-      requestSentTime: timestamp,
+      status: -1,
+      request: {
+        url,
+        method,
+        headers,
+        bytes: postData?.length || 0,
+      },
+      response: {
+        statusCode: -1,
```
andrewvc, line 131: Why do we set this to `-1` vs `null`

src/plugins/network.ts
```diff
@@ -117,15 +118,25 @@ export class NetworkManager {
       step: this._currentStep,
       timestamp: getTimestamp(),
       url,
-      request,
       type,
       method,
-      requestSentTime: timestamp,
+      status: -1,
```
andrewvc, line 123: We previously set this to `0`, also, why not `null`? It'd be great to have a comment somewhere nearby explaining these numeric values.

src/plugins/network.ts
```diff
@@ -194,6 +226,9 @@ export class NetworkManager {
     }
     if (encodedDataLength >= 0) {
       record.transferSize = encodedDataLength;
```
andrewvc, line 228: Do we need both values? I'm wondering if `response.body` is sufficient

andrewvc (commented):
Left some questions, mostly looking good! I still need to test against a kibana instance (unless someone else gets to it first)

andrewvc:
Completed manual testing against a simple site, touring the Uptime browser UI nothing seems to be missing.

__tests__/fixtures/networkinfo.ts
```diff
@@ -25,184 +25,130 @@
...
   {
-    browser: { name: 'HeadlessChrome', version: '90.0.4392.0' },
-    step: {
-      name: 'go to app',
-      index: 1,
-    },
-    timestamp: 1612482095137858,
-    url: 'https://vigneshh.in/',
+    browser: { name: 'HeadlessChrome', version: '94.0.4595.0' },
+    step: null,
```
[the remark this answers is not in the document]
vigneshshanmugam: We do, I just removed from the fixture. Can add it back if you feel it's good to have.

__tests__/fixtures/networkinfo.ts
```diff
@@ -25,184 +25,130 @@
...
+    timestamp: 1646782451347652.5,
+    url: 'http://localhost:56899/index',
+    type: 'Document',
+    method: 'GET',
+    status: 200,
     request: {
-      url: 'https://vigneshh.in/',
+      url: 'http://localhost:56899/index',
```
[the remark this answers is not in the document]
vigneshshanmugam: It was from this test `network.test.ts` - https://github.com/elastic/synthetics/blob/41330fd464afacb8cfb129c63271ebc9a6370d6d/__tests__/plugins/network.test.ts#L117

src/plugins/network.ts
vigneshshanmugam: We use `transferSize` in the UI. the `response.body` is just for the ECS spec.

src/plugins/network.ts
vigneshshanmugam: Hmmm good point, status: 0 denotes a successful response when the browser was having CORS issue so to avoid mistaking it for the aborted requests, I have set it to `-1` which feels right. 

`null` would denote a separate type. Keeping `number` type seems better for status codes.

src/plugins/network.ts
vigneshshanmugam: explained below. https://github.com/elastic/synthetics/pull/467#discussion_r823293893

src/plugins/network.ts
```diff
@@ -165,12 +177,32 @@ export class NetworkManager {
...
-      responseReceivedTime: timestamp,
-      transferSize: response.encodedDataLength,
-    });
+
+    record.status = response.status;
+    record.responseReceivedTime = timestamp;
+    record.transferSize = response.encodedDataLength;
+    record.response = {
+      url: response.url,
+      statusCode: response.status,
```
dominiqueclarke, line 186: The Uptime UI expects this to be `status`, not `statusCode`. Status is missing from waterfall chart in this version of the synthetics agent.

dominiqueclarke (changes requested):

src/plugins/network.ts
vigneshshanmugam: I thought we used the status from the top level not inside the `response`. You are right - https://github.com/elastic/kibana/blob/affbed7a161ae5b41d4dde8264b597b7c21f453e/x-pack/plugins/uptime/server/lib/requests/get_network_events.ts#L70

I will remove the top level fields that are unused in this case.

Issue #470, opened 2022-03-14
New `push` command to sync with monitor management


andrewvc:
# Summary & ACs

1. There should be a `push` command in the `synthetics` agent which sends requests to Kibana to save a new monitor for each journey it found.
2. It should be possible to filter journeys to push in the same way we filter journeys to run
3. It should be possible to connect to the user's stack using CLI options, `config.ts` options, or environment variables (in this order of precedence).

> ⚠️ This issue covers exclusively the `synthetics` package changes: the CLI interface, filtering, bundling, compressing, and sending requests to Kibana.

### The push process

The monitor push process should bundle all the necessary files to run the journey, compress the bundle, and push the `zip` to Kibana using `base64` encoding.

> ⚠️ For now we won't worry with source maps or minification.

> ⚠️ Downloading, decoding and using this zip is covered in a separate Heartbeat issue.

### Connection options

Connection info for the user's stack can be specified via any of the options below, listed in order of precedence.

1. CLI options
2. `config.ts` options
3. Environment variables

> 👉  The new options would use either dashes for the cli version (e.g. `--kibana-space`), object notion for config.ts (e.g. `kibana.space`) and caps / underscores for env vars (e.g. `KIBANA_SPACE`).

The options are enumerated below in dot notation:

1. `cloud.id` `cloud.auth`  OR `kibana.url` / `kibana.username` / `kibana.password` : These let the user specify how to connect to the relevant kibana instance. In practice we only expect the `cloud.*` options to be used, but the `kibana.*` ones are useful for local development. For a description of parsing `cloud.id` for kibana info see [this file](https://github.com/elastic/logstash/blob/main/logstash-core/src/main/java/org/logstash/util/CloudSettingId.java#L28)
2. `kibana.space`: the kibana space to push the monitor to
3. `suite.id` / `suite.name` :  The id and name of this suite. If only the name is specified that is used as the ID

### Additonal fields sent to kibana

This should be able to use the current monitor creation APIs most likely with some minor changes. We would want to serialize the following additional fields for future use if we need to move / migrate data:

- `suite.id`
- `suite.name`
- `journey.name`
- `journey.id`

[...]

PR #475, opened 2022-03-22
chore: restructure to add sub commands to cli

vigneshshanmugam:
+ part of #470 
+ Restructure the CLI and commander program to allow adding subcommands like `push` and other commands in the future. 
+ Introduce loader and normalizer concepts to make the boundary clear on how test suites are loaded and how the options are normalized from the CLI args. **No functionality changes has been introduced**. 
+ Removed `--debug` and `--quiet-exit-mode` from CLI options, debug logs can be enabled by setting env value `DEBUG=synthetics` and quiet mode is default enabled for HB using `--rich-events` flag so its not required. 

### How to test
```
npm run build
cat examples/inline/badssl-self-signed.js | node dist/cli.js --inline
```


src/cli.ts
```diff
@@ -25,200 +23,123 @@
...
-    environment,
-    playwrightOptions,
-    ...options,
+program
+  .command('push')
+  .description(
+    'Push monitors to create new montors with Kibana monitor management UI'
+  )
+  .action(() => {
+    throw new Error('TODO: Implement me');
```
vigneshshanmugam, line 142: Placeholder -> will be implemented in follow up PR's.

src/common_types.ts
```diff
@@ -194,33 +194,51 @@ export type PluginOutput = {
...
+  metrics?: boolean;
+  ssblocks?: boolean;
+  network?: boolean;
+  trace?: boolean;
+  filmstrips?: boolean;
+  environment?: string;
+  playwrightOptions?: PlaywrightOptions;
+  networkConditions?: NetworkConditions;
+  reporter?: CliArgs['reporter'] | Reporter;
 };
```
lucasfcosta, line 242: Nit: I'd probably have a `BaseArgs` type which I'd then only add to with intersection types, so you'd have something like:

```ts
type BaseArgs = {
  params?: Params;
  screenshots?: ScreenshotOptions;
  dryRun?: boolean;
  match?: string;
  tags?: Array<string>;
  outfd?: number;
  wsEndpoint?: string;
  pauseOnError?: boolean;
  ignoreHttpsErrors?: boolean;
  playwrightOptions?: PlaywrightOptions;
}

type CliArgs = BaseArgs & {
  config?: string;
  reporter?: Reporters;
  pattern?: string;
  inline?: boolean;
  require?: Array<string>;
  headless?: boolean;
  sandbox?: boolean;
  richEvents?: boolean;
  capability?: Array<string>;
  throttling?: boolean | string;
}

type RunOptions = BaseArgs & { 
  quietExitCode?: boolean;
  metrics?: boolean;
  ssblocks?: boolean;
  network?: boolean;
  trace?: boolean;
  filmstrips?: boolean;
  environment?: string;
  playwrightOptions?: PlaywrightOptions;
  networkConditions?: NetworkConditions;
  reporter?: CliArgs['reporter'] | Reporter;
}
```

IMO that makes it easier to understand what are the commonalities between the options for different commands.

src/cli.ts
```diff
@@ -1,5 +1,3 @@
-#!/usr/bin/env node
```
lucasfcosta, line 1: Removing this actually broke CI because it doesn't know this is a node executable given you're transforming it into JS and you can't simply run JS from bash (whatever shell).

If you put it back the build should pass :)

lucasfcosta (approved):
LGTM so far! Love these changes. Also had a tip on how to fix CI.

By the way, I tested this locally by re-adding the shebang and it worked fine :)

src/cli.ts
vigneshshanmugam: Oops copy mistake, Will fix it. Thanks for catching it 👍🏽

src/common_types.ts
vigneshshanmugam: Good point ++

src/cli.ts
```diff
@@ -25,200 +23,123 @@
...
+program
+  .command('push')
+  .description(
+    'Push monitors to create new montors with Kibana monitor management UI'
+  )
+  .action(() => {
+    throw new Error('TODO: Implement me');
   });
 
-  if (!options.quietExitCode) {
```
andrewvc, line 211: I'm OK with this being removed it was causing a significant issue in creating this PR, but it is a nice feature.

On some CI systems this would make sense with JUnit XML output since you might want the test command to always succeed, but base the pass/fail nature of the CI task on a subsequent step that parses that output.

vigneshshanmugam: I thought we introduced this feature only for HB, Thanks for the explanation Andrew. I can bring it back.

vigneshshanmugam: Done.

Issue #373, opened 2021-09-09
Popup handling in Synthetic


shahidmgr:
Hi @vigneshshanmugam I want to handle Popup if any comes in any of the steps.

I am not sure when the popup will come in my recorded scripts. 

Suppose I have 5 steps, in any of the steps popup may come or may not. If popup triggers any of the fix steps then it is easy for me to handle.

I want something like

If (popup) 
then handle 
else 
complete my steps

If I am using page.waitForEvent("popup") suppose popup come then its working fine otherwise its simply waiting for event and timeout.

Regards,
Shahid

vigneshshanmugam:
Have  you tried with explicitly with timing down the popup event as you are unsure when they will be opened in new window. You can do `waitForEvent("popup")` with a timeout. 
```
await page.click("button")
await page.waitForEvent("popup", {timeout: 1000});
// waits for 1000 seconds 
// or combine them in single await 
const [popup] = await Promise.all([
  page.waitForEvent('popup', { timeout: 1000})),
  page.click("button");
]);
```

vigneshshanmugam:
> In this Application 6-10 steps we need to scripts but I am not sure when pop up will come. Pop up may come any of the steps or may not be also.

Do you have an example script that you can share, Would like to play around and provide an answer.

shahidmgr:
Hi @vigneshshanmugam I have shared a sample script via email to you.

vigneshshanmugam:
Closing this as we have have agreed to a solution offline.

PR #483, opened 2022-04-01
Upgrade playwright to 1.17.0

andrewvc:
This PR attempts to upgrade playwright to 1.17.0 (and also skips a test that was flakey on the current verison) in order to get ARM support on ubuntu, as required for https://github.com/elastic/beats/issues/29034 . 1.17.0 was, according to the playwright changelogs, the first version to support this combo. Additionally I suspect that our old version of chromium may be part of https://github.com/elastic/beats/issues/30426 

The main issue we have here is that some of the import stuff has changed, especially for the recorder code generation. Playwright has removed the recorder classes we use from the package exports. 

This PR does not yet work, I just tried copying one folder of playwright into here, but that doesn't work. I suspect we'll have to perhaps even fork playwright to make this work, or just stop relying on internal objects.

@vigneshshanmugam @kyungeunni @justinkambic any thoughts on this? Maybe there's a better way to generate code, or at least reduce the coupling here.

It's risky for us to be so far behind playwright core, so anything we can do to make upgrades more streamlined would be great

### DIFF
--- a/__tests__/core/runner.test.ts
+++ b/__tests__/core/runner.test.ts
@@ -656,7 +656,9 @@ describe('runner', () => {
     });
   });
 
-  it('run - timestamps must be in order', async () => {
+  // Skipped for now since filmstrips sometimes don't show up
+  // It appears they aren't part of the trace events sometimes during test runs
+  it.skip('run - timestamps must be in order', async () => {
     const j1 = journey('journey1', async ({ page }) => {
       step('step1', async () => {
         await page.goto(server.TEST_PAGE);
--- a/__tests__/options.test.ts
+++ b/__tests__/options.test.ts
@@ -63,7 +63,7 @@ describe('options', () => {
         ignoreHTTPSErrors: undefined,
         isMobile: true,
         userAgent:
-          'Mozilla/5.0 (Linux; Android 8.0.0; SM-G965U Build/R16NW) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/94.0.4595.0 Mobile Safari/537.36',
+          'Mozilla/5.0 (Linux; Android 8.0.0; SM-G965U Build/R16NW) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/98.0.4695.0 Mobile Safari/537.36',
         viewport: {
           height: 658,
           width: 320,
--- a/examples/todos/package.json
+++ b/examples/todos/package.json
@@ -5,7 +5,7 @@
   "scripts": {},
   "license": "MIT",
   "dependencies": {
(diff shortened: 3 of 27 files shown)


vigneshshanmugam:
Updating PW is going to be tricky this time for us as we have not been keeping up with the versions and we need to do the following steps. I am going to layout the plan that I had

1. Update Synthetics with the latest PW version that would effectively fix these issues #373 and add `playwright-core` to the deps to extend the JavaScriptLanguageGenerator, PW switched to monorepo based architecture hence the need for `playwright-core` and we still need `playwright-chromium` as it would install the chrome browser during install step.

2. Synthetics Formatter has to be changed to output code based out of Locators API - https://playwright.dev/docs/locators instead of selectors to keep in sync with the PW inspector code. 

3. Recorder should still point to older version of synthetics, It should follow these steps
   - Recorder which is based on a PW fork - https://github.com/elastic/playwright has to be updated which would let us register for actions and listen for them in the Recorder
   - Once the upstream changes are rebased and merged, Recorder should be updated to the latest Synthetics which would make it fix the compatibility between Synthetics PW and the Recorder forked PW. 

4. Heartbeat can anytime updated to the latest Synthetics and does not need to wait for the recorder changes.

vigneshshanmugam:
Notes from the meeting.

1. **Synthetics Agent** : Synthetics will be updated to the latest version of Playwright 
    - Decouple the Synthetics formatter and move it to the recorder for the time being.
    - Fastest way to solve some of the issues that users are experiencing with core dumps - https://github.com/elastic/beats/issues/30426
    - Moving to ubuntu based images 

2. **Heartbeat**: We will try to release the synthetics agent with updated PW to be part of 8.2. Things to be taken care of
    - Extensive testing has to be done as it means dropping support for CentOS based systems and moving to Linux based ones
    - Make sure the dependencies needed for running latest version of PW is installed in the HB and agent complete image.

3. **Recorder**: 
   - Synthetics formatter code will be duplicated in the recorder and will be referenced instead of importing from the Agent.
   - Recorder will pin to the latest Synthetics version prior to updating the playwright version - https://github.com/elastic/synthetics/releases/tag/v1.0.0-beta.23
   - We will continue to use the forked PW version which will recorder the scripts in Older version of Chromium and might generate code with older API syntax.

4. **Forked PW**: 
   - The patch that was done to support extending the Playwright actions would be copied over to the forked version so we can continue recording on the latest PW
   - We would also need to change the way formatter is generating code to be able to use the new API syntax.

andrewvc:
Closing in favor of https://github.com/elastic/synthetics/pull/485

PR #485, opened 2022-04-01
fix: update to the latest playwright

### LINKED ISSUE #484: Update PW to the latest version
+  Synthetics needs to be updated to the latest version of Playwright
+ Decouple the Synthetics formatter and move it to the recorder for the time being - https://github.com/elastic/synthetics-recorder/issues/185
+ Fastest way to solve some of the issues that users are experiencing with core dumps - https://github.com/elastic/beats/issues/30426

vigneshshanmugam:
+ fix #484


andrewvc (approved):
LGTM, test suite runs well and runs correctly with Elasticsearch. Nice work!

src/index.ts
```diff
@@ -53,11 +53,3 @@ export type { default as Runner } from './core/runner';
 export type { Reporter, ReporterOptions } from './reporters';
 
 export type { SyntheticsConfig } from './common_types';
-export type {
```
vigneshshanmugam, line 56: I will re add this as formatter is working now.

PR #487, opened 2022-04-01
chore: fix flaky trace events test

### LINKED ISSUE #486: flaky tests due to filmstrip trace events
+ Sometimes if the sampling frequency is too low for the Chrome tracer to capture filmstripms, the data will not be present which leads to failed tests. We should account for these scenarios in the tests.

Failed test - https://apm-ci.elastic.co/blue/organizations/jenkins/apm-agent-rum%2Felastic-synthetics/detail/PR-485/2/

1. Places to fix - https://github.com/elastic/synthetics/blob/c500e52624b1819335ccc7d5470cb94c138018c4/__tests__/plugins/tracing.test.ts#L48
2. https://github.com/elastic/synthetics/blob/c500e52624b1819335ccc7d5470cb94c138018c4/__tests__/core/runner.test.ts#L686

vigneshshanmugam:
+ fix #486

### DIFF
--- a/__tests__/plugins/tracing.test.ts
+++ b/__tests__/plugins/tracing.test.ts
@@ -44,6 +44,11 @@ describe('tracing', () => {
     await driver.page.goto(server.TEST_PAGE, { waitUntil: 'networkidle' });
     const { filmstrips, traces } = await tracer.stop();
     await Gatherer.stop();
+    /**
+     * Sometimes if the test gets completed before the sampling frequency is hit,
+     * chrome tracer would not have time to capture filmstripms, We account for
+     * these scenarios by checking them conditionally
+     */
     if (filmstrips.length > 0) {
       expect(filmstrips[0]).toMatchObject({
         blob: expect.any(String),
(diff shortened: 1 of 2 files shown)


__tests__/plugins/tracing.test.ts
```diff
@@ -37,19 +37,26 @@ describe('tracing', () => {
...
+    await driver.page.goto(server.TEST_PAGE, { waitUntil: 'networkidle' });
+    const { filmstrips, traces } = await tracer.stop();
     await Gatherer.stop();
-    expect(filmstrips.length).toBeGreaterThan(0);
-    expect(filmstrips[0]).toMatchObject({
-      blob: expect.any(String),
-      mime: 'image/jpeg',
-      start: { us: expect.any(Number) },
-    });
+    if (filmstrips.length > 0) {
```
andrewvc, line 47: If we're doing this sort of thing we should definitely have a long comment explaining why. It's just encapsulating the flakiness in the test, which is an improvement, but we should do our best to explain before merging

vigneshshanmugam: Added comment for the same.

PR #489, opened 2022-04-12
chore: add test for cached res timings

vigneshshanmugam:
+ fix https://github.com/elastic/uptime/issues/460
+ The issue itself was fixed by moving to the network events from PW instead of CDP logic #488
+ Added test case now for timings for the cached resource check. 

Screen Shot 2022-04-12 at 12 14 16 PM
Screen Shot 2022-04-12 at 12 14 23 PM

### Steps to check
1. Create inline journey with the example given on the issue.
```js
step("My Page", async () => {
   await page.goto('https://paulb-elastic.github.io/testPageWithCounter.html');
   await page.waitForSelector('#counter >> text=Seconds since page loaded: 2');
});
step("Cached version of My Page", async () => {
   await page.goto('https://paulb-elastic.github.io/testPageWithCounter.html');
   await page.waitForSelector('#counter >> text=Seconds since page loaded: 2');
});
```
2. Run the synthetics CLI from local and ensure the timing data is correct
```
cat examples/inline/example.js | node dist/cli.js --inline --rich-events
```
3. Install the current synthetics version globally `npm install -g .`and run it as inline monitor.
```yaml
  - type: browser
    id: cache-test
    name: Cached-steps
    schedule: "@every 1m"
    source:
      inline:
        script: step("My Page", async () => {
          await page.goto("https://paulb-elastic.github.io/testPageWithCounter.html");
          });
          step("Cached version of My Page", async () => {
          await page.goto('https://paulb-elastic.github.io/testPageWithCounter.html');
          });
```

### DIFF
--- a/__tests__/plugins/network.test.ts
+++ b/__tests__/plugins/network.test.ts
@@ -150,7 +150,6 @@ describe('network', () => {
     await driver.page.goto(server.PREFIX + '/index', {
       waitUntil: 'networkidle',
     });
-    await driver.page.waitForLoadState();
     await Gatherer.stop();
     const netinfo = await network.stop();
     expect(netinfo.length).toBe(2);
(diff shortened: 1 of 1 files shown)


__tests__/plugins/network.test.ts
```diff
@@ -150,7 +150,6 @@ describe('network', () => {
     await driver.page.goto(server.PREFIX + '/index', {
       waitUntil: 'networkidle',
     });
-    await driver.page.waitForLoadState();
```
vigneshshanmugam, line 153: redundant, we wait until the idle part anyway in the previous line.

justinkambic (approved):
LGTM, CI is passing, test looks good

PR #492, opened 2022-04-15
feat: introduce new reporter interface

vigneshshanmugam:
+ part of #482 
+ changes the reporter interface to not be based on Event Emitter pattern instead use a dedicated interface which makes it easier to capture errors that arise out of `UnhandledPromiseRejections`.
+ Gives a clean interface which lets users to extend the Reporter if required to create new ones like `HTMLReporter` which could be used by our recorder. 
```ts
export interface Reporter {
  onStart?(params: StartEvent): void;
  onJourneyRegister?(journey: Journey): void;
  onJourneyStart?(journey: Journey, result: JourneyStartResult): void;
  onStepStart?(journey: Journey, step: Step): void;
  onStepEnd?(journey: Journey, step: Step, result: StepEndResult): void;
  onJourneyEnd?(
    journey: Journey,
    result: JourneyEndResult
  ): void | Promise<void>;
  onEnd?(): void | Promise<void>;
}
```

### Testing
+ Checkout the code and build the code `npm run build`.
+ Play with different reporters. 
```sh
# Default one
node dist/cli.js examples/todos --config ./examples/todos/synthetics.config.ts

# Mimick HB reporter JSON
node dist/cli.js examples/todos --config ./examples/todos/synthetics.config.ts --rich-events

# Junit
node dist/cli.js examples/todos --config ./examples/todos/synthetics.config.ts --reporter junit
```
+ Test with HB by installing the build globally `npm i -g .`
+ Run any HB inline monitors to check if the new reporter changes are working correctly.


src/reporters/index.ts
```diff
@@ -26,11 +26,34 @@
...
+  StartEvent,
+  JourneyEndResult,
+  JourneyStartResult,
+  StepEndResult,
+} from '../common_types';
 
-export { Reporter, ReporterOptions };
-export const reporters = {
+export type ReporterOptions = { fd?: number; colors?: boolean };
+export type BuiltInReporter = 'default' | 'json' | 'junit';
```
lucasfcosta, line 38: Perhaps:

```suggestion
export type BuiltInReporterName = 'default' | 'json' | 'junit';
```

?

lucasfcosta (approved):
LGTM, tested all the reporters using the `--reporter` option. All of them generated OK output.

Also ran Heartbeat with a browser monitor to ensure the `--rich-events` flag was still emitting adequate events.

PR #502, opened 2022-05-12
feat: scaffold a synthetics node project

vigneshshanmugam:
+ part of #470 
+ Add a new scaffolding command called `init` to the CLI. 
+ `npx @elastic/synthetics init <project-dir>` - You would get everything set up from running tests locally to pushing monitors to the Kibana Monitor management. After scaffolding, the users could do 
      - `npm run test ` - Runs the synthetics tests locally
      - `npm run push` - Pushes the local journeys with configured monitor values to the uptime monitor management UI.

### DIFF
--- a/src/generator/index.ts
+++ b/src/generator/index.ts
@@ -54,7 +54,7 @@ export class Generator {
     );
 
     // Setup example journey file
-    const journeyFile = 'journeys/todos.journey.ts';
+    const journeyFile = 'journeys/example.journey.ts';
     fileMap.set(
       journeyFile,
       await readFile(join(templateDir, journeyFile), 'utf-8')
@@ -122,7 +122,7 @@ export class Generator {
 
     // Add push command
     const project = basename(this.projectDir);
-    pkgJSON.scripts.push = `npx @elastic/synthetics push journeys --project ${project} --url http://localhost:5601 --auth <apiKey|basic-auth>`;
+    pkgJSON.scripts.push = `npx @elastic/synthetics push journeys --project ${project} --url http://localhost:5601 --auth basic-auth`;
 
     await this.createFile(
       filename,
--- a/src/generator/index.ts
+++ b/src/generator/index.ts
@@ -25,7 +25,7 @@
 import { execSync } from 'child_process';
 import { existsSync } from 'fs';
 import { mkdir, readFile, writeFile } from 'fs/promises';
-import { bold, cyan } from 'kleur/colors';
+import { bold, cyan, yellow } from 'kleur/colors';
 import { join, relative, dirname, basename } from 'path';
 import { progress, write as stdWrite } from '../helpers';
 import { getPackageManager, runCommand } from './utils';
@@ -122,7 +122,7 @@ export class Generator {
 
     // Add push command
     const project = basename(this.projectDir);
-    pkgJSON.scripts.push = `npx @elastic/synthetics push journeys --project ${project} --url http://localhost:5601 --auth elastic:changeme`;
+    pkgJSON.scripts.push = `npx @elastic/synthetics push journeys --project ${project} --url http://localhost:5601 --auth <apiKey|basic-auth>`;
 
     await this.createFile(
       filename,
@@ -139,6 +139,10 @@ All set, you can run below commands inside: ${this.projectDir}:
   Run synthetic tests: ${cyan(runCommand(this.pkgManager, 'test'))}
 
   Push monitors to Kibana: ${cyan(runCommand(this.pkgManager, 'push'))}
+
+  ${yellow(
+    'Make sure to update the Kibana url and authentication info before pushing monitors to Kibana.'
+  )}
 
 Visit https://www.elastic.co/guide/en/observability/master/synthetics-journeys.html to learn more.
     `)
--- a/src/generator/index.ts
+++ b/src/generator/index.ts
@@ -1,0 +1,155 @@
+/**
+ * MIT License
+ *
+ * Copyright (c) 2020-present, Elastic NV
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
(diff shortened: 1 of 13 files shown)


src/generator/index.ts
```diff
@@ -0,0 +1,155 @@
...
+    fileMap.set(
+      configFile,
+      await readFile(join(templateDir, configFile), 'utf-8')
+    );
+
+    // Setup example journey file
+    const journeyFile = 'journeys/todos.journey.ts';
+    fileMap.set(
+      journeyFile,
+      await readFile(join(templateDir, journeyFile), 'utf-8')
```
lucasfcosta, line 60: IMO we should actually have a simpler journey file as the first example. Maybe accessing elastic's website or something along those lines.

I found the current example a bit too convoluted.

templates/journeys/todos.journey.ts
```diff
@@ -0,0 +1,18 @@
...
+  });
+  step('launch application', async () => {
+    await page.goto(params.url);
+  });
+
+  step('assert title', async () => {
+    const header = await page.$('h1');
+    expect(await header.textContent()).toBe('todos');
+  });
+});
```
lucasfcosta, line 18: IMO this could be a simpler example against a live webpage.

templates/synthetics.config.ts
```diff
@@ -0,0 +1,16 @@
...
+    params: {
+      url: 'https://elastic.github.io/synthetics-demo/',
+    },
+    playwrightOptions: {
+      ignoreHTTPSErrors: true,
+    },
+  };
+  if (env !== 'development') {
+    config.params.url = 'https://elastic.github.io/synthetics-demo/';
+  }
```
lucasfcosta, line 14: Not sure we need this in the template.

src/generator/index.ts
```diff
@@ -0,0 +1,155 @@
...
+
+    if (!pkgJSON.scripts) {
+      pkgJSON.scripts = {};
+    }
+    // Add test command
+    pkgJSON.scripts.test = 'npx @elastic/synthetics journeys';
+
+    // Add push command
+    const project = basename(this.projectDir);
+    pkgJSON.scripts.push = `npx @elastic/synthetics push journeys --project ${project} --url http://localhost:5601 --auth elastic:changeme`;
```
lucasfcosta, line 125: Should we add an indication that one may need to change the URL here in some way?

lucasfcosta (approved):
LGTM!

Also E2E tested by:

1. Checking out this branch
2. Running `npm link`
3. Running `synthetics init`
4. Checking the created folder, running and pushing its examples.

Just had UX considerations wrt to the files created, but none are blockers.

src/generator/index.ts
vigneshshanmugam: So the main reason for not touching any external systems is that It generates robots traffic with Synthetics user agent and kind of being their pain when it comes to analytics. I would still to a simpler todo mvc page or any page but not a company website.

templates/journeys/todos.journey.ts
vigneshshanmugam: Feel free to suggest, I am open to new examples.

templates/synthetics.config.ts
vigneshshanmugam: Thought it would be helpful for users to find it easier to override params if desired.

src/generator/index.ts
vigneshshanmugam: Hmmm, Cant think of anything at the top of my mind. Thoughts?

vigneshshanmugam: We can add some helpful message on the banner text? 

```
Push monitors to Kibana: ${cyan(runCommand(this.pkgManager, 'push'))}
 
Make sure to change url and auth for Kibana for pushing the monitors

```

src/generator/index.ts
lucasfcosta: How would you feel about using `example.com`, which was [built by IANA](https://www.iana.org/domains/reserved) exactly for this purpose?

I was thinking of something along the lines of:

```js
import { journey, step, monitor, expect } from '@elastic/synthetics';

journey('Check Example Website Title', ({ page, params }) => {
  // Only relevant for the push command to create
  // monitors in Kibana
  monitor.use({
    id: 'example-monitor',
    schedule: 10,
  });
  
  step('launch website', async () => {
    await page.goto("https://example.com");
  });

    expect(await header.textContent()).toBe('Example Domain');
  });
});
```

If not doing that then I think just the file below (with no references to TodoMVC) would be enough:

```js
import { journey, step, monitor, expect } from '@elastic/synthetics';

journey('My Example Journey', ({ page, params }) => {
  // Only relevant for the push command to create
  // monitors in Kibana
  monitor.use({
    id: 'example',
    schedule: 10,
  });
  
    await page.goto(params.url);
  });

    expect(await header.textContent()).toBe('My Webpage');
  });
});
```

templates/synthetics.config.ts
lucasfcosta: Hmm, I'm not strongly opinionated about that, but I'm fine to keep it. If we end up going with the `example.com` environment then we can just change this to `example.org` and have the default URL being `example.com` I guess.

src/generator/index.ts
lucasfcosta: That sounds like enough to me when it comes to the URL, although I'd set  `USERNAME` and `PASSWORD` to be taken from environment variables so that users don't end-up just putting theirs in `package.json` because they're distracted.

vigneshshanmugam: Very good point.

src/generator/index.ts
vigneshshanmugam: I do like the example.com idea, But in our docs we always refer to TodoMVC so I am bit skeptical here. @andrewvc WDYT about using example.com for the scaffolding?

src/generator/index.ts
vigneshshanmugam: I am going to leave it empty with the pattern `auth <apikey|basic-auth>` and provide banner text with explanation as I am bit scared to use someother auth details user had in their env variables which would hit the Kibana server and might get logged.

src/generator/index.ts
andrewvc: What about https://elastic.github.io/synthetics-demo/ which is a hosted version of TODO MVC we already have? Example.com is a bit basic.

vigneshshanmugam: That was the one which the PR was based on https://github.com/elastic/synthetics/blob/da32e68b484938cfd3f001200ca3c819e4464794/templates/journeys/todos.journey.ts and Lucas suggested to move to example.com

andrewvc: I do think we should remove the whole embedded webserver business, that was my doing but just confusing in the end.

PR #506, opened 2022-05-20
chore: handle 404 erorrs on push endpoint

vigneshshanmugam:
+ Part of #470 
+ Handle cases when invalid Kibana endpoint was provided. Most likely happens when development when there is more chance of base path being different between every restart.


andrewvc (approved):
LGTM

PR #508, opened 2022-05-23
feat: interactive scaffolding for picking default monitor config

vigneshshanmugam:
+ fix https://github.com/elastic/synthetics-dev/issues/119
+ part of #513 
+ `init` command scaffolding asks interactive questions for both `Locations` and `Schedule` which would be set as default on the Synthetics.config.ts file to be used when pushing the monitors. 
+ Error out when push command is invoked without passing the journey file paths
+ Make the location and schedule option as required `--locations <>` in the CLI and error out when not set. 
+ Allow users to configure the default values via `synthetics.config.ts` file

### How to test
1. Build the agent `npm run build`
2. Run `node dist/cli.js init test-directory`
3. Verify if it sets up all the default schedule and locations in the `synthetics.config.ts` generated file.

### DIFF
--- a/__tests__/core/runner.test.ts
+++ b/__tests__/core/runner.test.ts
@@ -29,10 +29,13 @@ import Runner from '../../src/core/runner';
 import { step, journey } from '../../src/core';
 import { Journey, Step } from '../../src/dsl';
 import { Server } from '../utils/server';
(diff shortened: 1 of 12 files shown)


andrewvc (approved):
LGTM

PR #510, opened 2022-05-24
fix: write .gitnore files for scaffolding

### LINKED ISSUE #509: Synthetics init fails because `templates` weren't included in distributable
# Summary

Right now, the `synthetics init` command fails because the `templates` folder is lacking files. If you build the package locally it runs fine because the templates will be copied, but if you use `npx`, for example, you will see the following error:

```
Error: ENOENT: no such file or directory, open '/Users/lucasfcosta/.npm/_npx/XYZ/node_modules/@elastic/synthetics/templates/.gitignore'
```

# How to reproduce

1. Run `npx @elastic/synthetics init my-project`

vigneshshanmugam:
+ fix #509 
+ We update the `.gitignore` contents if the init is called on an existing directory and override it instead of using it from the templates directory.

### DIFF
--- a/templates/.gitignore
+++ b/templates/.gitignore
@@ -1,2 +1,0 @@
-node_modules
-.synthetics
--- a/__tests__/generator/index.test.ts
+++ b/__tests__/generator/index.test.ts
@@ -24,7 +24,7 @@
  */
 
 import { existsSync } from 'fs';
-import { rm } from 'fs/promises';
+import { readFile, rm } from 'fs/promises';
 import { join } from 'path';
 import { CLIMock } from '../utils/test-config';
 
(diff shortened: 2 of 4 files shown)


templates/.gitignore
```diff
@@ -1,2 +0,0 @@
-node_modules
```
lucasfcosta, line 1: It's funny that the template `.gitignore` was ignored itself and thus we had this bug 😆

lucasfcosta (approved):
@vigneshshanmugam I did just test again locally and it works. I've also checked NPM docs to be sure that the root cause analysis was 100% correct.

I also found this: https://github.com/npm/npm/issues/3763. Apparently we could also just add an `.npmignore` to prevent the `.gitignore` from being moved as `.npmignore`.

However, the ultimate test here would be to deploy that to NPM and check whether it works there too 😄  Can we do a release under a beta tag to try this one out btw?

vigneshshanmugam:
> However, the ultimate test here would be to deploy that to NPM and check whether it works there too 😄 Can we do a release under a beta tag to try this one out btw?

We publish everything under beta as we are in the Beta stage. So should be good here.

PR #512, opened 2022-05-24
feat: use apiKey instead of basic auth

### LINKED ISSUE #511: use apiKey for pushing monitors instead of basic auth
+ Now that we have the Kibana side of the PR ready - https://github.com/elastic/kibana/pull/130920  to go in the next release, we can switch to ApiKey - https://www.elastic.co/guide/en/kibana/master/api-keys.html#api-keys for communicating with Kibana for pushing monitors instead of using Basic Auth.

vigneshshanmugam:
+ fix #511 
+ Sends the push request using the format `Authorization: APiKey <token>` instead of basic auth headers - https://www.elastic.co/guide/en/kibana/master/api-keys.html

### DIFF
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -158,7 +158,7 @@ program
   .addOption(
     new Option(
       '--locations <locations...>',
-      'The default list of locations from which your monitors will run.'
+      'default list of locations from which your monitors will run.'
     ).choices(SyntheticsLocations)
   )
   .requiredOption(
@@ -168,7 +168,7 @@ program
   .requiredOption('--url <url>', 'kibana URL to upload the monitors')
   .requiredOption(
     '--auth <auth>',
-    'user authentication (in user:password format) or API key that will be used to communicate with Kibana.'
+    'API key used for Kibana authentication(https://www.elastic.co/guide/en/kibana/master/api-keys.html).'
   )
   .option(
     '--space <space>',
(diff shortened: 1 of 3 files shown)


src/cli.ts
```diff
@@ -168,7 +168,7 @@ program
   .requiredOption('--url <url>', 'kibana URL to upload the monitors')
   .requiredOption(
     '--auth <auth>',
-    'user authentication (in user:password format) or API key that will be used to communicate with Kibana.'
+    'API key used for Kibana authentication(https://www.elastic.co/guide/en/kibana/master/api-keys.html).'
```
dominiqueclarke, line 171: It seems like we should link to the specific project monitors docs, with a section about generating your api key, rather than the generic documentation. Going through Uptime to create your api key will ensure that the api key they create has the appropriate permissions. In fact, we don't have a good story for creating an api key outside of uptime, because the api key management page doesn't currently support api keys for kibana permissions, but our uptime implementation does.

dominiqueclarke (approved):
LGTM. Tested e2e with api key generated from Uptime. I think we should update the docs at a later point, however.

src/cli.ts
vigneshshanmugam: Good point, Would we have those docs available post 8.3? I am happy to do a followup.

PR #514, opened 2022-05-26
fix: only append 'Elastic/Synthetics' to default UA

andrewvc:
When we originally implemented https://github.com/elastic/synthetics/issues/232 it appears we didn't implement the third AC, allowing users to override our custom UA string "Elastic/Synthetics" which is appended.

Some users need to fully replace the UA, this allows them to do that. Rather than adding a boolean, this assumes that if you're overriding the UA you want full control, which is more in line with the principle of least surprise.

CC @paulb-elastic @drewpost 

PS this is currently a draft because there are no tests, I'll add tests if we agree on this approach.

### DIFF
--- a/__tests__/core/gatherer.test.ts
+++ b/__tests__/core/gatherer.test.ts
@@ -117,7 +117,7 @@ describe('Gatherer', () => {
   });
 
   describe('Elastic UA identifier', () => {
-    it('works on a single page', async () => {
+    it('present when UA is not overriden', async () => {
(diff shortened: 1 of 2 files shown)


vigneshshanmugam (commented):
@andrewvc Should be good to go after we fix the tests.

vigneshshanmugam (approved):

shahzad31:
POST FF Testing looks good !!

Issue #523, opened 2022-06-07
Users should only be able to push from the current folder


lucasfcosta:
# Summary

When using `push`, [the agent will use the `cwd` as the path from which it will read a configuration file](https://github.com/elastic/synthetics/blob/201f7da7db8d8ce9c13855023c8566b73ebaf91a/src/config.ts#L32).

That behaviour is problematic because if the user is using the agent to push journeys from a different target folder, they'll read the incorrect configuration file (or none), not the configuration in the target folder.

**Therefore, we have decided that we will only allow pushing from the _current working directory_. That eliminates the confusion people could have if they were pushing multiple different folders (wondering which config would be used for each).**

# How to reproduce

1. Create a Synthetics project using the `init` command
2. Update `config` file in the new folder created by `init`, changing its schedule
   ```ts
    import type { SyntheticsConfig } from '@elastic/synthetics';

    export default env => {
      const config: SyntheticsConfig = {
        params: {
          url: 'https://elastic.github.io/synthetics-demo/',
        },
        playwrightOptions: {
          ignoreHTTPSErrors: false,
        },
        /**
         * Configure global monitor settings
         */
        monitor: {
          schedule: 25,
          locations: ["brazil"],
        },
      };
      if (env !== 'development') {
        /**
         * Override configuration specific to environment
         * Ex: config.params.url = ""
         */
      }
      return config;
    };
   ```
3. Copy this config file to `/tmp` and change the `schedule`'s value to something else, like `50`, for example.
4. Navigate (`cd`) into `/tmp` and run the push command from there, specifying the other directory (the one you created with `init`)
5. You'll see the monitors get created with the schedule from `/tmp`, not the one within the target folder's config file.

The desired behaviour is to always use `cwd` and not allow a path to be passed, as we discussed on Slack.

lucasfcosta:
I have repurposed this issue to reflect our discussions on Slack.

shahzad31:
Push command works without following error now

```
errors: [
    {
      detail: [Error],
      id: '',
      location: null,
      notes: [Array],
      pluginName: 'esbuild-multiasset-plugin',
      text: "ENOENT: no such file or directory, open '/Users/shahzad-16/elastic/synthetics/examples/todos/file:/Users/shahzad-16/elastic/synthetics/examples/todos/basic.journey.ts'"
    }
  ],
```

lucasfcosta:
[...]

PR #526, opened 2022-06-09
Allow pushing single paths only [+ fix flaky tests]

lucasfcosta:
# Summary

This PR solves #523 by allowing users to only upload the current path of monitors.

Because now users could only point to a single target file, I made sure to implement the `--match` and `--tags` flags so that users could filter journeys. Otherwise, the ergonomics of the command could become inconvenient.

# Implementation Details

The implementation of the changes described above was quite simple: I basically pass the path to the CWD when we require the journeys so that we can still reuse the same code we had before to load them.

Then, I went on and used the same matching function to make sure we'd only bundle monitors matching the tags or names specified.

# Flaky tests

The flaky tests were a funny one.

The reason they were flaky is that one of the tests would sometime take longer to return the output, and, therefore, by the time it returned it, there was more than one event in it.

```
  {"type":"journey/start","@timestamp":1654791216548408.8,"journey":{"name":"inline","id":"inline"},"root_fields":{"os":{"platform":"darwin"},"package":{"name":"@elastic/synthetics","version":"1.0.0-beta.27"}},"payload":{"params":{"url":"non-dev"},"source":"({ page, context, browser, params }) => {\n        scriptFn.apply(null, [core_1.step, page, context, browser, params, expect_1.expect]);\n    }"},"package_version":"1.0.0-beta.27"}
    {"type":"step/end","@timestamp":1654791216551033.8,"journey":{"name":"inline","id":"inline"},"step":{"name":"fake step","index":1,"status":"succeeded","duration":{"us":599}},"root_fields":{"os":{"platform":"darwin"},"package":{"name":"@elastic/synthetics","version":"1.0.0-beta.27"}},"payload":{"source":"async () => {}","url":"about:blank","status":"succeeded"},"url":"about:blank","package_version":"1.0.0-beta.27"}
    {"type":"journey/end","@timestamp":1654791216552023.8,"journey":{"name":"inline","id":"inline","status":"succeeded"},"root_fields":{"os":{"platform":"darwin"},"package":{"name":"@elastic/synthetics","version":"1.0.0-beta.27"}},"payload":{"start":288041.905244907,"end":288041.909911118,"status":"succeeded"},"package_version":"1.0.0-beta.27"}
```

As you can see above, one of these events is opening a blank page.

So what was happening was: sometimes there would be enough time to open a blank page and return that in the output, and sometimes there wouldn't be, in which case we'd get a single line:

```
  {"type":"journey/start","@timestamp":1654791216548408.8,"journey":{"name":"inline","id":"inline"},"root_fields":{"os":{"platform":"darwin"},"package":{"name":"@elastic/synthetics","version":"1.0.0-beta.27"}},"payload":{"params":{"url":"non-dev"},"source":"({ page, context, browser, params }) => {\n        scriptFn.apply(null, [core_1.step, page, context, browser, params, expect_1.expect]);\n    }"},"package_version":"1.0.0-beta.27"}
```

Now, `JSON.parse` could parse a single line, but not multiple ones. Therefore, we'd see an error.

Then, because cleanup (the assignment to `NODE_ENV` and its restoration) was being done _within_ the test, it was skipped when the test failed, thus causing other tests to fail.

I fixed that by parsing the output line-by-line and picking the correct event + moving cleanup to the top of the file on a hook.


__tests__/generator/index.test.ts
```diff
@@ -42,6 +42,9 @@ describe('Generator', () => {
     });
   });
   it('generate synthetics project - NPM', async () => {
+    // In a few development environments, this test may take
+    // a few millisseconds more than the default 15000ms timeout
+    jest.setTimeout(30000);
```
vigneshshanmugam, line 47: Its already this way - https://github.com/elastic/synthetics/blob/de8feeec94558b78268d54b2d14f5e60d80c1039/__tests__/generator/index.test.ts#L89

src/cli.ts
```diff
@@ -147,14 +148,20 @@ program
...
   .command('push')
-  .addArgument(
-    new Argument(
-      '[journeys...]',
-      'file path to journey directory and individual files'
-    ).argRequired()
-  )
   .description(
-    'Push journeys to create montors within Kibana monitor management UI'
+    'Push journeys in the current directory to create montors within the Kibana monitor management UI'
```
vigneshshanmugam, line 152: typo: monitors

src/cli.ts
```diff
@@ -147,14 +148,20 @@ program
...
+  )
+  .option(
```
vigneshshanmugam, line 154: Not sure why we need these options here? Subcommands should be able to access these `pattern, tags, match` without any problem IIRC.

vigneshshanmugam (commented):
Great catch on the NODE_ENV stuff.

andrewvc:
I've merged in some changes that fix the remaining failures. ae66e63 solved a lot of the problems, with 989cd7d solving the rest. The cert stuff in retrospect was a red herring

lucasfcosta (commented):
@andrewvc Just looked through the changes you added, thanks for those! They LGTM.

@vigneshshanmugam also fixed the typo and responded to your comment about the `option`s.

I'll go ahead and merge this one once I get an approval from one of you if that's okay.

Please feel free to just give this a 👍 when you think we can merge it.

andrewvc:
I'm ++ on merging this ASAP!

PR #532, opened 2022-06-15
fix: allow controlling timeouts via actions

vigneshshanmugam:
+ Playwright by default waits for action timeouts which is set to 30 seconds. So if the user calls `page.go` without explicitly setting timeout, then synthetics agent would wait for that timeout of 30 seconds plus the timeout that the synthetics agent set for the action `page.waitForLoadEvent`. This was a legacy code which was put there to make sure we capture the screenshot after page load for all steps. 
+ The PR addresses the problem by not waiting for the page load as not each step would end up in a load event and we explicitly set a screenshot timeout of 5 seconds. If the agent cannot capture the screenshot within that timeframe, it would end up not writing that document. 
+ Source of this issue is found from the Synthetics service.

### How to test
1. Use an inline journey
```
step('timeout step', async () => {
  await page.goto('http://www.elastic.co', {timeout: 5000});
});
```
2. Check if the step waits only for the above timeout + screenshot timeout (which can vary depending on how long it takes - but no more than 5000ms)
3. Make sure, we dont wait for the default 30 seconds timeout.

### DIFF
--- a/__tests__/core/runner.test.ts
+++ b/__tests__/core/runner.test.ts
@@ -644,6 +644,36 @@ describe('runner', () => {
     });
   });
 
+  it('run - use step timeouts', async () => {
+    const delay = ms => new Promise(resolve => setTimeout(resolve, ms));
+    server.route('/delay500', async (req, res) => {
+      await delay(500);
+      res.end('delay 500');
+    });
+    const j1 = journey('timeout', async ({ page }) => {
+      step('delayed step', async () => {
+        // keep 200ms timeout for navigation
+        await page.goto(server.PREFIX + '/delay500', { timeout: 200 });
+      });
+    });
+    await runner.addJourney(j1);
+
+    class StepCheckReporter implements Reporter {
+      onStepEnd(j, s, result: StepEndResult) {
+        const durationMs = (result.end - result.start) * 1000;
+        // Should not wait till the page is loaded which is after 500ms delay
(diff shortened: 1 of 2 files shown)


lucasfcosta (approved):
LGTM, also E2E tested by using the aforementioned inline journey:

```
➜ pbpaste | node ./dist/cli.js --inline

Journey: inline
   ✖  Step: 'timeout step' failed (5142 ms)
   ---
      stack: |-
        page.goto: Timeout 5000ms exceeded.
        =========================== logs ===========================
        navigating to "http://www.elastic.co/", waiting until "load"
        ============================================================
            at Step.eval [as callback] (eval at loadInlineScript (/Users/lucasfcosta/Repositories/synthetics/src/loader.ts:89:20), <anonymous>:4:14)
            at Runner.runStep (/Users/lucasfcosta/Repositories/synthetics/src/core/runner.ts:213:18)
            at Runner.runSteps (/Users/lucasfcosta/Repositories/synthetics/src/core/runner.ts:263:27)
            at Runner.runJourney (/Users/lucasfcosta/Repositories/synthetics/src/core/runner.ts:353:38)
            at async Runner.run (/Users/lucasfcosta/Repositories/synthetics/src/core/runner.ts:448:11)
            at async Command.<anonymous> (/Users/lucasfcosta/Repositories/synthetics/src/cli.ts:131:23)
   ---
```

Issue #533, opened 2022-06-16
Remove the –delete option


paulb-elastic:
On reflection of the behaviour of project monitors, the `--delete` option should be removed, and this should become the default behaviour (i.e. `push` will always delete monitors that were defined in that project, which are no longer in the source folder being `push`ed)

ACs:
* When `push`ing project monitors to Kibana, monitors that were previously defined in the source folder (that exist in Kibana), will be deleted if they no longer exist in the source, when `push`ed
  * Only the monitor configuration (saved object) will be removed, not result data
* The `--delete` option is removed as a CLI option
* The help text is updated to remove references to `--delete`

PR #534, opened 2022-06-16
fix: delete stale monitors by default on push command

shahzad31:
fixes https://github.com/elastic/synthetics/issues/533

Remove --delete option

### DIFF
--- a/src/push/request.ts
+++ b/src/push/request.ts
@@ -23,7 +23,7 @@
  *
  */
 
-import { bold, gray, yellow } from 'kleur/colors';
+import { bold } from 'kleur/colors';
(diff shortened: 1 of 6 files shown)


src/push/request.ts
```diff
@@ -34,7 +34,6 @@ const { version } = require('../../package.json');
 
 export type APISchema = {
   project: string;
-  keep_stale: boolean;
```
vigneshshanmugam, line 37: We would need this flag if i remember correctly from my conversation with Dominique.

dominiqueclarke: @shahzad31 actually decided to delete that requirement from the Kibana side as well.

PR #540, opened 2022-06-27
fix: use fixed playwright version

vigneshshanmugam:
+ Broken by the PW upgrade - https://github.com/elastic/synthetics/pull/485, we have always used to fix the version of PW and reduce the breakage footprint that comes with new versions which might have new system dependencies. 
+ Without this zip urls monitors that uses the latest version of the Synthetics package will install PW versions that are the latest version `1.23.0` that also requires other system dependencies and fails to invoke the Synthetics. 

Produces validation errors like this.
```
 Host system is missing dependencies to run browsers

║ Please install them with the following command: ║
║ sudo npx playwright install-deps ║
║ Alternatively, use apt: ║
║ sudo apt-get install libwayland-client0 ║

```


andrewvc:
I'm not sure that goes back far enough. I tested that version of playwright by running the current todos demo with a modified package.json shown below in a `heartbeat:7.3.0-SNAPSHOT` container

```json
{
  "name": "todos",
  "private": true,
  "description": "This suite tests the examples that ship with the open source Vue.js project.",
  "scripts": {},
  "license": "MIT",
  "dependencies": {
    "@elastic/synthetics": "*",
    "playwright-core": "=1.20.1"
  },
  "devDependencies": {
    "node-static": "^0.7.11"
  }
}
```

I still got the error:

```

heartbeat@2bc283fe3c0f:/tmp/synthetics-demo/todos/synthetics-tests$ npx @elastic/synthetics .
browserType.launch:
╔══════════════════════════════════════════════════════╗
║ Host system is missing dependencies to run browsers. ║
║ Please install them with the following command:      ║
║                                                      ║
║     sudo npx playwright install-deps                 ║
║                                                      ║
║ Alternatively, use apt:                              ║
║     sudo apt-get install libwayland-client0          ║
║                                                      ║
║ <3 Playwright Team                                   ║
╚══════════════════════════════════════════════════════╝
```

andrewvc (approved):
LGTM, pinning both core and chromium fixed it

PR #556, opened 2022-07-14
Upgrade proxy agent

andrewvc:
Upgrade a number of dependencies, including the proxy agent `undici`. This wasn't quite so simple, upgrading that library had a cascading effect requiring the updating of many libs which in turn required changing the `expect` definitions 🤷 .


devcorpio (approved):
LGTM!

Issue #557, opened 2022-07-18
[Meta] Synthetics Agent enhancements for Private Node


dominiqueclarke:
To support private node, and enhance the current experience of the project monitors, updates are needed to the synthetics `init` command, `monitor.use` DSL, and `push` command line arguments.

Enhancements to `init`
--
- [ ] The new init command will update the existing questionnaire to the following options. Exact content/phrasing TBD.

* Are you using elastic cloud? Y/N
     Y: What is your cloud.id (You can find this....) ? 
     N: What is the url of your Kibana instance?
* What api key would you like to use for auth? (Include link to Kibana to set up API Key)
* Choosing a default location
     * IF locations are available, user will be able to select froma list of locations that includes both publc (if in cloud), and private locations together, with each location clearly indicating if it's public or private. 
     * IF locations are not available (for example, an on-prem cluster that has not yet set up private locations): A link to docs for how to set up your first private location is displayed
 * Choose default interval
 * Choose project id
 * Choose space
 
DSL experience with `monitor.use`
--
 - [x] Add new key for private locations `privateLocations`
 
`push` command override
--
- [x] Add new --private-locations option to the push command

Locations discovery
--
- [ ] Provide additional discovery for new locations via new synthetics discovery command: https://github.com/elastic/synthetics/issues/551

PR #559, opened 2022-07-20
feat: support private locations in dsl and push command arguments

dominiqueclarke:
Relates to https://github.com/elastic/synthetics/issues/557

If desired, this PR can be tested e2e via the corresponding Kibana PR: https://github.com/elastic/kibana/pull/136777

Enables the ability to define `privateLocations` via the monitor dsl or the `push` command arguments. 

Example usage via `monitor.use`

```
journey('check-if-title-is-present', ({ page, params }) => {
  monitor.use({
    schedule: 3,
    privateLocations: ["YOUR_LOCATION_NAME"],
  })
    await page.goto(params.url);
  });

  });
});
```

Example usage via command-line.
`npx @elastic/synthetics --url YOUR_KIBANA_URL --project test-project --auth YOUR_AUTH_KEY --schedule 3 --locations us_east --privateLocations "YOUR_DEFAULT_PRIVATE_LOCATION_NAME"`

### DIFF
--- a/src/push/monitor.ts
+++ b/src/push/monitor.ts
@@ -36,7 +36,8 @@ export type MonitorSchema = Omit<MonitorConfig, 'locations'> & {
   filter: Monitor['filter'];
 };
 
-function translateLocation(locations: MonitorConfig['locations']) {
+function translateLocation(locations?: MonitorConfig['locations']) {
+  if (!locations) return [];
   return locations.map(loc => LocationsMap[loc]).filter(Boolean);
 }
 
--- a/__tests__/core/runner.test.ts
+++ b/__tests__/core/runner.test.ts
@@ -741,6 +741,7 @@ describe('runner', () => {
     runner.updateMonitor({
       schedule: 5,
       locations: ['us_east'],
+      privateLocations: ['germany'],
       throttling: { download: 100, upload: 50 },
       params: { env: 'test' },
       playwrightOptions: { ignoreHTTPSErrors: true },
@@ -752,6 +753,7 @@ describe('runner', () => {
       id: 'test-j1',
       schedule: 2,
       locations: ['united_kingdom'],
+      privateLocations: ['spain'],
     });
     j2.updateMonitor({ throttling: { latency: 1000 } });
     runner.addJourney(j1);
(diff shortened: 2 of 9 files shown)


vigneshshanmugam (commented):
Couple of things missing. 

1. Private location is not normalized with Synthetics config and CLI Options - https://github.com/elastic/synthetics/blob/d248a557cd0c7657c7219fb1277fda13e19b41c9/src/options.ts#L138

2. Its not added as part of the Schema sent to the Kibana server - https://github.com/elastic/synthetics/blob/d248a557cd0c7657c7219fb1277fda13e19b41c9/src/push/monitor.ts#L63. Not sure how it worked with Kibana.

dominiqueclarke:
> 2. Its not added as part of the Schema sent to the Kibana server - https://github.com/elastic/synthetics/blob/d248a557cd0c7657c7219fb1277fda13e19b41c9/src/push/monitor.ts#L63
>    . Not sure how it worked with Kibana.

It ends up being bundled and spread under `config`, I believe. Also, private locations does not need a translation helper, because they are sent wholesale as the location name.

src/push/monitor.ts
```diff
@@ -36,7 +36,8 @@ export type MonitorSchema = Omit<MonitorConfig, 'locations'> & {
   filter: Monitor['filter'];
 };
 
-function translateLocation(locations: MonitorConfig['locations']) {
+function translateLocation(locations?: MonitorConfig['locations']) {
+  if (!locations) return [];
```
vigneshshanmugam, line 40: Do we need this? Does privateLocations also expect a default array when not set?

shahzad31: i think this can be removed. i was getting some validation error But it was probably unrelated to this.

vigneshshanmugam (approved):
LGTM

src/push/monitor.ts
dominiqueclarke: @vigneshshanmugam, actually I looked it over again, and this allows a user to only define private locations if desired. Otherwise, you get a reference error for the map function. This will be critical for off-cloud users.

We can also define a default value instead of the conditional. 

```
function translateLocation(locations: MonitorConfig['locations'] = []) {
  return locations.map(loc => LocationsMap[loc]).filter(Boolean);
}
```
@vigneshshanmugam let me know your preferences.

vigneshshanmugam: Gotcha, thanks for the details. I am fine with both. Let's get this in.

PR #560, opened 2022-07-21
Add ability to tag a stack release

amannocci:
## Description

* Add the ability to tag a stack release as part of the main process.

Signed-off-by: Adrien Mannocci <adrien.mannocci@elastic.co>

### DIFF
--- a/.ci/Jenkinsfile
+++ b/.ci/Jenkinsfile
@@ -163,26 +163,27 @@ pipeline {
       }
       agent { label 'ubuntu-18 && immutable' }
       stages {
-        stage('Notify') {
+        stage('Approve to Release') {
           options { skipDefaultCheckout() }
           steps {
             notifyStatus(slackStatus: 'warning', subject: "[${env.REPO}] Release ready to be pushed",
                          body: "Please go to (<${env.BUILD_URL}input|here>) to approve or reject within 12 hours.")
+            setEnvVar('RELEASE', prompt(message: 'Should we release a new version?'))
           }
         }
         stage('Release CI') {
-          options { skipDefaultCheckout() }
-          input {
-            message 'Should we release a new version?'
-            ok 'Yes, we should.'
+          when {
+            beforeAgent true
+            expression { return env.RELEASE == 'true' }
           }
+          options { skipDefaultCheckout() }
           environment {
             NPM_PACKAGE = '@elastic/synthetics'
           }
           steps {
             cleanup()
-            withNodeJSEnv(){
-              dir("${BASE_DIR}"){
+            withNodeJSEnv() {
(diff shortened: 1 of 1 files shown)


.ci/Jenkinsfile
```diff
@@ -220,6 +220,66 @@ pipeline {
         }
       }
     }
+    stage('Stack Release Tag') {
```
v1v, line 223: I'd say to include this particular stage as a nested stage in `stage('Release') {`

.ci/Jenkinsfile
```diff
@@ -220,6 +220,66 @@ pipeline {
...
+                         body: "Please go to (<${env.BUILD_URL}input|here>) to approve or reject within 12 hours.")
+          }
+        }
+        stage('Stack Release CI') {
+          options { skipDefaultCheckout() }
+          input {
+            message 'Should we tag the current version as a stack release?'
+            ok 'Yes, we should.'
+          }
+          environment {
```
v1v, line 249: Will this `input` support for a `cancel`? If so, what's gonna happen?

.ci/Jenkinsfile
```diff
@@ -220,6 +220,66 @@ pipeline {
...
+                    setEnvVar('VERSION', sh(script: 'npm show . version', label: 'tag the version as stack release', returnStdout: true).trim())
+                    sh(script: "npm dist-tag --otp=${env.TOTP_CODE} add ${env.NPM_PACKAGE}@${env.VERSION} stack_release", label: 'tag stack_release')
+                  }
+                }
+              }
+            }
+          }
+          post {
+            success {
+              notifyStatus(slackStatus: 'good', subject: "[${env.REPO}] Stack Release '${env.VERSION}' published", body: "Great news, the stack release has been done successfully. (<${env.RUN_DISPLAY_URL}|Open>). \n Release URL: (<https://preview.npmjs.com/package/${env.NPM_PACKAGE}/v/${env.VERSION}|${env.VERSION}>)")
```
v1v, line 267: This URL matches with the one in L207, is the same URL? Or a different one pointing to `stack_release`?

amannocci: It should be the same URL since there is only one page where we can find the tag and the version.
For example: https://www.npmjs.com/package/@elastic/synthetics/v/1.0.0-beta.31

.ci/Jenkinsfile
amannocci: I moved the whole logic within the release stage.

.ci/Jenkinsfile
amannocci: I switched to a prompt pattern instead of using an input pattern.

cachedout (approved):

PR #562, opened 2022-07-21
Add a jenkins job to allow dist-tag

amannocci:
## Description

* Early support for dist-tag using a dedicated Jenkins job.
* Allow to dist-tag a specific version independently.
* We need to merge it to build able to run it 🐔 🥚 .

Ref: https://github.com/elastic/observability-robots/issues/1248

Signed-off-by: Adrien Mannocci <adrien.mannocci@elastic.co>


v1v (commented):
Will this be a Jenkins Multibranch Pipeline? If so, I'd say to split this PR in two:
1) the one with `.ci/jobs`
2) the one with `.ci/dist-tag.groovy`

This will help to test the changes for `dist-tag.groovy` , since the `JJBB` definition is quite straight forward

.ci/jobs/dist-tag-mbp.yml
```diff
@@ -0,0 +1,42 @@
...
+    script-path: .ci/dist-tag.groovy
+    scm:
+      - github:
+          branch-discovery: no-pr
+          discover-pr-forks-strategy: merge-current
+          discover-pr-forks-trust: permission
+          discover-pr-origin: merge-current
+          discover-tags: false
+          disable-pr-notifications: true
+          head-filter-regex: '^main$'
```
v1v, line 16: I'd say to add this PR, so you can test your changes if the other PR is related to the JJBB

.ci/dist-tag.groovy
```diff
@@ -0,0 +1,56 @@
...
+            withNpmrc(secret: "${env.NPMRC_SECRET}", path: "${env.WORKSPACE}/${env.BASE_DIR}") {
+              withTotpVault(secret: "${env.TOTP_SECRET}", code_var_name: 'TOTP_CODE') {
+                sh(script: "npm dist-tag --otp=${env.TOTP_CODE} add ${env.NPM_PACKAGE}@${env.VERSION} ${env.DIST_TAG}", label: "tag ${env.DIST_TAG}")
+              }
+            }
+          }
+        }
+      }
+    }
+  }
```
v1v, line 55: What do you think to send a notification to slack when this pipeline finished (either with a failure or success)?

.ci/jobs/dist-tag-mbp.yml
amannocci: I will use this PR to define the Jenkins job pipeline and move the jjbb jenkins job definition in another PR.

.ci/dist-tag.groovy
amannocci: That sounds like a good idea. I added a slack notification to report the job status.

.ci/dist-tag.groovy
```diff
@@ -0,0 +1,62 @@
...
+pipeline {
+  agent { label 'linux && immutable' }
+  environment {
+    REPO = "synthetics"
+    BASE_DIR = "src/github.com/elastic/${env.REPO}"
+    NPMRC_SECRET = 'secret/jenkins-ci/npmjs/elasticmachine'
+    TOTP_SECRET = 'totp/code/npmjs-elasticmachine'
+    NPM_PACKAGE = '@elastic/synthetics'
+    SLACK_CHANNEL = '#synthetics-user_experience-uptime'
+  }
```
v1v, line 14: https://github.com/elastic/apm-pipeline-library/blob/3102b08c8aed84818d98406d91b40cc3401619ac/resources/JenkinsfileTemplate.groovy#L166-L168

```suggestion
    DIST_TAG = "${params.DIST_TAG}"
    VERSION = "${params.VERSION}"
  }
```

.ci/dist-tag.groovy
```diff
@@ -0,0 +1,62 @@
...
+        deleteDir()
+        gitCheckout(basedir: "${BASE_DIR}")
+      }
+    }
+    stage('Dist tag') {
+      options {
+        skipDefaultCheckout()
+      }
+      steps {
+        withNodeJSEnv() {
```
v1v, line 45: Minor validation in case the parameters are empty

```suggestion
       whenFalse(env.VERSION?.trim()) {
         error("VERSION cannot be empty: ${env.VERSION}")
       }
       whenFalse(env.DIST_TAG?.trim()) {
         error("DIST_TAG cannot empty: ${env.DIST_TAG}")
       }
        withNodeJSEnv() {
```

.ci/dist-tag.groovy
```diff
@@ -0,0 +1,62 @@
...
+  agent { label 'linux && immutable' }
+  environment {
+    REPO = "synthetics"
+    BASE_DIR = "src/github.com/elastic/${env.REPO}"
+    NPMRC_SECRET = 'secret/jenkins-ci/npmjs/elasticmachine'
+    TOTP_SECRET = 'totp/code/npmjs-elasticmachine'
+    NPM_PACKAGE = '@elastic/synthetics'
+    SLACK_CHANNEL = '#synthetics-user_experience-uptime'
+  }
+  options {
```
v1v, line 15: if no concurrency is allowed then I'd say to use the below option:
- https://www.jenkins.io/doc/book/pipeline/syntax/#available-options

```suggestion
  options {
    disableConcurrentBuilds()
```

.ci/dist-tag.groovy
```diff
@@ -0,0 +1,62 @@
...
+    timeout(time: 1, unit: 'HOURS')  // to support releases then we will add a timeout in each stage
+    buildDiscarder(logRotator(numToKeepStr: '20', artifactNumToKeepStr: '20', daysToKeepStr: '30'))
+    timestamps()
+    ansiColor('xterm')
+    disableResume()
+    durabilityHint('PERFORMANCE_OPTIMIZED')
+    rateLimitBuilds(throttle: [count: 60, durationName: 'hour', userBoost: true])
```
v1v, line 22: probably this is not needed if no concurrency

```suggestion
```

.ci/dist-tag.groovy
amannocci: AFAIK, concurrency on this Jenkins job should not be a problem?

.ci/dist-tag.groovy
amannocci: I added minor validation in a separate stage a bit earlier to avoid unneeded git checkout and so on.

.ci/dist-tag.groovy
v1v: This might be good to have so env variables for parameters for the very first build can have the right value

.ci/dist-tag.groovy
```diff
@@ -0,0 +1,80 @@
...
+  post {
+    cleanup {
+      notifyBuildResult(slackComment: true)
+    }
+  }
+}
+
+def isNullOrEmpty(String value) {
+  return value == null || value.trim().isEmpty()
+}
```
v1v, line 80: You can use https://github.com/elastic/apm-pipeline-library/blob/main/vars/isEmpty.groovy

v1v (changes requested):
params to env might be needed, as pointed in https://github.com/elastic/synthetics/pull/562/files#r927506252

or alternatively to use the `params` map https://www.jenkins.io/doc/book/pipeline/syntax/#available-parameters  instead using env.VERSION or env.DIST_TAG

Issue #551, opened 2022-07-06
Discovery option for public locations


paulb-elastic:
As a user of Synthetic Project Monitors
I want to be able to determine the public locations I can use for running tests from (via the service)
So that I know what configuration to add to my scripts

ACs:
* A mechanism in Synthetics to get the locations from the command line (e.g. `npx @elastic/synthetics locations --list`)
* This will return a list of locations
  * Parsed from https://github.com/elastic/synthetics/blob/main/src/dsl/locations.ts (i.e. is based on what the current version of the Synthetics package installed, knows about)
  * It must be clear from the list, what string the user needs to include in `monitor.use.locations: []`

lucasfcosta:
Just E2E tested this using the latest released version of synthetics.

As a note, the command doesn't include the `--list`. It actually only requires `@elastic/synthetics locations`.

Running that command using `npx` to get the latest package version from the registry I see, as expected, the output below:

```
➜ npx @elastic/synthetics locations
Available locations:
   * japan
   * india
   * singapore
   * australia_east
   * united_kingdom
   * germany
   * canada_east
   * brazil
   * us_east
   * us_west

Set default location for monitors via
  - Synthetics config file 'monitors.locations' | 'monitors.privateLocations' field
  - Monitor API 'monitor.use({ locations: ["japan"], privateLocations: ["custom-location"] }')
```

PR #566, opened 2022-07-29
feat: improved project initialization settings

### LINKED ISSUE #557: [Meta] Synthetics Agent enhancements for Private Node

vigneshshanmugam:
+ fix #557 
+ fix  https://github.com/elastic/synthetics-dev/issues/130
+ Improves the synthetics project initialization experience with new sets of fields that helps streamline the experience for new users to onboard and start creating and running monitors seamlessly in Synthetics service and also on private nodes. 
+ The way it works is to ask set of questions to the users around how the users are running the Elastic stack and extracts the Kibana URL if the user is on cloud and requires API Key 
  - To fetch all location metadata from the Kibana
  - Creates a Project settings under `.synthetics/project.json`
  - Uses the scaffolded project skeleton along with `project.json` and `synthetics.config.ts` as source of truth
  - Errors on pushes with incorrect project settings
  - Allow users to manually override the values with CLI flags or via DSL.
+ If the user has already scaffolded the project using prior version of Synthetics, Installing the latest version would ask the users to `Initialize` the project again to establish the default settings. Users can follow along and manually override specific files if they want to or skip steps to onboard faster. 
+ Pushing the monitors via different project name prompts the user explicitly and follows the user input. 

### Testing.
1. Build the local project using - `npm run build`
2. Setup a brand new Synthetics project - `node dist/cli.js init [optional-dir]`
3. Follow along 
     - Elastic Cloud vs On Prem customers
     - Public locations & Private Locations
     - Kibnana Auth
4. Pushing monitors with default settings - only required arg is `auth`. 
5. Try different scenarios like pushing under different project name - `SYNTHETICS_API_KEY="" npm run push -- --project different-name`
6. Delete schedule and locations and try again to experience different project settings related errors. 
7. Go to any existing scaffolded synthetics project using prior versions of the agent and try pushing to experience different errors. For this to work properly. You would need to link to the current synthetics version. You can do so by changing `package.json` to `@elastic/synthetics: file:/path/to/locally/built/synthetics`


src/cli.ts
```diff
@@ -229,4 +233,23 @@ program
...
+  .description('List all locations to run the synthetics monitors')
+  .option('--url <url>', 'kibana URL to upload the monitors')
+  .requiredOption(
+    '--auth <auth>',
+    'API key used for Kibana authentication(https://www.elastic.co/guide/en/kibana/master/api-keys.html).'
+  )
+  .action(async (cmdOpts: LocationCmdOptions) => {
+    try {
+      console.log(await getLocations(cmdOpts));
+      // TODO: Display list of locations
```
vigneshshanmugam, line 248: This will be covered as part of - https://github.com/elastic/synthetics/issues/551

lucasfcosta (commented):
I've just followed the instructions specified both for a cloud deployment and for a local deployment.

I also tried a variation of pushing to custom locations and existing managed locations, and multiple different projects, that worked fine.

As expected, changing the project resulted in a "push aborted".

The only problem I found was that when I interactively select a private location it's not actually shown as selected after. [See this video to understand](https://user-images.githubusercontent.com/6868147/182426709-45e4e95c-d4b4-4121-8b24-c6d61652825c.mov).


Just before I re-review, I just wanna understand better what you mean with this testing step:

```
Go to any existing scaffolded project and try pushing to experience different errors. For this to work properly. You would need to link to the current synthetics version. You can do so by changing package.json to @elastic/synthetics: file:/path/to/locally/built/synthetics
```

What exactly does it mean to go into any scaffolded project and try pushing? Shouldn't the push work within the scaffold already?

If with this paragraph you mean it shouldn't work without pointing to the locally built synthetics then it's okay.

vigneshshanmugam:
> The only problem I found was that when I interactively select a private location it's not actually shown as selected after. [See this video to understand](https://user-images.githubusercontent.com/6868147/182426709-45e4e95c-d4b4-4121-8b24-c6d61652825c.mov).

Did you click `Enter` ? I think since its a multiselect, need to do it via `Space`. Hmmm this feels like we need to tweak the settings to make Enter select a single item. Can you try it again? 

>What exactly does it mean to go into any scaffolded project and try pushing? Shouldn't the push work within the scaffold already?

I was trying to say, If you already have an existing Synthetics project scaffolded from priort synthetics versions. Running `init` with this PR should overwrite all default settings and make your project ready for working with private locations.

vigneshshanmugam:
> Did you press Enter ? I think since its a multiselect, need to do it via Space. Hmmm this feels like we need to tweak the settings to make Enter select a single item. Can you try it again?

Added hints now which should inform users on how to select the locations. Also added validation to pick at-least one default location.

dominiqueclarke (changes requested):
Hi there. I noticed a few things.

Private locations configuration
---
Screen Shot 2022-08-02 at 2 47 06 PM

Default private locations are stored in `synthetics.config.ts` with the `(private)` post fix. This will not work as Kibana will not be able to find the associated location. When pushing, my monitors indeed were not assigned to the private location. 

Trailing slashes
--
Screen Shot 2022-08-02 at 2 42 41 PM
We should probably remove trailing slashes when formatting KIbana url

Cloud flow
--
We discussed using cloud id vs Kibana url to find the Kibana instance. Is that still the plan, or did we want to go with Kibana url as an MVP first?
Screen Shot 2022-08-02 at 2 39 49 PM

vigneshshanmugam:
@dominiqueclarke Fixed all the issues mentioned in the review. Please have another look.

lucasfcosta (changes requested):
I've just tested this and found one more problem.

Now, it seems I always need to specify a `--url` value although that's already defined in the `.synthetics` file of my project.

```
{
  "url": "https://my-example.us-west2.gcp.elastic-cloud.com:443",
  "space": "default",
  "project": "examplelocal2"
}
```

Here's what I see:

```
➜ SYNTHETICS_API_KEY="<REDACTED>" npm run push -- --locations "japan" --project test

> examplelocal3@1.0.0 push
> npx @elastic/synthetics push --auth $SYNTHETICS_API_KEY "--locations" "japan" "--project" "test"

error: required option '--url <url>' not specified
```

Before the most recent changes I could do this just fine. If I manually use the URL as the URL param, the push works just fine 🤔 

I also double checked the new defaults and the changes to the CLI UI and they LGTM, it's just the URL thing now.

lucasfcosta (approved):

lucasfcosta:
After re-testing with a new build it worked! My bad for not pointing the NPM script to the right place.

dominiqueclarke (approved):
LGTM. Concerns addressed.

PR #568, opened 2022-08-01
feat: abort push on duplicate monitors

### LINKED ISSUE #442: Duplicate journey id, silently groups synthetic pings under the last journey in list
If an id is repeated across journeys in a suite, in Uptime, all the pings of the journeys with the same id will go under the journey occurring last in order. Uptime will show only one monitor in monitors overview page, with monitor title as the title of the journey read last.

No warning/error is reported in logs.

Correct data and screenshots are shown under each ping though.

Example:
A journey script as:
```typescript
import { journey, step, expect } from '@elastic/synthetics';

journey({id: 'todo-journey-id', name: 'check if title is present'}, ({ page, params }) => {
    await page.goto(params.url);
  });

    await page.type('input.new-todo', 'New Title');
  });
});

journey({id: 'todo-journey-id', name: 'check if input placeholder is correct'}, ({ page, params }) => {
    await page.goto(params.url);
  });

  step('assert placeholder value', async () => {
    const input = await page.$('input.new-todo');
    expect(await input.getAttribute('placeholder')).toBe(
      'What needs to be done?'
    );
  });
});
```

Will result in Uptime showing:
image

Whereas the the list of pings contains pings of both monitors with correct data and screenshots:
image

**Heartbeat version**: `heartbeat version 8.1.0 (amd64), libbeat 8.1.0 [84924e9f32d018e360303714c26f754dbab34f93 built 2021-12-30 16:14:09 +0000 UTC]`

**Synthetics version**:  `1.0.0-beta.17`

vigneshshanmugam:
+ fix #442 
+ Aborts the push command if any of the monitors contain duplicate ids with a clean error message and also linking users to the duplicate monitor files where required. 
+ Decided to handle the logic here as Kibana currently does not have any way to know if there are any duplicates without looking up the saved objects which is not performant when we have all the list already available via the agent. 

Screen Shot 2022-08-01 at 3 30 41 PM

### Testing.
2. Setup a brand new Synthetics project - node dist/cli.js init [optional-dir]
3. Change the monitior ids for some of the monitors. 
4. Run push command - node dist/cli.js push --auth <>

### DIFF
--- a/__tests__/fixtures/duplicate.journey.ts
+++ b/__tests__/fixtures/duplicate.journey.ts
@@ -25,6 +25,8 @@
 
 import { journey, monitor } from '../../src';
 
-journey('journey 1', () => monitor.use({ id: 'duplicate' }));
+journey('journey 1', () => monitor.use({ id: 'duplicate id' }));
+journey('journey 2', () => monitor.use({ id: 'duplicate id' }));
 
-journey('journey 2', () => monitor.use({ id: 'duplicate' }));
+journey('duplicate name', () => monitor.use({ schedule: 10 }));
+journey('duplicate name', () => monitor.use({ schedule: 20 }));
--- a/__tests__/fixtures/duplicate.journey.ts
+++ b/__tests__/fixtures/duplicate.journey.ts
@@ -1,0 +1,30 @@
+/**
+ * MIT License
+ *
+ * Copyright (c) 2020-present, Elastic NV
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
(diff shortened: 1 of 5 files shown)


__tests__/fixtures/duplicate.journey.ts
```diff
@@ -0,0 +1,30 @@
...
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ *
+ */
+
+import { journey, monitor } from '../../src';
+
+journey('journey 1', () => monitor.use({ id: 'duplicate' }));
```
dominiqueclarke, line 28: Since names are used as ids in the absense of `id`, can we add two monitors with the same name to this fixture as well?

vigneshshanmugam: Oh good callout, sure will do.

dominiqueclarke (approved):
LGTM, tested four scenarios

1. Two monitors with the same id configured via `monitor.use`
2. Two monitors with the same id configure via journey options
3. Two monitors with the same name without id configured via the other two options
4. Two monitors, one with an id specified, one without an id specified, where the name of the former and the id of the latter are the same.

__tests__/fixtures/duplicate.journey.ts
vigneshshanmugam: Done - https://github.com/elastic/synthetics/pull/568/commits/4b3837a9ef3fcf1ceea8a0bfb87f6e367109d5b5

PR #571, opened 2022-08-02
feat: discovery command for locations

### LINKED ISSUE #551: Discovery option for public locations

vigneshshanmugam:
+ fix #551 
+ Support discovery option for both Public and Private locations. 
+ By default we render all default public locations, and when user provides `--auth` and `--url` options, we also render the private locations which comes in handy without doing project init setup. 

### Testing
- List Public locations via - node dist/cli.js locations
- List public and private locations - node dist/cli.js locations --auth apiKey --url kibana-url


andrewvc (changes requested):
Something's not right for private locations:

```
andrewvc@LAPTOP-BSF625CT ~/projects/synthetics (location-command) [1]
❯ node dist/cli.js locations --url=https://snap-3a06be.kb.us-central1.gcp.foundit.no:9243/ --auth=MYAUTH
TypeError: Cannot read properties of undefined (reading 'toLowerCase')
```

andrewvc:
@mergifyio update

andrewvc (approved):
LGTM, the bug I found was only an issue on main

Issue #548, opened 2022-06-30
`--params` option should be listed under `push` subcommand


andrewvc:
The push subcommand does support the `--params` option to push parameters to kibana, but it doesn't show up under `elastic-synthetics push --help`. This is confusing and should be fixed.

PR #572, opened 2022-08-03
fix: allow subcommand to receive common options

vigneshshanmugam:
+ fix https://github.com/elastic/synthetics/issues/548
+ Now we have a global common options that would be created shared between different subcommands and can be passed around as well.


andrewvc (approved):
LGTM, much nicer! Confirmed on the CLI

PR #575, opened 2022-08-09
chore: override project settings and locations

vigneshshanmugam:
This PR solves two UX issues
+ Allows the push command to work with Dev and Staging Kibana instances where the location naming scheme might be bit different. `Contitent - Location` to `Location`. 
+ Overrides synthetics project settings and configuration with updated locations when `init` command is run multiple times.

### Testing
+ Build the project `npm run build`
+ Initialize the new synthetics project - `node dist/cli.js init [dir]`
+ Run init again on the same directory and notice how Synthetics project and config gets updated without asking user override prompts.


justinkambic (approved):
LGTM!

PR #576, opened 2022-08-09
feat: support streaming mode for push command

justinkambic:
Depends on https://github.com/elastic/synthetics/pull/575.

This patch will make it possible for the Synthetics Agent to `push` monitors to Kibana via a streams endpoint.

### Configure Kibana

Firstly, build and run the [Kibana streaming PR](https://github.com/elastic/kibana/pull/138069). It is ok if it runs against a fresh local Elasticsearch cluster.

Once Kibana is running, log in as a user who can write (I simply used the `elastic` user), and create an API key, which we will supply to Synthetics for `push`ing.

Lastly, go to the Uptime application and enable monitor management. You should be prompted to create an Agent policy. Do this, following default options. Once your agent policy is created, the baseline configuration for Kibana is ready.

### Configure Synthetics Project

0. Build synthetics: `npm run build`, `npm link`
1. Create a directory (can be a subdir in synthetics) `mkdir beep`, `node dist/cli init beep`, follow the prompt
2. Enter your new project: `cd beep`, `npm link @elastic/synthetics`
3. Perform a push: `env SYNTHETICS_API_KEY={API_KEY} npm run push`

### Expected output

```
(node:92955) ExperimentalWarning: stream/web is an experimental feature. This feature could change at any time
(Use `node --trace-warnings ...` to show where the warning was created)
> preparing all monitors
> creating all monitors
> {MONITOR_NAME}: monitor updated successfully
```


src/push/index.ts
```diff
@@ -70,11 +70,16 @@ export async function push(monitors: Monitor[], options: PushOptions) {
       const { error, message } = await body.json();
       throw formatAPIError(statusCode, error, message);
     }
-    const { failedMonitors } = await body.json();
-    if (failedMonitors.length > 0) {
-      throw formatFailedMonitors(failedMonitors);
-    }
-    done('Pushed');
+    body.on('data', async data => {
```
vigneshshanmugam, line 73: Prefer doing this
```
for await (const data of body) {
  console.log('data', data)
}
```

Otherwise you would have to explicitly handle the stream related errors. `body.on('error') etc`

justinkambic: Sure, to be honest we just didn't vet this code thoroughly as the focus has been on Kibana. But we are using `for await` in our API tests so it makes sense to follow the same pattern here.

vigneshshanmugam: single try catch would work, we dont need any explicit handling if we move to async interables `for await loop.

src/helpers.ts
```diff
@@ -321,6 +321,23 @@ export function wrapFnWithLocation<A extends unknown[], R>(
...
 
+// Safely parse ND JSON (Newline delimitted JSON) chunks
+export function safeNDJSONParse(chunks: string[]) {
+  // chunks may not be at proper newline boundaries, so we make sure everything is split
+  // on proper newlines
+  chunks = Array.isArray(chunks) ? chunks : [chunks];
+  const lines = chunks.join('\n').split(/\r?\n/);
+  return lines
+    .filter(l => l.match(/\S/)) // remove blank lines
+    .map(data => {
```
justinkambic, line 332:
```suggestion
    .map(line => {
```

I think naming this `line` instead of `data` will make the code more readable.

src/push/index.ts
```diff
@@ -70,11 +70,23 @@ export async function push(monitors: Monitor[], options: PushOptions) {
...
+          progress(chunk);
+          continue;
+        }
+        const { failedMonitors } = chunk;
+        if (failedMonitors && failedMonitors.length > 0) {
+          throw formatFailedMonitors(failedMonitors);
+        }
+      }
     }
-    done('Pushed');
```
vigneshshanmugam, line 77: @justinkambic Seems like your last merge removed this line.

src/helpers.ts
vigneshshanmugam: Sure can you push the change along with the merge change https://github.com/elastic/synthetics/pull/576#discussion_r944535091

src/push/index.ts
justinkambic: oh weird, sorry it was a mistake merging. Will fix.

src/helpers.ts
justinkambic: ++

src/push/index.ts
```diff
@@ -72,6 +72,9 @@ export async function push(monitors: Monitor[], options: PushOptions) {
       throw formatAPIError(statusCode, error, message);
     }
     body.setEncoding('utf-8');
+    body.on('error', err => {
```
vigneshshanmugam, line 75: IMO, This should not be needed, when we consume the stream, any error should be propagate and should blow up inside try-catch, can you give the test case where we need this?

vigneshshanmugam: And the way we are treating this error is not the correct way. In anycase, we need a test case for this.

justinkambic: ++ - I tested this again and with this code the result is that for an error like the socket closing, we log it and then see it logged from the catch again. So yes, I think we should remove this.

My main concern here was we are providing a call to the RXJS `Subject.error` in the Kibana side. This presumably would send a message for something like a Saved Object failure, or some other strange situation. We seem not to handle that case (i.e. connection did not fail but some other error happened and was sent over the stream). Maybe this is a knowledge gap on my part as it relates to streams, but that was why I was trying this. I can't actually create an this state on purpose in my testing, so I think it's ok if we just remove this and rely on the `try...catch`.

vigneshshanmugam: So in the RX side, subject.error or subject.complete would actually end the connection and close the stream with the final bytes. In theory, it would need to call `stream.emit("close")` or `stream.emit("error")`. In both cases we would catch it as the iteration would need to consume the body and end up throwing.

justinkambic: Thanks for the explanation 👍

dominiqueclarke (approved):
This looks great to me!

Tested in 8.4.0 Kibana with streams, as well as in 8.3.0 Kibana without streams.

Tested creation, updates, and deletion.

Tested individual monitor errors (via failedMonitors), and the formatting for that is also working well.

PR #578, opened 2022-08-15
fix: add warnings for stale monitors

### LINKED ISSUE #565: [Push command] Add error reporting for failedStaleMonitors
The Kibana api returns `failedStaleMonitors` a list of monitors intended to be deleted but unable to be due to downstream errors. This list is currently dropped on the floor. 

We should consume this list and display the feedback to the user.

vigneshshanmugam:
+ fix #565 
+ Creates warnings for the stale monitors if the push succeeded without any failures. If there are errors then these warnings are dropped on the floor to make room for users to fix the errors first before checking the warnings.

### DIFF
--- a/__tests__/push/__snapshots__/request.test.ts.snap
+++ b/__tests__/push/__snapshots__/request.test.ts.snap
@@ -18,3 +18,14 @@ exports[`Push api request format failed monitors 1`] = `
 
 "
 `;
+
+exports[`Push api request format failed stale monitors 1`] = `
+"⚠ Warnings
+   > Failed to delete monitor.: monitor(monitor-1)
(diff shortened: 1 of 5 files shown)


dominiqueclarke (approved):
LGTM! 

Screen Shot 2022-08-18 at 1 35 15 PM


Btw, I noticed a fix is needed on the Kibana side, if you're able to review. https://github.com/elastic/kibana/pull/139133/files

vigneshshanmugam:
> Btw, I noticed a fix is needed on the Kibana side, if you're able to review. https://github.com/elastic/kibana/pull/139133/files

Approved yours, good catch.

PR #579, opened 2022-08-15
fix: bundled monitor content must be idempotent

### LINKED ISSUE #573: pushed monitor base64 content is not consistent
+ When we bundle and push the monitor, the base 64 content of the journey is different every time we push a new journey to Kibana. This ignores the Kibana Optimisation which works based on the monitor source content being unchanged.

vigneshshanmugam:
+ fix #573 
+ With this change, we effectively removed the date property which was set by the zip stream to always refer to the Unix epoch time instead of defaulting to current time in UTC which was causing the zipped contents to be changed everytime when we run push command. 

### Testing
1. Build the files, `npm run build`
2. Initialize a project - `node dist/cli.js init dir`
3. cd into `dir` and change the synthetics package to link to the current one. `@elastic/synthetics: file:../` in package.json file.
4. Install the synthetics package to link - `npm install`
5. Run the push command - `npm run push`

### DIFF
--- a/__tests__/push/bundler.test.ts
+++ b/__tests__/push/bundler.test.ts
@@ -30,7 +30,7 @@ import { join } from 'path';
 import { generateTempPath } from '../../src/helpers';
 import { Bundler } from '../../src/push/bundler';
 
-async function validateZip(content, file) {
+async function validateZip(content) {
   const decoded = Buffer.from(content, 'base64');
   const pathToZip = generateTempPath();
   await writeFile(pathToZip, decoded);
@@ -70,7 +70,7 @@ journey('journey 1', () => {
 
   it('build journey', async () => {
     const content = await bundler.build(journeyFile, generateTempPath());
-    await validateZip(content, journeyFile);
+    await validateZip(content);
   });
 
   it('bundle should be idempotent', async () => {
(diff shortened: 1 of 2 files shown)


dominiqueclarke (changes requested):
Getting a JSON error that doesn't appear in main when pushing to Kibana on main in a directory generated by init. 

Screen Shot 2022-08-18 at 1 23 12 PM

vigneshshanmugam:
I just rebased on top of main to get the streaming changes. It could be related to that. Can you check it again?

lucasfcosta (approved):
@dominiqueclarke I think that might be because you're using a version of synthetics which doesn't parse the individual lines of the Kibana output after we changed it to do streaming.

I'll approve this as the changes LGTM. Also, pushes from this branch are okay on my machine, but may be good to wait for @dominiqueclarke to verify on her side as well.

dominiqueclarke:
> I just rebased on top of main to get the streaming changes. It could be related to that. Can you check it again?

I think it was related to this, will let you know once I get another chance to test.

dominiqueclarke (approved):
LGTM with main merged! 🚢

PR #580, opened 2022-08-17
Allow users to bundle node_modules

lucasfcosta:
> Please notice that I had to rebase this PR on top of #576 so that we can actually parse streaming responses from Kibana.

# Summary

This PR allows users to bundle modules into the scripts they push.

I've marked it as draft as it still needs E2E tests.

# How to test this PR

To test this PR, try including a JS library into your journey, by importing it into the journey file.

Then push your journey with the extra imported module to Kibana.

You should be able to run the Journey just fine.


src/push/plugin.ts
```diff
@@ -88,13 +88,14 @@ export function MultiAssetPlugin(callback: PluginCallback): esbuild.Plugin {
...
           path.extname(args.importer);
+
         // Spin off another build to copy over the imported modules without bundling
         const result = await esbuild.build({
           ...commonOptions(),
           entryPoints: {
             [entryPath]: entryPath,
           },
-          bundle: false,
+          bundle: true,
```
vigneshshanmugam, line 98: this would change our journey files format IIRC.

lucasfcosta: That's interesting, do you mean the file format itself (the extension)? Or do you mean the content format (i.e. bundling the CommonJS infra)?

In my tests, I added an `fs.writeFile` to the code so I could save the generated zips to my disk. Then, once I opened them, I did manage to see that the file extension was `.js` and that the content was also valid JavaScript.

I also checked the docs here: https://esbuild.github.io/api/#bundle and couldn't see a reference for format changes.

Can you go into a bit more detail about your concern here? Perhaps I'm missing something or misunderstood.

vigneshshanmugam: I mean the content of the journey itself. Lets say when you push a journey file `file.journey.ts`, What does the content look like? Did it get changed to something else and whether the files were bundled - concatenated in to a single file for the files(utils, helpers) that were imported by the journey?

lucasfcosta:
CC @vigneshshanmugam feel free to merge this whenever you're happy with it, all tests are now passing 😄

vigneshshanmugam (changes requested):
Even though this solution works, this defeats the purpose of our Plugin and kind of bundles everything together in a single file. 

The goal of our push command is to replicate exactly how your folder structure is on your local and run the same inside HB when pushed. This would make it presentable in Kibana when we add unzipping support and allow users to view and edit monitors. 

TLDR: If i push the `advanced.journey.ts` to Kibana, the structure should like
```
// Main branch
/example/todos/advanced.journey.ts
/example/todos/helper.ts


// PR branch
/example/todos/advanced.journey.ts
```

lucasfcosta:
@vigneshshanmugam can you double check whether you had the last version during your test? That behaviour you mentioned was happening before the changes I did yesterday morning.

Right now, I just tried to push a journey which depends on another local file and it did separate the necessary files.

To test that, I added the following line to `src/push/bundler.js` in the `build` function to get the ZIP file:

```ts
  async build(entry: string, output: string) {
    await this.prepare(entry);
    await this.zip(output);
    const data = await this.encode(output);
    await writeFile('/tmp/new-synth.zip', data, { encoding: 'base64' });
    await this.cleanup(output);
    return data;
  }
```

Then I extracted the `new-synth` zip file and got two separate files:

Screenshot 2022-08-24 at 13 19 45

Is this what you meant? Or did I misunderstand?

paulb-elastic:
As discussed in Tech Sync, we should ensure this feature includes a check and handling the failure of the compressed bundle being too big for either the Saved Object, or configmap to support, and feed this back to the user (e.g. _...failed to create the monitor as the bundle size is bigger than the maximum 800KB..._)
cc @lucasfcosta

lucasfcosta:
@vigneshshanmugam I was looking into not bundling the `node_modules` but in that case we'd have to implement the resolution ourselves as the plugin we use bundles them by default and it doesn't seem like we can change that.

Considering that Kibana only shows the code _inside_ the step, as shown in the screenshot below, what do you think of us merging this one as is?

Screenshot 2022-08-31 at 10 35 28

CHANGELOG.md
```diff
@@ -5,15 +5,79 @@ All notable changes to this project will be documented in this file.
 **For detailed release notes, please refer to the [GitHub
 releases](https://github.com/elastic/synthetics/releases) page.**
 
+## v1.0.0-beta.32 (2022-08-24)
```
vigneshshanmugam, line 8: can you rebase @lucasfcosta ?

lucasfcosta: Done!

src/push/plugin.ts
```diff
@@ -61,21 +68,43 @@ export function MultiAssetPlugin(callback: PluginCallback): esbuild.Plugin {
...
+  // refer to the package itself.
+  const isLocalSynthetics = (entryPath: string) => {
+    return entryPath.startsWith(sourceSyntheticsPath);
+  };
+
+  // When importing the local synthetics module directly
+  // it may import its own local dependencies, so we must
+  // make sure those will be resolved using Node's resolution
+  // algorithm, as they're still "node_modules" that we must bundle
+  const isLocalSyntheticsModule = (str: string) => {
```
vigneshshanmugam, line 84: Why not do str.includes`@elastic/synthetics`, it would work with source node modules and synthetics itself? We would get absolute path all the time, so this should be good enough?

lucasfcosta: I might have misunderstood your comment, but yes we do get the absolute path (when importing the `.js`, not always). That's the exact reason we can't use the `includes` you suggested because when importing the local file the alias will not be the name we're passing in.

Is this what you meant? I thought I might have misunderstood because even if what you suggest happens always that means we can never use the suggested `includes`.

StuGray86:
Hi guys.

We're currently waiting on this change, is there any timeframe for its release?

vigneshshanmugam (approved):
LGTM 🎉

awahab07:
**Post FF Testing**

I was able to import a simple library like `uuid` and successfully create, push and run the journey.

Script I used:        
```typescript
import { journey, step, monitor } from '@elastic/synthetics';

import { v4 as uuidv4 } from 'uuid';

journey('Journey uuid Test', ({ page, params }) => {
  monitor.use({
    id: 'uuid-lib-test-001',
    enabled: true,
    schedule: 20,
  });

  step('Generate UUID', async () => {
    const uuid = uuidv4();
    await page.goto('https://onlinetexttools.com/convert-text-to-image');
    const textAreaLocator = page.locator('textarea').first();
    await textAreaLocator.fill(uuid);
    await textAreaLocator.type(' ');
  });
});

```
image

However, I did find an issue with bundling certain imports and raised a [bug](https://github.com/elastic/synthetics/issues/619).

PR #586, opened 2022-09-01
fix: push monitor requests individually

dominiqueclarke:
Pushes project monitor requests individually, following up with a larger request to account for stale monitors.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,15 +5,79 @@ All notable changes to this project will be documented in this file.
 **For detailed release notes, please refer to the [GitHub
 releases](https://github.com/elastic/synthetics/releases) page.**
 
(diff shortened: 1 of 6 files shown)


lucasfcosta (approved):
Tested this manually both on private locations and against the service.

I tested this a few ways: one with 50 monitors against a local Kibana using my local service — all monitors synced successfully and the output on the console looks OK.

Once all were pushed I waited for syncs and they seemed okay: all 50 monitors synced and none disappearing after the cleanup interval.

I setup a private node and pushed the 50 monitors to that private node, and the behaviour looked okay too.

For service pushes, I couldn't see the hyper slow times too as mine remained ~30s for 50 monitors (both on creates and updates). I only saw slow responses on private locations.

vigneshshanmugam (approved):
LGTM

PR #587, opened 2022-09-01
chore - remove zip url tests for e2e suite

dominiqueclarke:
Removes zip url tests for the elastic-agent e2e suite

### DIFF
--- a/__tests__/e2e/synthetics.journey.ts
+++ b/__tests__/e2e/synthetics.journey.ts
@@ -172,31 +172,6 @@ journey(`${stackVersion} e2e test synthetics - tcp`, async ({ page }) => {
 //   });
 // });
 
-journey(`${stackVersion} e2e test synthetics - browser`, async ({ page }) => {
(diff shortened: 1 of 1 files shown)


vigneshshanmugam (approved):
LGTM, you would need to remove this from here to not even run it on the first place. 

Heartbeat config - https://github.com/elastic/synthetics/blob/e99cd80b9e8d3872fc8bc8c0ea997217320bcc12/examples/docker/heartbeat.docker.yml#L20

dominiqueclarke:
> LGTM, you would need to remove this from here to not even run it on the first place.
> 
> Heartbeat config -
> 
> https://github.com/elastic/synthetics/blob/e99cd80b9e8d3872fc8bc8c0ea997217320bcc12/examples/docker/heartbeat.docker.yml#L20

This docker config isn't actually used for this set of tests, but for the original synthetics-agent tests Shahzad wrote a while back. We're actually in the process of auditing and documenting the differences between all these tests.

PR #588, opened 2022-09-06
feat: support SYNTHETICS_API_KEY env variable for auth 

shahzad31:
read auth from `SYNTHETICS_API_KEY` if available

### DIFF
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -146,20 +146,20 @@ program
     }
   });
 
-const requiredAuthOption = new Option(
-  '--auth <auth>',
-  'API key used for Kibana authentication(https://www.elastic.co/guide/en/kibana/master/api-keys.html).'
-).env('SYNTHETICS_API_KEY');
-
-requiredAuthOption.mandatory = true;
-
 // Push command
 program
   .command('push')
   .description(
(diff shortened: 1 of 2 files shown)


src/cli.ts
```diff
@@ -146,16 +146,20 @@ program
...
 program
   .command('push')
   .description(
     'Push all journeys in the current directory to create monitors within the Kibana monitor management UI'
   )
-  .requiredOption(
-    '--auth <auth>',
-    'API key used for Kibana authentication(https://www.elastic.co/guide/en/kibana/master/api-keys.html).'
-  )
+  .addOption(requiredAuthOption)
```
vigneshshanmugam, line 162: can you change to this instead? 

```
  .addOption(
    new Option(
      '--auth <auth>',
      'API key used for Kibana authentication(https://www.elastic.co/guide/en/kibana/master/api-keys.html).'
    )
      .env('SYNTHETICS_API_KEY')
      .makeOptionMandatory(true)
  )
```

vigneshshanmugam (changes requested):
Also, if we are doing this, then need to change the project generation command a bit differently. 

We can remove this - https://github.com/elastic/synthetics/blob/5d4ccc0c895c35a1bf7d74a927836109b03b46d0/src/generator/index.ts#L264

PR #592, opened 2022-09-09
fix: save project settings in synthetics config

### LINKED ISSUE #590: Error when pushing monitors from the CI environment
+ Currently we rely on `project.json` internally during the initialization phase to be able to reuse the configuration when pushing monitors later once the synthetics project is set up. This works good on local dev machines, But wont work on the CI as we need the configurations like `url, project and space` to be able to push to Kibana.

Push should be idempotent when running from Dev machine and also on CI.

vigneshshanmugam:
+ fix #590 
+ Changed the way how the Synthetics agent works with project monitor settings and how we can keep the settings idempotent when users run push on their local dev machine and also on the CI.  Sparked out of discussion here - https://elastic.slack.com/archives/C022SU2U6DC/p1662556348346669
+ As a result, Introduced the project setting in the synthetics.config file instead of having yet another project.json file
There was other way, we could have solved this using the same project.json  and constructing them on demand and still be able to provide the same idempotentcy. But having them on the config file, has tons of advantages right now and also in the future.
  - Users can immediately see how the values like id, space and url  values are filled in when they initialize the agent, instead of hidden under .synthetics folder.
  - Helps with code reviews, when one user change a given value, others can validate it before getting pushed from CI
   - This is the best one so far, it will allow us to support multiple projects in the long term with few lines of code and can be explicit about it.
```
    projects: [
      { id: 'uptime', tags: ['synthetics:ui'],space: 'synthetics',url: 'uptime-synthetics.kibana.com'},
      {id: 'apm',tags: ['apm:ui'],space: 'apm',url: 'apm-synthetics.kibana.com'},
    ]
```

### Testing

#### Existing Synthetics project
- We used `project.json` before where configuration were stored, Its removed from this PR, so you should see the CLI throwing error if you try to push from the repo.
- Perform init again to reinitialize the project OR follow along the errors and set up required values in the config file.
- Go to any existing scaffolded synthetics project using prior versions of the agent and try pushing to experience different errors. For this to work properly. You would need to link to the current synthetics version. You can do so by changing package.json to @elastic/synthetics: file:/path/to/locally/built/synthetics

#### Brand new synthetics project
- Setup a brand new Synthetics project - node dist/cli.js init [optional-dir]
- Follow along and check if project settings have been initialized properly
- Validate by pushing with auth specified `npm run push -- --auth <key>`.

### DIFF
--- a/__tests__/push/index.test.ts
+++ b/__tests__/push/index.test.ts
@@ -66,9 +66,11 @@ describe('Push', () => {
     expect(output).toContain(`required option '--auth <auth>' not specified`);
   });
 
-  it.skip('error when project is not setup', async () => {
+  it('error when project is not setup', async () => {
     const output = await runPush();
-    expect(output).toContain('Aborted. Synthetics project not set up');
+    expect(output).toContain(
+      'Aborted (missing synthetics config file), Project not set up corrrectly.'
+    );
   });
 
   it('error on empty project id', async () => {
--- a/__tests__/push/index.test.ts
+++ b/__tests__/push/index.test.ts
@@ -45,15 +45,11 @@ describe('Push', () => {
   }
 
   async function fakeProjectSetup(settings, monitor) {
-    await mkdir(join(PROJECT_DIR, '.synthetics'), { recursive: true });
-    await writeFile(
-      join(PROJECT_DIR, '.synthetics', 'project.json'),
-      JSON.stringify(settings, null, 2)
-    );
-
     await writeFile(
       join(PROJECT_DIR, 'synthetics.config.ts'),
-      `export default { monitor: ${JSON.stringify(monitor)} }`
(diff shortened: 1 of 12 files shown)


__tests__/push/index.test.ts
```diff
@@ -70,7 +66,7 @@ describe('Push', () => {
     expect(output).toContain(`required option '--auth <auth>' not specified`);
   });
 
-  it('error when project is not setup', async () => {
+  it.skip('error when project is not setup', async () => {
```
vigneshshanmugam, line 69: Figuring out a better way to test this yet.

vigneshshanmugam: Done.

andrewvc (approved):
LGTM

PR #593, opened 2022-09-09
feat: support pushing lightweight monitors

### LINKED ISSUE #542: Support for lightweight monitors via projects
With project monitors we now let users manage browser based monitors easily from the CLI, but that story does not extend to lightweight monitors. This limits the utility of our overall solution since users need multiple workflows, when they ideally could use project monitors for everything if desired.

This issue proposes supporting the YAML format used in heartbeat's `monitor.d` directory with project monitors. These would act the same as browser based monitors, only their definitions would be YAML based and included in the source tree. An example is listed below.

```
- type: http 
  id: my-http-monitor
  name: Lightweight HTTP Monitor
  schedule: '@every 5s' 
  hosts: ["https://www.elastic.co"]
  # New keys for service listed below
  locations: ["us-west", "japan"]
```

## Why YAML?

- Users will need to migrate existing configs, this being the exact same format they already use makes migration a breeze. 
- Lightweight monitors are more frequently used by administrators who are more comfortable with YAML than JS/TS and Node. Asking users with those personas to use JS/TS may hinder adoption.
- Ease of implementation, should be a simple translation to the API

## Gotchas / Known issues

- The IDs are not exactly portable since project monitors include the space and journey in the monitor ID. We could add a special config key `preserve_exact_monitor_id: true|false` to handle this situation.
- We won't be able to support cron syntax, or, if we do, we should support the subset that is translatable to `@every` syntax (the regex `(\*(\/\d)? )(\*(\/\d)? )\* \* \*` should match and capture cron jobs meant to run every N minutes or every N hours.
- Running locally is something we'll need to determine in the future. I propose that we initially just validate these and allow them to be pushed. Later we can determine a strategy for running these locally. Perhaps the simplest answer would be to just invoke `docker run` in `run_once` mode. For now, we could just validate the syntax of these monitors locally and mark them in some sort of `skipped` status.

vigneshshanmugam:
+ fix #542 
+ Adds support for pushing lightweight monitors via `.(yml|yaml)` files pattern to Kibana endpoint when invoked using `npx @elastic/synthetics push` command. 
+ Works by checking all the yml files inside the synthetics project and creates a Monitor object that matches with browser monitors to be able to push to the Kibana endpoint. Also performs duplication and error reporting as we extract the line number and other information to go back to the source location. 
+ Supports only `@every <1m>` syntax for scheduling lightweight monitors, Throws Error when it sees other CRON format for schedules. 
+ Uses location information from two places
  - Default locations from Synthetics.config.ts when project is initialized using `npx @elastic/synthetics init` command
  - Allows overriding for individual monitors via `locations` and `private_locations` fields in Heartbeat yml files.
+ Added couple of validations 
    - Ignores pushing `browser` and `disabled` monitors
    - Aborts if monitor `id | name` is not provided. 

### Testing
- Build the local project using - `npm run build`
- Run `npm link`
- Setup a brand new Synthetics project - node dist/cli.js init [optional-dir]
- Navigate to the new directory created with `init` and run `npm link @elastic/synthetics`
- Follow along and check if project settings have been initialized properly
- Create Heartbeat yaml files, copy over existing ones or use the below (some additional heartbeat yaml is included in the linked Kibana PR)
```yml
- type: http
  enabled: true
  id: my-monitor
  name: My Monitor
  urls: ["http://localhost:9200"]
  schedule: @every 10m
  locations: ["singapore"]
  private_locations: ["private-1"]
  timeout: 16s
  check.request:
    method: POST
    headers:
      'Content-Type': 'application/x-www-form-urlencoded'
  check.response:
    status: [200]
    body:
      - Saved
      - saved
```
- Validate by pushing with auth specified `SYNTHETICS_API_KEY=<key> npm run push`.


andrewvc:
How does this let users specify location in the YAML? There's a weird overlap with the new `location` field monitors have in https://github.com/elastic/beats/pull/30632 . Thoughts?

vigneshshanmugam:
> It's possible to push a monitor without an id, and it's sent to Kibana as undefined.

I have seen Kibana reporting `ID is required` in the response for browser monitors when attempting to save browser monitors. Is it not valid still?

shahzad31:
it should throw an early validation error if two monitors with same id are defined in light weight monitors.

shahzad31 (changes requested):
User can defined monitors with same id

shahzad31:
throws error if empty heartbeat.monitors is defined

```
➜  test-project git:(lightweight-suport) ✗ node ../dist/cli.js push
TypeError: Cannot read properties of undefined (reading 'items')
    at createLightweightMonitors (/Users/shahzad-16/elastic/synthetics/src/push/monitor.ts:112:38)
    at async Command.<anonymous> (/Users/shahzad-16/elastic/synthetics/src/cli.ts:207:25)
```

vigneshshanmugam:
> User can defined monitors with same id

This was already handled correctly, see the test case here - https://github.com/elastic/synthetics/blob/20f98f29ec456317c10bf13969c8238c2300541c/__tests__/push/index.test.ts#L152

> throws error if empty heartbeat.monitors is defined

Fixed it, test case - https://github.com/elastic/synthetics/blob/20f98f29ec456317c10bf13969c8238c2300541c/__tests__/push/monitor.test.ts#L148

dominiqueclarke (approved):
LGTM

PR #595, opened 2022-09-14
feat: support cross-linking synthetics & apm

### LINKED ISSUE #265: [Proposal] Crosslinking Synthetics with APM
# Crosslinking Synthetics with APM

Provide visibility into how the synthetic journeys are executed and what actions are happening inside every step.

### Linking via trace headers (`traceparent`)

The previous proposal of linking between the Synthetics/Heartbeat is to generate a unique `Traceid` for each run and propagate the trace context details to the other APM agents which could pick up the id and generate the transactions/spans respectively.

How this would work from the Synthetics agent,

```ts
journey('Synthetics + APM', ({ page }) => {
  step('propagate traceparent ', async () => {
    await page.route('**', (route, request) => {
      if (request.isNavigationRequest()) {
       // add trace headers for all requests.
        const headers = {
          traceparent:
            '00-dc739d61ca520efa486e48aae80a66c7-4de1afd949fe4b1d-01',
          ...request.headers(),
        };
        route.continue({ headers });
      }
      route.continue();
    });
    await page.goto('http://localhost:8080/index');
  });
});

```

**Note: Backend server running on port `8080` is running Node.js agent and injects RUM agent for the index page.**

The above code propagates the trace headers `traceparent` to all the backend servers which would help identify the overall trace from serving the index page to what requests went inside that page since we use RUM and also Node.js agent. 

Screen Shot 2021-04-26 at 16 43 49

### Synthetics as an APM agent

Instead of injecting the trace details for every run, Synthetics could work as an APM agent and start creating transactions for both journeys and steps which could be propagated to the other APM agents  including Node.js and RUM. This would let the user visualize the overall picture from how the browser got launched to how each actions in journey and step impacted the whole test. 

I have patched both Synthetic and RUM agent to make it work locally, See an example of how this would look like 

Screen Shot 2021-04-28 at 16 57 54

Think about how we could leverage the Network details in Synthetics with APM and RUM.

Both approaches will result in browser ignoring the DISK/HTTP cache as they modify the headers before the request is being made to the server. We might need to dig more and figure out how to do this without affecting cache. However as its a test mode, it might not be a big of a deal. 

Would like to thank @spalger from Kibana Operations team for brainstorming.  

/cc @graphaelli @andrewvc @paulb-elastic

vigneshshanmugam:
+ fix #265 
+ Relies on the properties being injected from the Heartbeat side to make it work seamless. 
+ Adds support for crosslinking the Synthetics Waterfall information with the APM data to be able to understand the performance monitoring data across Elastic observability solutions. The key fields that are added to make the transition easier are
   - `synthetics.monitor.id` - monitor id of the synthetics monitor
   - `syntheitcs.monitor.type` -  type of the synthetics monitor, can be `browser or http`
   - `synthetics.trace.id` - checkgroup/exec id that denotes execution of a journey at any point
  - `synthetics.monitor.location` - location id where the monitor was run from
  - `tracestate: es=s:1` - Sampling decision is propagated via tracestate header which is understood by APM agents

+ Adds the relevant headers only for the routes/urls that are configured via the `apm.traceUrls`. Sampling decision can also be propagated via `apm.sampleRate`. These options can be modified via CLI or Synthetics config file. 

```ts
journey("test", () => {
   step("trace check", async() => {
      await page.goto("https://example.com")
   });
});

// config 
{
  "apm": { "traceUrls":  ["**/*"] } // add to all requests
}

// Headers
baggage: `synthetics.trace.id=<test>;synthetics.monitor.id=<test>;`
tracestate: `es=s:1`

```

+ Env variables will be added by the Heartbeat to make this complete - https://github.com/elastic/beats/issues/36249


src/plugins/apm.ts
```diff
@@ -0,0 +1,58 @@
...
+  constructor(private driver: Driver, private options: ApmOptions) {}
+
+  async traceHandler(route: Route, request: Request) {
+    const traceId = randomBytes(16).toString('hex');
+    // We dont want to track Synthetics, so sending a dummy id
+    const parentId = '0'.repeat(16);
+    const traceparent = `00-${traceId}-${parentId}-01`;
+    const headers = {
+      ...(await request.allHeaders()),
+      traceparent,
```
sqren, line 40: Can you send the `monitor.id` and `execution.id` in the `tracestate` header? Then I can talk to the APM agent devs to persist this field on APM docs. 

This will make it possible to to retrieve Synthetic monitors for a particular APM service.

Something like this:
```suggestion
      traceparent,
      tracestate: `synthetics.monitor.id=1234,synthetics.execution.id=5678`
```

See more about trace-state here: https://www.elastic.co/guide/en/apm/agent/rum-js/current/distributed-tracing-guide.html#enable-tracestate

gregkalapos: Just adding it here as well: there is an ongoing discussion whether to use `tracestate` or the `baggage`  header instead. `baggage` seems to be a better fit. Discussion is linked internally.

sqren: Thanks for chiming in @gregkalapos. I've updated my suggestion to use `baggage` instead of `tracestate`

sqren: One more change: the synthetics agent should only send `baggage` header - not the `traceparent` header. The reason is that if the synthetics client start a trace (by generating a trace id) but doesn't send a transaction (like the RUM agent would) the resulting trace will be "root-less because it does not have an entry transaction.

By not sending `traceparent` the receiving service will instead create the `trace.id` and will also send up the entry transaction.

syepes:
+1

shahzad31:
Superseded by #1162, which rebases this work onto current `main` (on top of the multi step api runner refactor #997) now that the Heartbeat side (elastic/beats#51083) has merged. Closing this in favor of #1162 — thanks @vigneshshanmugam for the original implementation.

PR #598, opened 2022-09-15
feat: delete entire project via push command

### LINKED ISSUE #581: add support to delete the entire project with push
+ This would be fairly trivial to implement without adding in any extra options like `--delete` or `--delete project`. Instead when users tries to run push in a empty project, we could prompt users with the message `Pushing empty journeys might nuke the journeys associated with the project id`, If user agrees, we could proceed with pushing the empty monitors and Kibana would nuke the entire project.

vigneshshanmugam:
+ fix #581 
+ users would be promoted with a dialog when pushing with empty monitors which then clicking `Yes` would nuke the entire project in Kibana. 
+ Also fixed a bug where we kept pushing even on auth errors, now we immediately exit when we find errors during pushing.

### DIFF
--- a/src/push/index.ts
+++ b/src/push/index.ts
@@ -25,7 +25,7 @@
 
 import { readFile, writeFile } from 'fs/promises';
 import { prompt } from 'enquirer';
-import { bold, yellow } from 'kleur/colors';
+import { bold } from 'kleur/colors';
 import {
   ok,
   formatAPIError,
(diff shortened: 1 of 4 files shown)


src/push/index.ts
```diff
@@ -49,24 +49,42 @@ import type { PushOptions, ProjectSettings } from '../common_types';
...
+    write('');
+    // Makes testing easier with overrides
+    let deleteAll = false;
+    if (process.env.TEST_OVERRIDE != null) {
+      deleteAll = Boolean(process.env.TEST_OVERRIDE);
+    } else {
+      ({ deleteAll } = await prompt<{ deleteAll: boolean }>({
+        type: 'confirm',
+        name: 'deleteAll',
+        message: `Pushing without any monitors would delete all monitors associated with the project.\n Do you want to continue?`,
```
andrewvc, line 76:
```suggestion
        message: `Pushing without any monitors will delete all monitors associated with the project.\n Do you want to continue?`,
```

We should be more definitive with phrasing here

src/push/index.ts
```diff
@@ -49,24 +49,42 @@ import type { PushOptions, ProjectSettings } from '../common_types';
...
-  }
+    progress(`creating all monitors`);
+    for (const schema of schemas) {
+      await pushMonitors({ schemas: [schema], keepStale: true, options });
+    }
+  } else {
+    write('');
+    // Makes testing easier with overrides
+    let deleteAll = false;
+    if (process.env.TEST_OVERRIDE != null) {
```
andrewvc, line 70: I think we should add a `--yes` flag to skip this in case someone wants to automate it for any reasons.

I did try running with `yes | npx @elastic/synthetics push` but our command really dislikes that and tries to process all the input thus stalling.

andrewvc (changes requested):
We really need a `--yes` flag here I think

src/push/index.ts
vigneshshanmugam: I will address this separately and open a PR.

andrewvc (approved):
LGTM, I'm OK with `--yes` coming in a future PR

PR #602, opened 2022-09-19
feat: support --yes flag to skip and force push

### LINKED ISSUE #601: add support for --yes flag for project monitors
+ Its good to bypass interactive commands in the CI when pushing project monitors. This gets applied to
   - When project id is different from existing project id, suppress the warning
   - When deleting the entire project 

Sparked from - https://github.com/elastic/synthetics/pull/598#discussion_r972440734

vigneshshanmugam:
+ fix #601 
+ `npx @elastic/synthetics push --auth <key> -y` would skip the questionnare and forcefully push the project monitors during two phases
   - When deleting the entire project
   - When project is pushed with different id than pushed before.

### DIFF
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -184,6 +184,7 @@ program
     '--space <space>',
     'the target Kibana spaces for the pushed monitors — spaces help you organise pushed monitors.'
   )
+  .option('-y, --yes', 'skip all questionares and do forceful push')
   .addOption(pattern)
   .addOption(tags)
   .addOption(match)
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -184,7 +184,7 @@ program
     '--space <space>',
     'the target Kibana spaces for the pushed monitors — spaces help you organise pushed monitors.'
   )
-  .option('-y, --yes', 'skip all questionnaires and run non-interactively')
+  .option('-y, --yes', 'skip all questions and run non-interactively')
   .addOption(pattern)
   .addOption(tags)
   .addOption(match)
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -184,7 +184,7 @@ program
     '--space <space>',
     'the target Kibana spaces for the pushed monitors — spaces help you organise pushed monitors.'
   )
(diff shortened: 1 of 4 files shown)


src/cli.ts
```diff
@@ -184,6 +184,7 @@ program
     '--space <space>',
     'the target Kibana spaces for the pushed monitors — spaces help you organise pushed monitors.'
   )
+  .option('-y, --yes', 'skip all questionares and do forceful push')
```
andrewvc, line 187:
```suggestion
  .option('-y, --yes', 'respond with the default option to all prompts, useful for CI and other automated environments')
```

I think this definition might be more future-proof, thoughts?

src/push/index.ts
```diff
@@ -65,19 +65,19 @@ export async function push(monitors: Monitor[], options: PushOptions) {
...
-    } else {
-      ({ deleteAll } = await prompt<{ deleteAll: boolean }>({
-        type: 'confirm',
-        name: 'deleteAll',
-        message: `Pushing without any monitors will delete all monitors associated with the project.\n Do you want to continue?`,
-        initial: false,
-      }));
-    }
-
+    const { deleteAll } = await prompt<{ deleteAll: boolean }>({
```
andrewvc, line 68: so much cleaner!

andrewvc (approved):
I left one suggestion, I could take or leave it, but LGTM regardless

src/cli.ts
vigneshshanmugam: Hmm, we dont default to all options like `projectid, url, space` so might need a different wording. Do you have any other suggestion?

vigneshshanmugam: Reworded a bit.

src/cli.ts
```diff
@@ -184,6 +184,7 @@ program
     '--space <space>',
     'the target Kibana spaces for the pushed monitors — spaces help you organise pushed monitors.'
   )
+  .option('-y, --yes', 'skip all questionnaires and run non-interactively')
```
vigneshshanmugam, line 187:
```suggestion
  .option('-y, --yes', 'skip all questions and run non-interactively')
```

PR #607, opened 2022-09-20
chore: upgrade deps and fix test types

vigneshshanmugam:
+ As part of the rotation schedule, upgrade all outdated deps to the nearest Minor/Patch versions. 
+ Fixed type issues that was not caught in tests.


andrewvc (approved):
LGTM

PR #608, opened 2022-09-20
chore: check build directory instead of root

vigneshshanmugam:
+ Currently for dev purposes, if we end up creating the Synthetics project as part of the Synthetics agent directory, it would result in locally imported helper files marked as `external` which when run on HB would result in this error. 
```
{
    "code": "BAD_CMD_STATUS",
    "stack_trace": null,
    "type": "io",
    "message": "journey did not finish executing, 0 steps ran: command '/tmp/elastic-synthetics-unzip-1725738249/node_modules/.bin/elastic-synthetics /tmp/elastic-synthetics-unzip-1725738249/node_modules/.bin/elastic-synthetics /tmp/elastic-synthetics-unzip-1725738249 --playwright-options {\"headless\":true,\"ignoreHTTPSErrors\":true} --screenshots on --throttling {\"download\":5,\"latency\":20,\"upload\":3} --rich-events --match addition and completion of single task been 2 4 --params \"{1 hidden params}\"' exited unexpectedly with code: 1"
}
```

This PR solves the above issue by skipping bundler (marked as external) only for
+ Source files of Synthetics package - `/@elastic/synthetics/dist`
+ Node modules that are part of the package - `/@elastic/synthetics/node_modules`

### DIFF
--- a/src/push/plugin.ts
+++ b/src/push/plugin.ts
@@ -27,8 +27,12 @@ import { isAbsolute, dirname, extname, join } from 'path';
 import fs from 'fs/promises';
 import * as esbuild from 'esbuild';
 
-const SOURCE_DIR = join(__dirname, '..', '..');
-const SOURCE_NODE_MODULES = join(SOURCE_DIR, 'node_modules');
+// ROOT directory of the Package - /
+const ROOT_DIR = join(__dirname, '..', '..');
+// Source of the package - /src, /dist, etc.
+const SOURCE_DIR = join(__dirname, '..');
+// Node modules directory of the package - /node_modules
+const SOURCE_NODE_MODULES = join(ROOT_DIR, 'node_modules');
 
 export function commonOptions(): esbuild.BuildOptions {
   return {


lucasfcosta (commented):
The changes themselves LGTM, but when I tested E2E, something very weird happened, which might actually be a problem with the initial repo structure. Here's what I did:

1. I created a new directory with `init`
2. I updated my `Bundler` with the lines below so I can get the zip file out of the bundling process
    ```
    await writeFile('/tmp/test.zip', data, { encoding: 'base64' });
    ```
3. I built the package
4. I went into the directory I created using `init` and ran a `push`
5. ⚠️ The problem I've found was that only the `example.journey.ts` was in the `zip` file. The advanced journey was not there despite being in the folder.

vigneshshanmugam:
>  The problem I've found was that only the example.journey.ts was in the zip file. The advanced journey was not there despite being in the folder.

I mean this could be because your zip was replaced with the last run journey? Could you write it to different id instead of using same filename?

lucasfcosta (approved):
Hey Vignesh, just double checked this and a couple of other things and that's correct. Because I was using the same filename it seems that the zip files were overwriting each other.

This time I:

* Checked that the hash of my tree is indeed `074b2670b7acee7d0c09e5ae2269044e741b38ab`, as is this commit's
* I updated the `zip` saving so that it adds the current timestamp to the generated file name (and rebuilt)
* I created a brand new directory and retried, now I get three zips (and the timestamp matches the current time)
* Each zip has a different journey with correct contents

Good shout on the new testing strategy!

PR #612, opened 2022-09-26
chore: use default schedule and warn for lightweight monitors

vigneshshanmugam:
+ Uses the default schedule from `Synthetics.config.ts` if the schedule is not defined in `heartbeat.yml` files.
+ Warns users about the resolution if the schedule is `< 60 seconds`.

### DIFF
--- a/src/push/monitor.ts
+++ b/src/push/monitor.ts
@@ -34,6 +34,7 @@ import {
   SYNTHETICS_PATH,
   totalist,
   indent,
+  warn,
 } from '../helpers';
 import { LocationsMap } from '../locations/public-locations';
 import { Monitor, MonitorConfig } from '../dsl/monitor';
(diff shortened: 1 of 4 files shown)


src/push/monitor.ts
```diff
@@ -94,6 +95,12 @@ export async function createLightweightMonitors(
       lwFiles.add(abs);
     }
   });
+  // Warn users about schedule that are less than 60 seconds
+  if (lwFiles.size > 0) {
```
dominiqueclarke, line 99: This shows always, even when a user doesn't have a 30s interval. Is that desired?

dominiqueclarke (approved):
LGTM. Only one comment.

src/push/monitor.ts
vigneshshanmugam: Yeah its hard to do it per monitor and also bubbling up that error to the overall monitor creation logic. Kept it here as its simpler and also serves as a warning.

PR #615, opened 2022-09-27
fix: include header and body for transferSize value

vigneshshanmugam:
+ part of #616 
+ Include the transferred header + body size when calculating the transferSize value for a given network request. 

**NOTE: No explicit testing is necessary as there is a dedicated test validating the transferred sizes of the response.**


lucasfcosta (approved):
The code LGTM. I also E2E tested with `--rich-events` and checked the fields on the network JSON and it LGTM.

PR #634, opened 2022-10-24
Update all NPM deps, and bump playwright to 1.27.1

andrewvc:
Just keeping things current


package.json
```diff
@@ -51,8 +51,8 @@
     "http-proxy": "^1.18.1",
     "kleur": "^4.1.5",
     "micromatch": "^4.0.5",
-    "playwright-chromium": "=1.26.0",
-    "playwright-core": "=1.26.0",
+    "playwright-chromium": "=1.27.1",
```
vigneshshanmugam, line 54: PW updates should not trigger those big changes in lock files. Something is going on. Did you happen to run npm audit fix or anything similar?

andrewvc: I also ran `npm update` . Is that a problem? We should be running that regularly, no?

vigneshshanmugam (approved):
LGTM

PR #639, opened 2022-10-31
Update github workflow 

andrewvc:
per https://github.com/elastic/synthetics-demo/pull/31 and https://github.com/elastic/synthetics-demo/pull/27

### DIFF
--- a/templates/.github/workflows/run-synthetics.yml
+++ b/templates/.github/workflows/run-synthetics.yml
@@ -8,19 +8,20 @@ on:
     - cron: '* 0 * * *'
 jobs:
   test:
+    defaults:
+      run:
+        working-directory: ./advanced-examples
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v2
-      - uses: actions/setup-node@v2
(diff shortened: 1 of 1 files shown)


templates/.github/workflows/run-synthetics.yml
```diff
@@ -8,19 +8,20 @@ on:
...
+        working-directory: ./advanced-examples
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v2
-      - uses: actions/setup-node@v2
+      - uses: actions/checkout@v3
+      - uses: actions/setup-node@v3
         with:
-          node-version: 16
+          node-version: 18
```
vigneshshanmugam, line 19: Synthetics uses node 16 by default, it would be better to use the same here.

andrewvc: I'm thinking we should update to 18 soon anyway, and it does seem to work. Could I put that into this PR as well? I wouldn't mind updating the heartbeat images etc. too to help keep us on the current LTS version.

Issue #662, opened 2022-11-28
Runner doesn't follow playwrightoptions timeout value


shahzad31:
Synthetics runner browser context always use default timeout of 30 seconds, it doesn't follow the value specified in playwrightoptions.

vigneshshanmugam:
Removing the bug label as it's not appropriate. 

Plus this needs to be solved in a different way given we diverse from the playwright timeout options - https://github.com/elastic/synthetics/issues/561#issuecomment-1215610563

Issue #577, opened 2022-08-15
Introduce global timeout for the runner


vigneshshanmugam:
+ Add support for the complete test run of the Synthetics runner which includes running multiple journeys to completion and takes in to account of other individual timeouts. This option would be configurable via
   - CLI  via `--globalTimeout  30000`
   - Synthetics config file (synthetics.config.ts) - `{ globalTimeout: 30000 }`

For individual journey timeouts, users would still rely on #133

Issue #561, opened 2022-07-21
[Synthetics] Add timeout option to push monitor config


dominiqueclarke:
Browser monitor timeouts are intended to be supported in 8.4.

A timeout option should be added to the monitor dsl to define timeouts on the monitor level.

paulb-elastic:
For reference, linking to the Heartbeat implementation https://github.com/elastic/beats/issues/32388

@dominiqueclarke does this supersede https://github.com/elastic/synthetics/issues/133?

paulb-elastic:
> Not really, This is applicable to the overall global timeout for a single Synthetics run. It encompasses `Journey` and `Step` level timeouts.

@vigneshshanmugam but isn't that what https://github.com/elastic/synthetics/issues/133 is _asking for_, being able to set timeouts at the step and/or UJ level?

vigneshshanmugam:
It's kind of related, but it differs the way we would support on the Synthetics agent. There are four different types of timeouts that comes in to place in the context of synthetics

1. Global runner timeout - Time it takes for the whole synthetics run to be completed, can include multiple journeys and obviously multiple steps, actions. 
2. Journey level timeout - Overall timeout for a single execution of a journey/monitor. Can include multiple steps and actions. 
3. Step level timeout - Timeout for a particular step inside a journey. Multiple action timeouts can be controlled together using this. 
4. Action timeout - Timeout for a particular playwright option. 

### CLI timeout

For the timeout that we want to support considering the whole Heartbeat and Monitors, the CLI option `--timeout` would be the timeout for the overall synthetics run which is the global runner timeout. For HB, it's going to be a single journey, but under the hood we can run any number of journeys and still be controlled by the `--timeout` flag. 

The issue https://github.com/elastic/synthetics/issues/133 is more about the second level timeouts which is out of context for this particular feature.

vigneshshanmugam:
Adding some details here as per the conversation with @paulb-elastic and our path for the timeout changes across synthetics. 

1. Users would be able to override individual Monitor timeout aka Journey timeout using the Monitor DSL or via CLI option `--timeout` and Synthetics config file. 
```ts
monitor.use({ timeout: 10000}) // 10 seconds timeout

npx @elasic/synthetics journeys --timeout 10000
```

[...]

PR #663, opened 2022-11-28
fix: use timeout specified in playwright options

shahzad31:
We needed this in kibana tests where often timout can increae from 30 seconds, so changing value in each selector is annoying 

fixes  https://github.com/elastic/synthetics/issues/662

### DIFF
--- a/examples/todos/basic.journey.ts
+++ b/examples/todos/basic.journey.ts
@@ -2,7 +2,7 @@ import { journey, step, expect } from '@elastic/synthetics';
 
 journey('check if title is present', ({ page, params }) => {
   step('launch app', async () => {
-    await page.goto(params.url);
+    await page.goto('https://www.elastic.co/');
   });
 
   step('assert title', async () => {
--- a/examples/todos/basic.journey.ts
+++ b/examples/todos/basic.journey.ts
@@ -2,7 +2,7 @@ import { journey, step, expect } from '@elastic/synthetics';
 
 journey('check if title is present', ({ page, params }) => {
   step('launch app', async () => {
-    await page.goto('https://www.elastic.co/');
+    await page.goto(params.url);
   });
 
   step('assert title', async () => {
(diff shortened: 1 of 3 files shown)


examples/todos/basic.journey.ts
```diff
@@ -2,7 +2,7 @@ import { journey, step, expect } from '@elastic/synthetics';
 
 journey('check if title is present', ({ page, params }) => {
   step('launch app', async () => {
-    await page.goto(params.url);
+    await page.goto('https://www.elastic.co/');
```
shahzad31, line 5:
```suggestion
    await page.goto(params.url);
```

emilioalvap:
IMO, this change could be confusing considering PlaywightOptions type is documenting a different effect, :
```js
 /**
   * Maximum time in milliseconds to wait for the browser instance to start. Defaults to `30000` (30 seconds). Pass `0` to
   * disable timeout.
   */
  timeout?: number;
```

Can we adapt the type object to provide the correct description?

vigneshshanmugam (changes requested):
Timeout needs to be handled differently. 

Have a read through this issue - [#577 ](https://github.com/elastic/synthetics/issues/561#issuecomment-1215610563)

shahzad31:
@vigneshshanmugam isn't it possible to do also global timeout for all selectors. Can you please tell me what would you like me to change for this pr. Instead of using playwrightOptions, adding a new cli param `timeout` will that be ok with you?

vigneshshanmugam:
Yeah there can be different behavior's for timeout as stated here on the issue - https://github.com/elastic/synthetics/issues/561#issuecomment-1215610563, But if you want to have a monitor with a specified timeout, then go for `--timeout` option. 

But if you are concerned around timeout for PW actions. We cannot do that properly instead users can already do that via
```
journey("test", ({context, browser}) => {
  context.setDefaultTimeout(10000);
})

```

shahzad31:
Closing this for now to have more discussions around it

PR #665, opened 2022-11-28
fix: parse buffered NDJSON response from kibana correctly

### LINKED ISSUE #660: Ndjson parsing is broken if data is bit large
It will throw an error if data is returned like this since it's not able to safely parse the json

`Error SyntaxError: Unexpected end of JSON input could not parse data '{"createdMonitors":[],"updatedMonitors":["production-logging-ap-northeast-1"],"staleMonitors":["production-metrics-gcp-asia-east1","production-metrics-eu-west-1","production-metrics-azure-westus2","production-metrics-azure-westeurope","production-metrics-azure-uksouth","production-metrics-azure-southeastasia","production-metrics-azure-southcentralus","production-metrics-azure-southafricanorth","production-metrics-azure-northeurope","production-metrics-azure-japaneast","production-metrics-azure-francecentral","production-metrics-azure-eastus2","production-metrics-azure-eastus","production-me'`

issue can be reproduced on cloud against 8.5.1 or 8.5.1 and on the latest version of the agent !!

vigneshshanmugam:
+ fix #660 
+ When Kibana NDJSON responses are buffered, this would result in parsing error on the chunking code as buffer length can be different when working on local vs on cloud (proxies).
+ To account for these varying buffer lengths, We consume all the stream body from Kibana as buffers and then parse the NDJSON response to avoid intereleaved parsing errors.

### DIFF
--- a/__tests__/push/index.test.ts
+++ b/__tests__/push/index.test.ts
@@ -263,6 +263,12 @@ heartbeat.monitors:
           req.on('data', chunks => {
             const schema = JSON.parse(chunks.toString()) as APISchema;
             res.write(JSON.stringify(schema.monitors[0].name) + '\n');
+            if (!schema.keep_stale) {
+              // write more than the stream buffer to check the broken NDJSON data
+              res.write(
+                JSON.stringify(Buffer.from('a'.repeat(70000)).toString()) + '\n'
+              );
+            }
           });
           req.on('end', () => {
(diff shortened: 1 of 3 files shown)


shahzad31:
Tested against cloud cluster and it works as expected

image

shahzad31 (approved):

PR #666, opened 2022-11-29
feat: create monitors in chunks of 10 using push command

shahzad31:
Since new API features are going to be used only for 8.6+, and while testing on cloud, i have noticed current 1 by 1 is way too slow for users. 

Increasing it by reasonable chunks of 10 , which should provide us good experience and caution.


__tests__/push/__snapshots__/index.test.ts.snap
```diff
@@ -4,7 +4,6 @@ exports[`Push API handle streamed response 1`] = `
 "> preparing all monitors
 > creating all monitors
 > journey 1
-> journey 2
```
vigneshshanmugam, line 7: any reason why we are not getting updates of journey 2?

shahzad31: this is just a test unit behaviour , since api is called only once, in the test first response is logged.

shahzad31:
@vigneshshanmugam  this test is very flaky, is there a way to improve it?

image

__tests__/push/__snapshots__/index.test.ts.snap
vigneshshanmugam: Can you change tests to account for multiple response chunks from the streaming endpoint. Thanks.

shahzad31: actually in case of multiple jounrey it still logs one response, but yeah the text can be improved , let me do that

vigneshshanmugam:
@shahzad31 Yeah noticed that too in separate issue, Lets me debug that in a new PR. Seems to start with new PW versions.

__tests__/push/__snapshots__/index.test.ts.snap
shahzad31: updated the test

.eslintignore
```diff
@@ -3,3 +3,4 @@ examples
 templates
 __tests__/e2e/tmp/
 src/sdk/lh-trace-processor.ts
+test-projects/
```
vigneshshanmugam, line 6: nit: I would suggest you to add these test-projects to global .gitignore and remove it from project specific settings as they are not related.

__tests__/push/index.test.ts
```diff
@@ -262,7 +262,12 @@ heartbeat.monitors:
           await new Promise(r => setTimeout(r, 20));
           req.on('data', chunks => {
             const schema = JSON.parse(chunks.toString()) as APISchema;
-            res.write(JSON.stringify(schema.monitors[0].name) + '\n');
+            res.write(
+              JSON.stringify(
+                schema.monitors.length + ' monitors created successfully'
+              ) + '\n'
+            );
+            res.write(JSON.stringify('1 monitor updated successfully') + '\n');
```
vigneshshanmugam, line 270: better to write something specific to monitor related details other than static string.

src/push/index.ts
```diff
@@ -61,8 +61,11 @@ export async function push(monitors: Monitor[], options: PushOptions) {
     schemas = await buildMonitorSchema(monitors);
 
     progress(`creating all monitors`);
-    for (const schema of schemas) {
-      await pushMonitors({ schemas: [schema], keepStale: true, options });
+
+    const chunks = getArrayChunks(schemas, 10);
+
```
vigneshshanmugam, line 66: nit: remove blank lines.

src/push/index.ts
```diff
@@ -252,3 +255,11 @@ export async function catchIncorrectSettings(
     await overrideSettings(settings.id, options.id);
   }
 }
+
+function getArrayChunks<T>(arr: T[], chunkSize: number): T[][] {
```
vigneshshanmugam, line 259: would move this to helpers.ts

vigneshshanmugam (approved):
LGTM with pending comments addressed.

PR #667, opened 2022-11-29
fix: working lightweight monitors with init

vigneshshanmugam:
+ fix the buggy example - https://github.com/elastic/synthetics/pull/650


shahzad31:
after building the branch tested it using init and new example file was present in the newly created project

image

shahzad31 (approved):

PR #673, opened 2022-12-07
feat: bundle package deps as local packages

### LINKED ISSUE #596: create separate bundles for external modules
+ With the support of bundling external packages, we bundle the included packages along with the source journey files to be able to run them via Heartbeat later without installing any packages. This is good, but it changes the source line numbers and column numbers of those particular journeys and might be a bad UX if we have to go back to the original source. 

More details in here - https://github.com/elastic/synthetics/pull/580#pullrequestreview-1082955330

The ultimate goal of our push command is to replicate exactly how your folder structure is on your local and run the same inside HB when pushed. This would make it presentable in Kibana when we add unzipping support and allow users to view and edit monitors.

TLDR: If i push below example journey code 
~~examples/bundled-code/test.journey.ts~~
```ts
import { journey, step, monitor, expect } from "@elastic/synthetics";
import isPositive from "is-positive";
import leftpad from "leftpad";
import * as utils from "./utils";

journey("example journey", ({ page, params }) => {
  step("example step", () => {});
});
```

Our bundled zip folder should be in the below format. 

```
examples/bundled-code/test.journey.ts
examples/bundled-code/utils.ts
examples/bundled-code/is-positive.js
examples/bundled-code/leftpad.js
```

### LINKED ISSUE #652: journey did not finish executing, 0 steps ran when calling exported function.
Hi there!

I'm trying to structure my synthetic tests project by extracting methods (API requests mostly) to dedicated TS files. The problem is that I often hit the issue where my tests run and pass locally (with `npm test`), but the version I push to ELK's Uptime app (our ELK cloud is running 8.4.3) does not run at all. The monitor appears in a `Failed` state, and when I go to the details, I see 

> There are no steps for this journey
This journey did not contain any steps.
The journey's check group is 3388a32c-642b-11ed-9db9-ce3a36e90cb7.
There is no further information to display.

Under the history section, in the error column, I can see the following:

```
journey did not finish executing, 0 steps ran: command '/tmp/elastic-synthetics-unzip-20303004/node_modules/.bin/elastic-synthetics /tmp/elastic-synthetics-unzip-20303004/node_modules/.bin/elastic-synthetics /tmp/elastic-synthetics-unzip-20303004 --playwright-options {"headless":true,"ignoreHTTPSErrors":false} --screenshots off --throttling {"download":5,"latency":20,"upload":3} --rich-events --match My Example Journey --params "{1 hidden params}"' exited unexpectedly with code: 1
```

I manage to create a primer with exactly what seems to cause this failure, using the [base template](https://github.com/elastic/synthetics/tree/main/templates), with the following files in the `journeys` directory:

```
example.journey.ts
foo.ts
get-foo.ts
```

With the following content:

```typescript
// foo.ts
export class Foo {
    private readonly bar: string;
}


// get-foo.ts
import { Foo } from './foo';

export function createFoo(): Foo {
    return new Foo();
}


// example.journey.ts
import { expect, journey, monitor, step } from '@elastic/synthetics';
import { createFoo } from './get-foo';

journey('Journey foo no params', ({ page, params }) => {
    // Only relevant for the push command to create
    // monitors in Kibana
    monitor.use({
        id: 'example-monitor',
        schedule: 10,
        screenshot: 'off',
    });

    step('getFoo()', async () => {
        const foo = createFoo();
        expect(foo).toBeDefined();
    });
});
```

Do you have any guidance on how to interpret `journey did not finish executing`?

Edit: removed unimportant `import` statement to reproduce the issue.

vigneshshanmugam:
+ fix #652
+ fix #596

### DIFF
--- a/__tests__/push/__snapshots__/bundler.test.ts.snap
+++ b/__tests__/push/__snapshots__/bundler.test.ts.snap
@@ -1,0 +1,38 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`Bundler build journey 1`] = `
+"
+__tests__/push/test-bundler/bundle.journey.ts
+import { journey, step, monitor } from \\"@elastic/synthetics\\";
+import isPositive from \\"./bundles/is-positive/index\\";
+import utils from \\"./utils\\";
+journey(\\"journey 1\\", () => {
+  utils();
+  monitor.use({ id: \\"duplicate id\\" });
+  launchStep(-1);
+});
+const launchStep = (no) => {
+  step(\\"step1\\", () => {
+    isPositive(no);
+  });
+};
+
+__tests__/push/test-bundler/bundles/is-positive/index.js
+\\"use strict\\";
+module.exports = function(n) {
+  return toString.call(n) === \\"[object Number]\\" && n > 0;
+};
+
+__tests__/push/test-bundler/utils.ts
+import isNegative from \\"./bundles/is-negative/index\\";
+export default utils = () => {
+  isNegative(1);
+};
+
+__tests__/push/test-bundler/bundles/is-negative/index.js
+\\"use strict\\";
+module.exports = function(n) {
+  return toString.call(n) === \\"[object Number]\\" && n < 0;
+};
+"
+`;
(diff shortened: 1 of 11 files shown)


vigneshshanmugam:
Closing this in favor of #687

PR #675, opened 2022-12-07
add url and api key options to init command

shahzad31:
This allow it easy to bootstrap project by providing opts and api key via options. 

This will allow us to show command in kibana for user, where they are getting api key like this

`npx @elastic/synthetics init --url <url> --api-key <apiKey>`

### DIFF
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -49,6 +49,7 @@ import {
   getLocations,
   renderLocations,
   LocationCmdOptions,
+  InitCmdOptions,
 } from './locations';
 import { resolve } from 'path';
 import { Generator } from './generator';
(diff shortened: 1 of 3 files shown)


vigneshshanmugam (commented):
I dont feel good about this, `Init` command is a one time setup and requires inputs from users. Having fewer options about `url, api-key` and leaving out spaces/project settings is not a great UX. 

Scaffolding projects should get you in clean slate and `init` command should never take any inputs as options, instead should walk user through an interactive way. 

I have used hundreds of scaffolding projects, every project follows the same pattern, aiming for the clean slate and guiding the users instead of taking options.

paulb-elastic:
@vigneshshanmugam @shahzad31 is this still being done, or should it be closed?

shahzad31:
Closing for now

PR #683, opened 2022-12-14
fix: update hash when monitor content changes

vigneshshanmugam:
+ Make sure the code changes for a given monitor updates the hash which would update the Monitors being pushed to Kibana instead of skipping them.
+ With this PR, Hash is changed to be based on three factors:
    - Monitor configuration 
    - Code changes (Any change in the code would need to be factored in as we will be using that showcase the stack trace)
    - File path changes (No need to check for line and column as they are irrelevant)

### DIFF
--- a/__tests__/push/monitor.test.ts
+++ b/__tests__/push/monitor.test.ts
@@ -78,24 +78,22 @@ describe('Monitors', () => {
       schedule: 10,
       type: 'http',
       enabled: true,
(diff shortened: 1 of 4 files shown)


dominiqueclarke (approved):
LGTM! Marks monitor as updated and pushes it when source is modified, global or local config is modified, or file path is changed.

PR #688, opened 2022-12-28
feat: prompt before overriding config file

### LINKED ISSUE #680: Prompt for overwriting files on existing scaffolded projects
+ Running init multiple times might be necessary when we add new features like adding lightweight monitors or upgrading experience if we want to support new locations and so on.
+ The agent would allow users to override the files, but prompts on overwrite. This gets the clean state, plus also keeps the push command from not breaking if the user tries to run it after scaffolding a existing project.

vigneshshanmugam:
+ fix #680 
+ This prompts the user before overwriting the existing `synthetics.config.file`. 
+ For the `package.json` - It will not be overwritten when the init is run multiple times
    - `npm init -y` process will skip the package.json installation if there is already a `package.json` file. 
    - Our patching logic will only modify the existing file if the `test` or `push` script is not defined. Safe operation.
+ For `.gitignore` - We overwrite only if the `node_modules` or `.synthetics` is not present - Safe operation.

### DIFF
--- a/src/generator/index.ts
+++ b/src/generator/index.ts
@@ -64,9 +64,6 @@ export const REGULAR_FILES_PATH = [
 ];
 export const CONFIG_PATH = 'synthetics.config.ts';
 
-// Files to be overriden by default if the project is initialized multiple times
-const DEFAULT_OVERRIDES = [CONFIG_PATH];
-
 export class Generator {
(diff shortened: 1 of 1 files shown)


andrewvc (approved):
LGTM

PR #694, opened 2023-01-11
fix: properly handle non error exceptions

### LINKED ISSUE #692: Thrown exceptions causes steps to fail, but not the whole journey
If, to a journey, I try to fail a step with the code:

`throw("I ded")`

In the kibana UI I'll see a state where the step shows as failed, but the overall journey / check shows as passed.

If I use a playwright assertion, such as

`expect("I ded").toBe("ded")`

Both will fail as expected

andrewvc:
Catch non `Error` objects that are thrown as exceptions. As a quick workaround devs sometimes throw using syntax like `throw("foo")` instead of the more correct `throw(new Error("foo"))`. 

The current code will swallow those errors.

Fixes #692 

## Testing

This PR includes a simple unit test. Given that the impact of #692 is low, and writing a true E2E test demonstrating the resolution of that issue, I think manual testing of that scenario is fine.

To do so, save the following file as `bad.ts`

```ts
//bad.ts

step("instant exception", () => {
  throw("simple exception");
})
```

Then, run 

```sh
npm run build && node dist/cli.js --inline --rich-events < bad.ts | grep journey/end |  jq .error`
```

after checking out this branch. You should see this error. Switching back to `main` the error should be missing.


src/helpers.ts
```diff
@@ -233,9 +233,16 @@ export function rewriteErrorStack(stack: string, indexes: [number, number]) {
   return stack;
 }
 
-export function formatError(error: Error) {
+// formatError prefers receiving proper Errors, but since at runtime
+// non Error exceptions can be thrown, it tolerates though. The
+// redundant type Error | any expresses that.
+export function formatError(error: Error | any) {
   if (!(error instanceof Error)) {
```
vigneshshanmugam, line 240: either we need to handle undefined objects here or https://github.com/elastic/synthetics/blob/41e6c2acc73baadb3582d46388ffbd5a835b4f0f/src/reporters/json.ts#L558

Otherwise even undefined objects get these errors reported which is incorrect.

src/helpers.ts
```diff
@@ -237,6 +237,10 @@ export function rewriteErrorStack(stack: string, indexes: [number, number]) {
 // non Error exceptions can be thrown, it tolerates though. The
 // redundant type Error | any expresses that.
 export function formatError(error: Error | any) {
+  if (error === undefined || error === null) {
```
vigneshshanmugam, line 240: error== null covers both. Or we could just do !error. I don't think users would be throwing false 🤷🏽‍♂️

src/helpers.ts
andrewvc: I think the clause above handles that right, or am I missing something?

vigneshshanmugam (approved):
LGTM

PR #696, opened 2023-01-24
fix: add monitor heading to lightweight scaffolding template 

paulb-elastic:
The `lightweight.yml` is not being pushed because it’s missing the `heartbeat.monitors:` heading


vigneshshanmugam (approved):
Good catch Paul, not sure how we missed it. LGTM

shahzad31:
POST FF Testing LGTM !!

PR #699, opened 2023-02-08
chore: remove params from journey/start event payload 

shahzad31:
We are pushing params values to journey start document payload. It serves no purpose. So in this PR i am removing those.


vigneshshanmugam (changes requested):
It was intentional and it requires a UI modifications to take full advantage. See the issue here - https://github.com/elastic/synthetics-dev/issues/62

__tests__/cli.test.ts
```diff
@@ -152,13 +152,9 @@ describe('CLI', () => {
         return getJourneyStart(cli.output());
       };
 
-      expect((await output()).payload).toMatchObject({
-        params: { url: 'non-dev' },
-      });
+      expect((await output()).payload).toMatchObject({});
```
vigneshshanmugam, line 155: this defaults the purpose of this test itself. We want to know whether the passed parameters from CLI are being used in the journey, but this change effectively discards that check itself. 

I would change the test to use a custom reporter and look for params being used as it is.

vigneshshanmugam (changes requested):
Except for the CLI tests, rest looks good

__tests__/cli.test.ts
shahzad31: i have tried to update the tests but ended up doing lot more than desired, i hope it;s ok :D

vigneshshanmugam: Fixed the tests in a different way, other than using params.

vigneshshanmugam (approved):
LGTM

PR #701, opened 2023-02-09
feat: added buildkite reporter

shahzad31:
Added build-kite friendly test reporter

Tested changes within kibana and it seems to be working as expected

### DIFF
--- a/__tests__/helper.test.ts
+++ b/__tests__/helper.test.ts
@@ -34,6 +34,7 @@ import {
   findPWLogsIndexes,
   microSecsToSeconds,
   wrapFnWithLocation,
+  replaceFileNameWithJourneyName,
   isMatch,
 } from '../src/helpers';
 
@@ -175,4 +176,13 @@ it('match tags and names', () => {
   // match both name and tags
   expect(isMatch(['bar'], 'foo', undefined, 'ba*')).toBe(true);
   expect(isMatch(['bar'], 'foo', undefined, 'test*')).toBe(false);
+});
+
+it('replace default path fileName with journey name', () => {
+  const newPath = replaceFileNameWithJourneyName({
+    filePath: '.journeys/videos/34567898765678237jnsdjkhfhjdsbf.webm',
+    journeyName: 'test-journey',
+  });
+
+  expect(newPath).toBe('.journeys/videos/test-journey.webm');
 });
--- a/src/core/runner.ts
+++ b/src/core/runner.ts
@@ -34,6 +34,7 @@ import {
   getTimestamp,
   runParallel,
   generateUniqueId,
+  replaceFileNameWithJourneyName,
 } from '../helpers';
 import {
   HooksCallback,
@@ -54,6 +55,7 @@ import {
 import { Gatherer } from './gatherer';
 import { log } from './logger';
 import { Monitor, MonitorConfig } from '../dsl/monitor';
+import fs from 'fs';
 
 type HookType = 'beforeAll' | 'afterAll';
 export type SuiteHooks = Record<HookType, Array<HooksCallback>>;
@@ -152,12 +154,13 @@ export default class Runner {
      * Set up the corresponding reporter and fallback
      * to default reporter if not provided
      */
-    const { reporter, outfd } = options;
+    const { reporter, outfd, playwrightOptions } = options;
+    const { recordVideo } = playwrightOptions ?? {};
     const Reporter =
       typeof reporter === 'function'
         ? reporter
         : reporters[reporter] || reporters['default'];
-    this.#reporter = new Reporter({ fd: outfd });
+    this.#reporter = new Reporter({ fd: outfd, recordVideo });
   }
 
   async runBeforeAllHook(args: HooksArgs) {
@@ -291,7 +294,7 @@ export default class Runner {
       params,
     });
     /**
-     * Exeucute the journey callback which would
+     * Execute the journey callback which would
      * register the steps for the current journey
      */
     journey.callback({ ...context.driver, params });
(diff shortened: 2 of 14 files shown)


__tests__/helper.test.ts
```diff
@@ -176,3 +177,13 @@ it('match tags and names', () => {
   expect(isMatch(['bar'], 'foo', undefined, 'ba*')).toBe(true);
   expect(isMatch(['bar'], 'foo', undefined, 'test*')).toBe(false);
 });
+
+it('replace default path fileName with journal name', () => {
```
devcorpio, line 181: nitpick: journey instead of journal

src/reporters/build_kite.ts
```diff
@@ -0,0 +1,183 @@
...
+    }
+
+    message += succeeded > 0 ? green(` ${succeeded} passed`) : '';
+    message += failed > 0 ? red(` ${failed} failed`) : '';
+    message += skipped > 0 ? cyan(` ${skipped} skipped`) : '';
+    message += ` (${renderDuration(now() / 1000)} seconds) \n`;
+    this.write(message);
+  }
+
+  write(message: any) {
```
vigneshshanmugam, line 177: Not sure why you are overriding this, we should let the base reporter handle this.

vigneshshanmugam (changes requested):
Before even we add this new enhanced repoter, This is by no means `buildkite` reporter. We need to adhere to the buildkite test spec if we have to build one for buildkite especially following an example reporter/collector here - https://github.com/buildkite/test-collector-javascript

Check a reference reporter for junit which follows Junit specification - https://github.com/elastic/synthetics/blob/d7c83d1609adae2dbda533eda31ed2cad29f9577/src/reporters/junit.ts#L37

The reporter added in this PR is just a enhanced CLI reporter which can be built without touching the synthetics agent. Few things that would make this easier to build `Artifact API - which can expose screenshot and other test output information without having to traverse the default `CACHE_PATH`. 

Give the above points, Lets close this PR and work on exposing the artifacts and build this custom reporter in the project where we need this capability instead of baking in to the Synthetics agent.

src/helpers.ts
```diff
@@ -427,3 +427,23 @@ export function tagsMatch(tags, pattern) {
   const matchess = micromatch(tags || ['*'], pattern);
   return matchess.length > 0;
 }
+
+export function replaceFileNameWithJourneyName({
```
vigneshshanmugam, line 431: Why do we do this? Is this for to avoid duplicating the videos between each runs?

src/reporters/build_kite_cli.ts
```diff
@@ -0,0 +1,187 @@
...
+    if (outputDir) {
+      const FAILED_STEPS_PATH = join(outputDir, 'failed_steps');
+      await fs.promises.mkdir(FAILED_STEPS_PATH, { recursive: true });
+
+      await writeScreenshotsToPath(screenshots, outputDir, status);
+    }
+  }
+
+  override onEnd() {
+    const failedJourneys = Array.from(this.journeys.entries()).filter(
```
vigneshshanmugam, line 130: Not sure if this is what you want, I see Step errors getting printed multiple times. Single journey with a step that fails. 
Screenshot 2023-09-01 at 4 33 12 PM

src/reporters/build_kite_cli.ts
```diff
@@ -0,0 +1,187 @@
...
+        try {
+          fs.unlinkSync(filePath);
+        } catch (e) {
+          // eslint-disable-next-line no-console
+          console.log('Failed to delete video file for path ' + filePath);
+        }
+      });
+    }
+
+    const { failed, succeeded, skipped } = this.metrics;
```
vigneshshanmugam, line 170: Better to move this inside the base reporter and print a summaru/epilogue of the tests and you can invoke the base reporter since all reporters extend it.

src/core/runner.ts
```diff
@@ -320,6 +323,28 @@ export default class Runner {
     });
     // clear screenshots cache after each journey
     await rm(Runner.screenshotPath, { recursive: true, force: true });
+
+    const pages = driver.context.pages();
+    const videoDir = options.playwrightOptions?.recordVideo?.dir;
```
vigneshshanmugam, line 328: We should not record video based on this, We should hide the internals of PW and let our runner take care of the video option with a value like `on|off|only-on-failure` like screenshots and ask users to pass that flag. Based on that, the runner would store the rest on the provided `output-dir` if present and write those video files once the test is completed and based on the video flag. 

By this way, you would just give the `path` to the reporter and the reporter can just use that path and share it to the buildkite CLI where the artifacts are present. 

Screenshots would be the same as well.

vigneshshanmugam (commented):
In the interest of moving this quicket and in chunks instead of these core changes which touches on the Screenshot and Video functionality, Can you split this PR

- Just the CLI output for the Buildkite reporter
- We will move the Artificats/Screenshots/Video as a separate PR gathering more ideas instead of getting blocked.

src/reporters/build_kite_cli.ts
vigneshshanmugam: Right now we repeat a lot of the same stuff from base reporter.

src/helpers.ts
shahzad31: because default videonames generated by playwright are random uuid strings, having a journeyname is helpful while viewing artifacts

src/reporters/build_kite_cli.ts
shahzad31: so the idea here is to print all the failed journeys at the end. 
if there are too many journeys, that's usually helpful.

vigneshshanmugam:
Closing in favor of https://github.com/elastic/synthetics/issues/878 and bare minimal reporter https://github.com/elastic/synthetics/pull/872

PR #706, opened 2023-02-22
enable use of custom data-test-id

shahzad31:
With this we can provide a custom data-test id like `data-test-subj`, will make life easier with kibana tests specifically.

### DIFF
--- a/__tests__/core/gatherer.test.ts
+++ b/__tests__/core/gatherer.test.ts
@@ -148,7 +148,7 @@ describe('Gatherer', () => {
         expect(request.headers()['user-agent']).toContain('Elastic/Synthetics');
       });
       await page.setContent(
-        '<a target=_blank rel=noopener href="/popup.html">popup</a>'
+        '<a target=_blank rel=noopener href="/username-page">popup</a>'
       );
       const [page1] = await Promise.all([
         context.waitForEvent('page'),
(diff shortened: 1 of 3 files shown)


__tests__/core/gatherer.test.ts
```diff
@@ -148,7 +148,7 @@ describe('Gatherer', () => {
         expect(request.headers()['user-agent']).toContain('Elastic/Synthetics');
       });
       await page.setContent(
-        '<a target=_blank rel=noopener href="/popup.html">popup</a>'
+        '<a target=_blank rel=noopener href="/username-page">popup</a>'
```
shahzad31, line 151:
```suggestion
        '<a target=_blank rel=noopener href="/popup.html">popup</a>'
```

vigneshshanmugam (commented):
@shahzad31 Whats the motivation behind this change? I have couple of questions on this

1. why cant Kibana set these attributes on `beforeAll` hook? 
2. Also, since you are doing this on the overall journey, How would it help in the Kibana context since the whole journey could contain multiple elements with these test attributes?

shahzad31:
> @shahzad31 Whats the motivation behind this change? I have couple of questions on this
> 
> 1. why cant Kibana set these attributes on `beforeAll` hook?
> 2. Also, since you are doing this on the overall journey, How would it help in the Kibana context since the whole journey could contain multiple elements with these test attributes?

setting it as  `beforeAll` will not work for inline scripts.  Plus we will have to import selectors from playwright-chromium directly. Seems hack to me going outside agent package.

`Also, since you are doing this on the overall journey, How would it help in the Kibana context since the whole journey could contain multiple elements with these test attributes?
`
can you please explain this question?

shahzad31:
> @shahzad31 Whats the motivation behind this change? I have couple of questions on this

basic motivation is ability to use `page.getByTestId` selector.

__tests__/core/gatherer.test.ts
```diff
@@ -164,6 +164,41 @@ describe('Gatherer', () => {
     });
   });
 
+  describe.only('Set Test ID Attribute', () => {
```
vigneshshanmugam, line 167: can you remove .only?

vigneshshanmugam (changes requested):

__tests__/core/gatherer.test.ts
shahzad31, line 167:
```suggestion
  describe('Set Test ID Attribute', () => {
```

vigneshshanmugam (approved):
LGTM

PR #713, opened 2023-03-06
add types for status alert

shahzad31:
add types for status alert

status alert can be enabled/disable using 
```

    monitor.use({
      schedule: 10,
      alert: {
        status: {
          enabled: false,
        },
      },
    });
```

in lightweight:
```

heartbeat.monitors:
- type: http
  name: Todos Lightweight
  id: todos-lightweight
  enabled: true
  urls: "https://elastic.github.io/synthetics-demo/"
  schedule: '@every 3m'
  timeout: 16s
  alert.status.enabled: true
```

### DIFF
--- a/src/push/monitor.ts
+++ b/src/push/monitor.ts
@@ -24,7 +24,6 @@
  */
 
 import { mkdir, rm, readFile } from 'fs/promises';
-import get from 'lodash/get';
 import { join } from 'path';
 import { LineCounter, parseDocument, YAMLSeq, YAMLMap } from 'yaml';
 import { bold, red } from 'kleur/colors';
@@ -240,6 +239,7 @@ export const parseAlertConfig = (config: MonitorConfig) => {
       },
     };
   }
+  return config.alert;
 };
 
 export function parseSchedule(schedule: string) {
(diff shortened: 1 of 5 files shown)


src/push/monitor.ts
```diff
@@ -24,6 +24,7 @@
  */
 
 import { mkdir, rm, readFile } from 'fs/promises';
+import get from 'lodash/get';
```
vigneshshanmugam, line 27: not used.

vigneshshanmugam (commented):
Probably this is a bit of a different usage case, Do you foresee a future where we might want to disable individual alerts for monitor instead of just disabling them all which is what the PR is doing. 

So its of the type here it would be like `alert: []Alerts` for a given monitor. /cc @dominiqueclarke

dominiqueclarke:
> Probably this is a bit of a different usage case, Do you foresee a future where we might want to disable individual alerts for monitor instead of just disabling them all which is what the PR is doing.

> So its of the type here it would be like alert: []Alerts for a given monitor. /cc @dominiqueclarke

I think we envision the type being an object representing the different alerts, rather than an array, which we can add to later. So let's say we added a tls alert, the type would be

```
alerts: {
   status: {
       enabled: boolean
   },
   tls: {
       enabled: boolean
   },
}
```

dominiqueclarke (approved):
LGTM but you may want to add a test.

Also, and this isn't a blocker for me, it doesn't support nested yaml for lightweight monitors, only `alert.status.enabled`

PR #719, opened 2023-03-24
fix: avoid hanging on capturing network data

vigneshshanmugam:
+ Fixes the issue when we might be waiting for the promises to be fulfilled forever if the responses is chunked which might keep out promises to be hanging forever. 
+ This issue is not easily reproducible and so, its quite hard to add a test case for this with chunked response. We already have a test case for capturing chunked request informatiion, so it should cover this already. 
+ Also added more debug logging as it was the only way to identify this issue on our global infrastructure. 
```
DEBUG=synthetics npm run test
```

### DIFF
--- a/src/core/gatherer.ts
+++ b/src/core/gatherer.ts
@@ -125,11 +125,13 @@ export class Gatherer {
   }
 
   static async dispose(driver: Driver) {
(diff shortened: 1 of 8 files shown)


andrewvc (approved):
LGTM

PR #726, opened 2023-03-30
Add initial workflow to activate GH Actions workflows

reakaleek:


.github/workflows/hello.yml
```diff
@@ -0,0 +1,7 @@
+on: pull_request
```
amannocci, line 1:
```suggestion
on:
  pull_request:
```

amannocci (commented):
There is a small syntax error

.github/workflows/hello.yml
reakaleek: AFAIK, both works. but yeah, your version is the one you see in the docs a lot, so I'll just commit the suggestion.

reakaleek:
Closing, I seems like creating a PR on the upstream triggered a workflow and now it's triggered for forks also.

PR #728, opened 2023-03-30
Migrate Jenkinsfile to GH Actions Workflow

reakaleek:
## Details

- Migrate CI part from Jenkinsfile to GH Actions workflow.

## Follow-Ups

- Release parts
- OpenTelemetry: https://github.com/elastic/synthetics/pull/731


amannocci (approved):

.github/workflows/test-reporter.yml
```diff
@@ -0,0 +1,20 @@
+---
+## Workflow to process the JUnit test results and add a report to the checks.
+name: test-reporter
+on:
+  workflow_run:
+    workflows:
+      - test
```
reakaleek, line 7:
```suggestion
      - ci
```

.github/workflows/ci.yml
```diff
@@ -0,0 +1,72 @@
+name: ci
+
+on: pull_request
```
v1v, line 3: and also on branches?

.github/workflows/ci-test-reporter.yml
```diff
@@ -0,0 +1,20 @@
+---
+## Workflow to process the JUnit test results and add a report to the checks.
+name: ci-test-reporter
```
v1v, line 3: Out of curiosity, what's the reason for using a different naming convention? we've been using the name `test-reporter` instead

.github/workflows/ci.yml
reakaleek: https://github.com/elastic/synthetics/pull/728/commits/13fc64e9c21bef3116416e7d8aefd0fb44887f14

.github/workflows/ci-test-reporter.yml
reakaleek: The main reason was because in https://github.com/elastic/synthetics/pull/730 there is another workflow that also has needs a test reporter.

I was thinking about merging it into one in a follow-up. But that would required a little bit of dynamicity I guess. But I already had something in mind.

reakaleek:
> A few comments:
> 
> * Can you add the `opentelemetry.yml` workflow?
> * If `notifyBuildResult(prComment: true)` in the `.ci/Jenkinsfile` then a GitHub comment will be produced, since this is moving away the entire CI to GH actions and the release stage drives the notification outside of that particular step, I'd say it can be removed to avoid those GH commands.

I wanted to add the opentelemetry as a follow-up. But I can also add it here.

.github/workflows/ci-test-reporter.yml
reakaleek: Or can I subscribe to multiple workflows in a single test reporter without any unwanted sideeffects?

v1v: It's possible to add as many workflows as needed as long as the name of the artifacts are the same.

Though I think I understand now your concern, if there are multiple test jobs/workflows then listing the holistic overview of all of them in one single place won't be possible (at the moment) hence splitting the test-report in different and specialised ones will help with knowing the test results for each test workflow. Or maybe I make up the above :)

v1v: Although in this particular case, tests are running as jobs within the same workflow, hence I'd expect there will be no issues if and only if the artifacts to be stored don't clash and have an unique name.

reakaleek: Yes, this was kinda my concern. Splitting it was to be on the safe side.
Also the paths are slightly different for each workflow. Of course I can do a more generic glob pattern.

reakaleek: But I got your point. Let me rename it again. 

Then if it's merged I will be able to see the output on the other PR.

v1v: Let me know when it's renamed it 🙏

v1v (approved):
LGTM, with a minor comment regarding the workflow `ci-test-reporter.yml`, in case it will be renamed to `test-reporter.yml` instead

.github/workflows/ci-test-reporter.yml
reakaleek: https://github.com/elastic/synthetics/pull/728/commits/c95167cecac7d0b0c5efd05b9d5fd25185442225

.github/workflows/ci.yml
```diff
@@ -0,0 +1,76 @@
...
+      - uses: actions/setup-node@v3
+        with:
+          cache: npm
+          node-version-file: .nvmrc
+      - run: npm ci
+      - run: npm run test:unit -- --ci --reporters=default --reporters=jest-junit
+      - uses: actions/upload-artifact@v3
+        with:
+          name: test-results
+          path: "**/junit.xml"
```
v1v, line 51: NIT: just in case there are not test results, then report a failure?

```suggestion
          path: "**/junit.xml"
          if-no-files-found: error
```

.github/workflows/ci.yml
```diff
@@ -0,0 +1,76 @@
...
+          path: dist
+      - run: npm run test
+        timeout-minutes: 10
+        working-directory: __tests__/e2e
+        env:
+          TERM: xterm # this is needed for tput
+      - uses: actions/upload-artifact@v3
+        with:
+          name: test-results
+          path: "**/junit.xml"
```
v1v, line 76:
```suggestion
          path: "**/junit.xml"
          if-no-files-found: error
```

v1v (approved):
LGTM with a tiny NIT

v1v (approved):
💯

vigneshshanmugam (approved):
LGTM

PR #732, opened 2023-04-05
fix: round off nearest schedule for lighweight monitors

vigneshshanmugam:
+ Round off to the nearest supported schedule instead of picking the lowest scheduled value all the time. 
+ Also, adjust the wording around the Lighweight monitors schedule being adjusted when pushed via project monitors.


paulb-elastic (approved):
LGTM

Issue #734, opened 2023-04-12
capture trace data including web vitals for step errors


vigneshshanmugam:
+ Sometimes when the step fails with timeout errors for heavy pages which happens most of the times due to throttling in place, the network data would be missing since we might not receive the `load` event on time or some of the network data might have not completed due to `Chunked` responses. 

In these scenarios, It would make sense to capture the relevant web vital/performance metrics like FCP, DCL, Load Event. So the user have more information on whats happening along with the network information.

PR #739, opened 2023-04-18
fix: capture trace metrics including web-vitals for failed steps

vigneshshanmugam:
+ fix https://github.com/elastic/synthetics/issues/734
+ This PR fixes the problem when the runner discards the trace events for failed steps. Now, we include all the trace metrics and web vitals including FCP, CLS, LCP for all steps even if they are succeeded or failed.


andrewvc (approved):
LGTM WFG

PR #741, opened 2023-04-19
fix: use headless value from playwrightOptions

### LINKED ISSUE #740: Passing headless `false` still executes tests in headless mode
+ If you pass `headless: false` to `PlaywrightOptions`, the test is expected to run in regular (non-headless) mode, but its not working correctly as CLI args `headless` takes precedence.

vigneshshanmugam:
+ fix #740 
+ Fixes the issue with not able to run in non headless mode in any environments. Also removes the unused `no-headless` flag that is not required anymore as we can simulate it based on `--playwright-options '{"headless": false}`

### DIFF
--- a/__tests__/options.test.ts
+++ b/__tests__/options.test.ts
@@ -33,7 +33,9 @@ describe('options', () => {
       params: {
         foo: 'bar',
       },
-      headless: true,
+      playwrightOptions: {
+        headless: false,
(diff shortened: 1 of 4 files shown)


dominiqueclarke (approved):
LGTM! `headless: false` to playwright options ran tests headful.

PR #742, opened 2023-04-20
feat: allow setting navigation and action timeouts

### LINKED ISSUE #738: Set the default navigation timeout to 50 seconds
We have been seeing numerous sites genuinely failing to load within the current default of 30 seconds, which leads to failures.  As such, the default timeout for navigation timeouts should be changed to 50 seconds.

vigneshshanmugam:
+ fix #738
+ PR changes the default navigation timeout for all journeys to `50 seconds` instead of `30 seconds`. Same applies to all actions as well which is controlled by `actionTimeout`. 
+ Allow setting the timeouts mainly `navigationTimeout` and `actionTimeout` via the `playwrightOptions` configuration. These are  prioritized over the default timeout of `50 seconds` which is set by the agent.
+ This is backwards compatible meaning, individual method timeouts would still be preferred over both of these timeouts. Ex: `page.go("url", {timeout: 2000})` would be preferred over `playwrightOptions.navigationTimeout` and `default 50 seconds` timeout.

### DIFF
--- a/__tests__/core/gatherer.test.ts
+++ b/__tests__/core/gatherer.test.ts
@@ -267,6 +267,52 @@ describe('Gatherer', () => {
     });
   });
 
+  describe('Timeouts', () => {
+    it('set navigation timeout', async () => {
+      const driver = await Gatherer.setupDriver({
+        wsEndpoint,
+        playwrightOptions: { navigationTimeout: 1, actionTimeout: 2 },
+      });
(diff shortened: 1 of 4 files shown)


dominiqueclarke (approved):
LGTM. I tested by setting the throttling extra low to force a page timeout. Was able to overwrite the timeout successfully and see the default 50 second timeout in place.

Issue #737, opened 2023-04-17
Include network request and page data for incomplete requests/pages


paulb-elastic:
When a page times out (often on the `page.goto`), we don’t currently record performance metrics (like the core web vitals) for the page, or network requests (as seen on the waterfall chart) for incomplete requests (i.e. HTTP requests that have been made and received, by point at which the timeout has been triggered).

These are valuable metrics that should be included even when a page times out.

In the case of the network requests, we should represent these (in the waterfall chart) with whatever data we have up to the point of the timeout.  Maybe that’s DNS, TCP, TLS, send request for example, but we don’t have the complete response.

It should be clear in the UI that the network request is incomplete (e.g. a separate colour for the bar on the waterfall chart).

There needs to be a corresponding change in Kibana to show these in the Synthetics UI.

PR #747, opened 2023-04-25
feat: capture network info for incomplete requests

vigneshshanmugam:
+ fix https://github.com/elastic/synthetics/issues/737
+ The network timing calculation logic has been moved to the `response` event (When the time to first byte is received) instead of waiting for the complete request to be finished/failed. This helps capture enough infomation when the request is timing out or killed in the middle of a test. 
+ Also the PR contains fallback information which can calculate the `total` duration of the request when the timing information is not available from Playwright. 
+ Added couple of guards where required so we dont get any negative values. There are multiple tests in the `network.test.ts` which accounts for Cached, Aborted, Fulfilled, Broken, Timed out requests.

### DIFF
--- a/__tests__/plugins/network.test.ts
+++ b/__tests__/plugins/network.test.ts
@@ -287,6 +287,37 @@ describe('network', () => {
     });
   });
 
+  it('timings for broken response', async () => {
+    const driver = await Gatherer.setupDriver({
+      wsEndpoint,
+    });
+    const network = new NetworkManager(driver);
+    await network.start();
+
+    server.route('/index', async (_, res) => {
+      res.setHeader('content-type', 'text/html');
+      await delay(10);
+      res.write(`<head></head>`);
+      await delay(500);
+      res.end(`<body></body>`);
+    });
+
+    await Promise.race([
+      driver.page.goto(server.PREFIX + '/index'),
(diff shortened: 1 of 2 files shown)


devcorpio:
LGTM! 🎉

__tests__/plugins/network.test.ts
```diff
@@ -287,6 +287,36 @@ describe('network', () => {
...
+      .goto(server.PREFIX + '/index', { timeout: 100 })
+      .catch(() => {});
+
+    const netinfo = await network.stop();
+    await Gatherer.stop();
+
+    expect(netinfo.length).toBe(1);
+    // responseEnd is never fired as we are not waiting for the
+    // `responseFinished` event instead cutting off after TTFB
+    expect(netinfo[0].timings.receive).toEqual(-1);
```
devcorpio, line 315: On the Kibana side we will identify the incomplete network requests thanks to the receive -1. Is it right?

devcorpio (approved):

__tests__/plugins/network.test.ts
vigneshshanmugam: Kibana side needs some tuning but for now it will not show any download event when receive is -1 which is a fallback.

PR #753, opened 2023-05-03
chore: remove unused network cdp command

vigneshshanmugam:
+ We have migrated away from plain CDP `devtools` network events and using PW events, so removing it as its not used.


devcorpio:
LGTM

PR #757, opened 2023-05-03
[Synthetics] Migrate `SyntheticsGenerator` code back to Synthetics

justinkambic:
Exports `SyntheticsGenerator` class from top-level core. Works around some export issues with playwright-core so the Synthetics Recorder can access this class.

Also imports tests and types from the recorder.


vigneshshanmugam (approved):
LGTM

PR #758, opened 2023-05-05
feat: disable throttling and emit warning

### LINKED ISSUE #756: Disable (ignore) throttling settings
For the launch of Synthetics, there is an issue where we need to disable throttling (this will be added in again in the future).

ACs:
* Synthetics should ignore any throttling settings when running the test (i.e. run with throttling disabled)
* When `push` is used, include a message informing the user
  * This needs to be generic enough because throttling will be enabled again at some future point (in the case of the service, this will be independent of the Synthetics version they are running `push` from)
  * This should be added alongside a warning triangle (similar to how we currently warn about frequencies being rounded
  * The copy will include the URL where they can find out more information
  * Copy suggestion (@florent-leborgne does this seem ok to you?):
  > ⚠️ Throttling may not be active when the tests run - see https://github.com/elastic/synthetics/blob/main/THROTTLING.md for more details.

Linking to the Kibana issue: https://github.com/elastic/kibana/issues/156512

vigneshshanmugam:
+ fix #756
+ Disables the Network emulation throttling settings via Chrome devtools in the browser CDP session and also provides warnings to the user at different places when users tries
   - Run tests locally with `--throttling` flag
   - Run tests locally with `--no-throttling` flag
   - Push monitors with `--throttling` and `--no-throttling` flag.
   - Push monitors with `monitor.use({throttling: {} )` or `monitor.use({throttling: false})`  on journey level. 
   - Push monitors when `monitor.use({throttling: {}))` set on file level. 
 + Cleaned up the old throttling formatting code as its not required anymore since we have disabled throttling altogether.

### DIFF
--- a/THROTTLING.md
+++ b/THROTTLING.md
@@ -1,1 +1,0 @@
-⚠️ Placeholder for throttling details in Elastic's global managed testing infrastructure (the service) and for Private Locations
--- a/__tests__/cli.test.ts
+++ b/__tests__/cli.test.ts
@@ -27,12 +27,7 @@ import { join } from 'path';
 import { devices } from 'playwright-chromium';
 import { Server } from './utils/server';
 import { CLIMock } from './utils/test-config';
-import {
-  DEFAULT_THROTTLING_OPTIONS,
-  getNetworkConditions,
-  megabitsToBytes,
-  safeNDJSONParse,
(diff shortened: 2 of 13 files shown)


andrewvc (approved):
LGTM

PR #759, opened 2023-05-05
chore: fix release gh action script

vigneshshanmugam:
+ The old Jenkins script was cleaning up manually after `npm install` step which is not required. Also, we cant call `npm run clean` as the dependencies wont be available till the environment is prepared with `npm ci`. Example error - https://github.com/elastic/synthetics/actions/runs/4898079942/jobs/8746832489


reakaleek (approved):
Thanks! Lgtm

PR #760, opened 2023-05-06
feat: support params for lw monitors with push

### LINKED ISSUE #725: `push` does not send params for lightweight monitors
When pushing lightweight project monitors, it does not include the params defined in either `synthetics.config.ts` or from `--params`.  With the implementation added in https://github.com/elastic/kibana/issues/147467, that means the behaviour is different between browser monitors and lightweight monitors.

i.e. The `push` command should also include the params defined as they can be for browser monitors (config file, `--params`) when sending to Kibana.  The priority of the solution should be the same as for browser monitors (see https://www.elastic.co/guide/en/observability/master/synthetics-params-secrets.html#synthetics-basic-params), i.e.

1. Global params in Kibana
2. Then `synthetics.config.ts`
3. Then `--params` defined in `push`

Ensure the order remains consistent and as defined above with both browser and lightweight monitors after this change.

vigneshshanmugam:
+ fix #725 
+ Add support for pushing params for Lightweight monitors when used in
    - Synthetics config file 
    - CLI flags `--params {"foo": "bar"}`
 + Added a backwards compatibility check to drop `params` config if the push is performed on stack `<8.7.2`. 
+ Check the changes with Kibana PR - https://github.com/elastic/kibana/pull/157226

### DIFF
--- a/__tests__/push/index.test.ts
+++ b/__tests__/push/index.test.ts
@@ -34,6 +34,7 @@ import { CLIMock } from '../utils/test-config';
 import { THROTTLING_WARNING_MSG } from '../../src/helpers';
 
 describe('Push', () => {
+  let server: Server;
   const PROJECT_DIR = join(__dirname, 'new-project');
   const DEFAULT_ARGS = ['--auth', 'foo'];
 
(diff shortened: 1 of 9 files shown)


vigneshshanmugam:
Update: Waiting for Kibana to accept the params support before making progress on the agent side.

vigneshshanmugam:
Kibana side is done https://github.com/elastic/kibana/pull/157226, ready for review and E2E.

shahzad31 (approved):
Tested while testing the kibana PR , so testing of this PR while testing kibana PR looks good, since kibana PR which i was testing also looks good, so testing of this also auto happened while testing that PR so testing of this PR also looks good !!

PR #761, opened 2023-05-08
feat: support pattern, match, tags for pushing lightweight

### LINKED ISSUE #754: Push command CLI options --pattern, --match and --tags cli not implemented light-weight monitors
Push command CLI options `--pattern`, `--match `and `--tags` cli not implemented for light-weight monitors.

vigneshshanmugam:
+ fix #754 
+ PR adds support for `--pattern, --tags, --match` which will be applied while pushing lighweight monitors as well. 
+ Added guards to catch errors that might arise when trying to apply transforms to Lightweight monitor files (.yml).

### DIFF
--- a/src/push/monitor.ts
+++ b/src/push/monitor.ts
@@ -24,14 +24,17 @@
  */
 
 import { mkdir, rm, readFile } from 'fs/promises';
-import { join } from 'path';
+import { extname, join } from 'path';
 import { LineCounter, parseDocument, YAMLSeq, YAMLMap } from 'yaml';
 import { bold, red } from 'kleur/colors';
 import { Bundler } from './bundler';
-import { SYNTHETICS_PATH, totalist, indent, warn } from '../helpers';
+import { SYNTHETICS_PATH, totalist, indent, warn, isMatch } from '../helpers';
 import { LocationsMap } from '../locations/public-locations';
 import { ALLOWED_SCHEDULES, Monitor, MonitorConfig } from '../dsl/monitor';
 import { PushOptions } from '../common_types';
+
+// Allowed extensions for lightweight monitor files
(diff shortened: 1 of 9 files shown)


src/push/monitor.ts
```diff
@@ -148,9 +151,17 @@ export async function createLightweightMonitors(
   options: PushOptions
 ) {
   const lwFiles = new Set<string>();
+  // Filter monitor files based on the provided pattern
+  const pattern = options.pattern
+    ? new RegExp(options.pattern, 'i')
+    : /.(yml|yaml)$/;
```
devcorpio, line 157: question:

don't we want here the insensitive case flag? 

I'm asking this because we are using that in the "RegExp" scenario and maybe we want to cover yml/yaml files that are not entirely lowercase

src/push/monitor.ts
```diff
@@ -148,9 +151,17 @@ export async function createLightweightMonitors(
...
   const ignore = /(node_modules|.github)/;
   await totalist(workDir, (rel, abs) => {
-    if (!ignore.test(rel) && /.(yml|yaml)$/.test(rel)) {
+    if (
+      !ignore.test(rel) &&
+      pattern.test(rel) &&
+      ALLOWED_LW_EXTENSIONS.includes(extname(abs))
```
devcorpio, line 163: my question about the insensitive case extension would also applied here since includes is case sensitive

devcorpio:
LGTM!

I simply added a "nit" question about case insensitive comparisons

devcorpio (approved):

src/push/monitor.ts
vigneshshanmugam: Good question, I am not sure if its worth the hassle, We could do a followup if necessary.

PR #771, opened 2023-05-17
fix: use correct schedule when scaffolding projects

### LINKED ISSUE #768: Selecting 3 minute schedule does not set the default correctly
Using Synthetics 1.0.0, when running `init` and selecting 3 minutes as the default, the `monitor.schedule` is not set correctly, so `push` fails.

**Steps to reproduce**
* Run `npx @elastic/synthetics init demo`
* Set the Cloud ID and API key
* Select any values for the options except for _schedule_ and set this to `3`
* Once the `init` has completed, you will see that `synthetics.config.ts` has set `monitor.schedule` to an empty value (`’’`)
* Trying to `push` fails because the `schedule` is not correctly defined

⚠️ not every combination of schedule has been checked, but `3` fails and `10` and `15` work, as several examples tried

image

vigneshshanmugam:
+ fix #768 
+ The underying library gets confused with the index vs the value being passed as initial value. So we correctly use the default schedule now `10` which will be highlighted by the prompt.

### DIFF
--- a/src/generator/index.ts
+++ b/src/generator/index.ts
@@ -133,8 +133,8 @@ export class Generator {
         type: 'select',
         name: 'schedule',
         message: 'Set default schedule in minutes for all monitors',
-        initial: '10', // set default schedule to 10 minutes
-        choices: ALLOWED_SCHEDULES.map(String),
+        initial: 3, // Index of the third array item which is 10 minutes
+        choices: ALLOWED_SCHEDULES.map(value => ({ name: value, value })),
         required: true,
         result(value) {
           return Number(value) as any;
--- a/src/generator/index.ts
+++ b/src/generator/index.ts
@@ -133,12 +133,9 @@ export class Generator {
         type: 'select',
         name: 'schedule',
         message: 'Set default schedule in minutes for all monitors',
-        initial: 3, // Index of the third array item which is 10 minutes
-        choices: ALLOWED_SCHEDULES.map(value => ({ name: value, value })),
+        initial: '10', // set default schedule to 10 minutes
+        choices: ALLOWED_SCHEDULES.map(String),
         required: true,
-        result(value) {
-          return Number(value) as any;
-        },
       },
       {
         type: 'input',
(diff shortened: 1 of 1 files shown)


devcorpio:
Vignesh, I added a change in this PR since there was still a bug related to the **3** minutes schedule.

The schedule option in the config file was still populated with an empty string.

Choosing the number **3** either if it was the default option or by selecting it manually (since it's not the default anymore) caused the issue.

My first reaction has been 🤯 "is there anything special about the number 3? This is not possible".

I debugged the Enquirer library to see what was going on behind the curtains, and the results were surprising:

**Example** of behaviour **with number 5**

Screenshot 2023-05-17 at 14 23 35

**Example** of behaviour **with number 3**

Screenshot 2023-05-17 at 14 24 46

As you can see in the second screenshot, the result for number 3 ended up disappearing in a particular step of the flow. Probably something was going on in the setter/event listeners associated with "this.value" that we can find in the file [prompt.js](https://github.com/enquirer/enquirer/blob/master/lib/prompt.js#L408)

So after seeing it, I ended up having bizarre theories. One of them made me change the allowed scheduled entries order from "1, **3**, 5, 10, 15, 20, 30, 60, 120, 240" to "1, 5, 10, **3**, 15, 20, 30, 60, 120, 240" and everything started working. 

Obviously, that had no sense at all, it seems a bug in the library. 

Fortunately, looking at the [types](https://github.com/enquirer/enquirer/blob/master/index.d.ts#L37) for a Select prompt, I've found there is another way to initialize the choices, instead of a collection of strings, a collection of [Choices](https://github.com/enquirer/enquirer/blob/master/index.d.ts#L19)

Making that change, everything works. You will see that I reverted your previous change, too. The initial value will be **3** again which resolves to index 3 of Choice[] which value is **10**

devcorpio:
something is going on with value 1, checking...
Screenshot 2023-05-17 at 15 10 26

devcorpio:
The **result** option that we were sending to the library Enquirer was not behaving well when using the casting and since the first approach explained [here](https://github.com/elastic/synthetics/pull/771#issuecomment-1551340054) was causing another misbehaviour from Enquirer.

```
      {
        type: 'select',
        name: 'schedule',
        message: 'Set default schedule in minutes for all monitors',
        initial: 3, // Index of the third array item which is 10 minutes
        choices: ALLOWED_SCHEDULES.map(String),
        required: true,
        result(value) {
          return Number(value) as any;
        },
      },
```

Because of that, from now on, the casting will performed by our code and not from a function that Enquirer library executes.

Now, all the schedule options work well 😰 🚀

emilioalvap (approved):
LGTM

devcorpio:
thank you reviewing it @emilioalvap

Issue #775, opened 2023-05-26
The `--tags` filter is failing to match monitors and deletes all monitors on `push`


paulb-elastic:
Having a browser monitor with the following configuration:

```
monitor.use({
  id: 'my-example-journey',
  schedule: 60,
  enabled: false,
  tags: ['foo']
});
```

Should identify this monitor with a `foo` tag.  In fact it does if you `push` this, you will see it shows `foo`:

![image](https://github.com/elastic/synthetics/assets/70908889/85a1bcc0-8a4c-4f22-9723-c1399bb6c3e1)

However, when using `push --tags foo`, it seems `push` is failing to match the monitor and deletes it (and in fact all monitors as nothing matches that tag).

![image](https://github.com/elastic/synthetics/assets/70908889/a5d8a346-9434-4b76-9e07-585c47ad76af)

paulb-elastic:
Closing as there was some confusion over behaviour - refer to https://github.com/elastic/synthetics/issues/780

PR #776, opened 2023-05-29
fix: filter monitors by tags added through monitor.use

devcorpio:
Fixes: https://github.com/elastic/synthetics/issues/775

# Context

the `--tags` flag in the push command was only taking into account tags passed as an argument in the `journey` function

```
journey({ name: 'My Example Journey', tags: ['foo']}, ({ page, params }) => {
   // ... 
})
```

# Fix

Tags added through `monitor.use` are now included in the journey before filtering monitors by `tag`.

## How to test

1. Create [project monitor](https://www.elastic.co/guide/en/observability/current/synthetics-get-started-project.html#synthetics-get-started-project-init)
2. Make sure you add the tag via `monitor.use`.  For instance:

  ```
journey('My Example Journey', ({ page, params }) => {
  monitor.use({
    tags: ['foo']
  });

    await page.goto(params.url);
  });

    const header = await page.locator('h1');
  });
});

  ```
 
3. execute `SYNTHETICS_API_KEY=YOUR_KEY npm run push`. You should see 4 added tests (at least if you stay with the default tests).
4. execute `SYNTHETICS_API_KEY=YOUR_KEY npm run push -- --tags foo`. You should see 3 removed tests and one unchanged

**Recording**:

https://github.com/elastic/synthetics/assets/15065076/7f1d3f60-0658-4257-9cee-1580a185522a

### DIFF
--- a/__tests__/core/runner.test.ts
+++ b/__tests__/core/runner.test.ts
@@ -809,4 +809,18 @@ describe('runner', () => {
     expect(monitors.length).toBe(1);
     expect(monitors[0].config.name).toBe('j1');
   });
+
+  it('runner - build monitors filtered through "tags" added updating monitor config', async () => {
+    const j1 = new Journey({ name: 'j1' }, noop);
+    const j2 = new Journey({ name: 'j2' }, noop);
+    const j3 = new Journey({ name: 'j3' }, noop);
+    j1.updateMonitor({ tags: ['first'] });
+    runner.addJourney(j1);
+    runner.addJourney(j2);
+    runner.addJourney(j3);
+
+    const monitors = runner.buildMonitors({ tags: ['first'], schedule: 1 });
+    expect(monitors.length).toBe(1);
+    expect(monitors[0].config.name).toBe('j1');
+  });
 });
--- a/__tests__/core/runner.test.ts
+++ b/__tests__/core/runner.test.ts
@@ -809,18 +809,4 @@ describe('runner', () => {
     expect(monitors.length).toBe(1);
     expect(monitors[0].config.name).toBe('j1');
   });
-
-  it('runner - build monitors filtered through "tags" added updating monitor config', async () => {
-    const j1 = new Journey({ name: 'j1' }, noop);
-    const j2 = new Journey({ name: 'j2' }, noop);
-    const j3 = new Journey({ name: 'j3' }, noop);
-    j1.updateMonitor({ tags: ['first'] });
-    runner.addJourney(j1);
-    runner.addJourney(j2);
-    runner.addJourney(j3);
-
-    const monitors = runner.buildMonitors({ tags: ['first'], schedule: 1 });
-    expect(monitors.length).toBe(1);
-    expect(monitors[0].config.name).toBe('j1');
-  });
 });
--- a/src/core/runner.ts
+++ b/src/core/runner.ts
@@ -399,9 +399,6 @@ export default class Runner {
 
     const monitors: Monitor[] = [];
     for (const journey of this.journeys) {
-      if (!journey.isMatch(options.match, options.tags)) {
-        continue;
-      }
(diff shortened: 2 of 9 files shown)


emilioalvap (approved):
LGTM

src/core/runner.ts
```diff
@@ -410,6 +407,15 @@ export default class Runner {
       journey.callback({ params: options.params } as any);
       journey.monitor.update(this.monitor?.config);
       journey.monitor.validate();
+
+      /**
+       * Add the tags created through `monitor.use` to the journey
+       */
```
vigneshshanmugam, line 413: Why not add this as part of update monitor which handles all config from 'monitor.use'?

src/core/runner.ts
```diff
@@ -410,6 +407,15 @@ export default class Runner {
...
+      journey.addTags(journey.monitor.config.tags);
+      if (!journey.isMatch(options.match, options.tags)) {
```
vigneshshanmugam, line 415: I don't think this is the correct behaviour. Monitor.use should never influenced the local runner. It should be only applicable to the pushed monitors.

vigneshshanmugam (changes requested):

__tests__/core/runner.test.ts
```diff
@@ -809,4 +809,18 @@ describe('runner', () => {
     expect(monitors.length).toBe(1);
     expect(monitors[0].config.name).toBe('j1');
   });
+
+  it('runner - build monitors filtered through "tags" added updating monitor config', async () => {
```
vigneshshanmugam, line 813: Instead of having this test in runner, it should be on the push command as monitor.use only touches the pushed monitors and never deals with runner parameters.

andrewvc:
Have we tested this with the `synthetics.config.ts` and tags declared there?

__tests__/core/runner.test.ts
devcorpio: thank you for the context 🚀 . I will have a look at this

devcorpio: done!

devcorpio:
> Have we tested this with the synthetics.config.ts and tags declared there?

Thanks for asking this, @andrewvc. I didn't know it was possible to add tags to that file. This scenario wasn't working, I added the required changes for that, too. 

These are the commits related to that change:
- https://github.com/elastic/synthetics/pull/776/commits/854d5d84f3003af8441529f7ee22e16066e36abd
- https://github.com/elastic/synthetics/pull/776/commits/be0febadb8d900ddbfabc78dc89c9142b8e37eee

You will see I added a few tests on that area as well

src/options.ts
```diff
@@ -115,6 +115,7 @@ export function normalizeOptions(cliArgs: CliArgs): RunOptions {
   options.locations = cliArgs.locations ?? monitor?.locations;
   options.privateLocations =
     cliArgs.privateLocations ?? monitor?.privateLocations;
+  options.tags = cliArgs.tags ?? monitor?.tags;
```
devcorpio, line 118: @vigneshshanmugam @andrewvc 

I have a question:

This block starts with:

 /**
   * Get the default monitor config from synthetics.config.ts file
  */

But I've noticed that we are just adding:

- schedule
- locations
- privateLocations
- tags (added in this PR)

If you have a look at https://github.com/elastic/synthetics/blob/main/src/dsl/monitor.ts#L47

you will see there are options that we are not getting there. Is there a reason for that? Probably I'm missing context, too. but I believe it's important to ask, just in case

devcorpio:
closing PR. Reason [here](https://github.com/elastic/synthetics/issues/775#issuecomment-1587849721)

devcorpio:
Closing PR. Reason [here](https://github.com/elastic/synthetics/issues/775#issuecomment-1587849721)

PR #779, opened 2023-06-09
fix: use screenshots from synthetics.config file

vigneshshanmugam:
+ Fixes the issue when applying the screenshot config from `monitor` field in the global `synthetics.config.ts` to all pushed monitors.
+ Added a new mode which merges all the relevant monitor config from the monitor key only during the push command instead of performing extra work during the local runner mode. 
+ `screenshot` is defaulted to `on` only when the tests are executed locally to keep the backwards compatibility as before. But during the push command, we prefer the `--screenshot` flag when its configured and use the `synthetics.config.ts`. Everything can be overriden by `monitor.use` API as before. 

Slack - https://elastic.slack.com/archives/CE4MRBU48/p1686321341637449

### DIFF
--- a/__tests__/options.test.ts
+++ b/__tests__/options.test.ts
@@ -43,9 +43,11 @@ describe('options', () => {
       pauseOnError: true,
       config: join(__dirname, 'fixtures', 'synthetics.config.ts'),
     };
+    console.log('normalizeOptions({})', normalizeOptions({}));
     expect(normalizeOptions({})).toMatchObject({
       environment: 'test',
       params: {},
+      screenshots: 'on',
     });
     expect(normalizeOptions(cliArgs)).toMatchObject({
       dryRun: true,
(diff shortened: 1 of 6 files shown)


__tests__/options.test.ts
```diff
@@ -43,9 +43,11 @@ describe('options', () => {
       pauseOnError: true,
       config: join(__dirname, 'fixtures', 'synthetics.config.ts'),
     };
+    console.log('normalizeOptions({})', normalizeOptions({}));
```
devcorpio, line 46: leftover here 💃

__tests__/options.test.ts
```diff
@@ -76,13 +78,27 @@ describe('options', () => {
...
-      })
+      normalizeOptions(
+        {
+          config,
+          schedule: 3,
+          screenshots: 'only-on-failure',
+          locations: ['australia_east'],
+          privateLocations: ['test'],
+        },
+        'push'
```
devcorpio, line 98: I'm seeing this in my vscode:

devcorpio:
LGTM! Although I have a question with the first scenario I tested, more info below:

Manually tested:

**execute the push command without setting the screenshots CLI flag**

```
options {
  ...other options...
  screenshots: undefined,
  ...other options...
}
```

Question here @vigneshshanmugam:

If the flag value is normalized as **undefined**, in the UI we will see "on". I'm assuming this is the expected behaviour - since it's the default value for the "run" command. Is this correct? I just want to be sure this is intended


Screenshot 2023-06-12 at 12 47 29




**execute the push command setting the screenshots CLI flag** (I chose only-on-failure)

```
options {
  ...other options...
  screenshots: 'only-on-failure',
  ...other options...
}
```

**execute the push command setting the screenshot config in the synthetics.config.file instead of CLI flag** (I chose off)

```
options {
  ...other options...
  screenshots: 'off',
  ...other options...
}
```


**execute the run command without setting the screenshots CLI flag**

```
run options {
  ...other options...
  screenshots: 'on',
  ...other options...
}
```

**execute the run command setting the screenshots CLI flag** (I chose only-on-failure)

```
run options {
  ...other options...
  screenshots: 'only-on-failure',
  ...other options...
}
```

vigneshshanmugam:
> If the flag value is normalized as undefined, in the UI we will see "on". I'm assuming this is the expected behaviour - since it's the default value for the "run" command. Is this correct? I just want to be sure this is intended

This is the correct behaviour as you already figured out. This is why we also have it as `on` in the screenshots for Run mode. 

Thanks for testing all scenarios :)

__tests__/options.test.ts
vigneshshanmugam: Thank you.

devcorpio (approved):

PR #784, opened 2023-06-14
feat: listen to console events in browser context

### LINKED ISSUE #721: listen to console events on browser context instead of page
+ We are not capturing some of the console logs from other pages/windows when the journey involves multiple pages. This can be fixed in our browser console plugin by listening to console events on the browser context, instead of the page. 

We need Playwright's next release which has the feature - https://github.com/microsoft/playwright/pull/22805

vigneshshanmugam:
+ fix #721 
+ Playwright added this capability recently, where we can listen to all the browser console events for all pages that are created via the created browser context. We make use of that and create all the events that are of type `warning, error` and report them for failed tests. 

Check this commit for review - https://github.com/elastic/synthetics/pull/784/commits/33a9320a0358f54f76ee29a5602655a037262821


andrewvc (approved):
LGTM

PR #788, opened 2023-06-19
set engine to node 16.19.1

shahzad31:
Since kibana is on node 16.19.1 and it has has force engine restrict. 

To upgrade synthetics in kibana, we need to lower this restriction. 

I tested it by packing. It seems to cause no issue with node 16.


shahzad31:
closing since kibana itself is being upgraded to node 18

PR #785, opened 2023-06-15
feat: support tls alert key for project monitors

shahzad31:
Support tls alert key

```
heartbeat.monitors:
- type: http
  name: Todos Lightweight
  id: todos-lightweight
  enabled: true
  urls: "https://elastic.github.io/synthetics-demo/"
  schedule: '@every 3m'
  timeout: 16s
  alert.status.enabled: true
  alert.tls.enabled: true
```

PR #795, opened 2023-06-28
fix: handle nested alert keys

shahzad31:
Handle alert config in following format as well

```
heartbeat.monitors:
- type: http
  name: Todos Lightweight
  id: todos-lightweight
  enabled: true
  urls: "${devUrl}"
  schedule: '@every 3m'
  timeout: 16s
  alert:
    status.enabled: false
    tls.enabled: false
```

Edit: 

Continuation of https://github.com/elastic/synthetics/pull/785 which is related to [elastic/kibana#159697](https://github.com/elastic/kibana/pull/159697)

### DIFF
--- a/src/push/monitor.ts
+++ b/src/push/monitor.ts
@@ -180,6 +180,7 @@ export async function createLightweightMonitors(
     const parsedDoc = parseDocument(content, {
       lineCounter,
       keepSourceTokens: true,
+      merge: true,
     });
     // Skip other yml files that are not relevant
     const monitorSeq = parsedDoc.get('heartbeat.monitors') as YAMLSeq<YAMLMap>;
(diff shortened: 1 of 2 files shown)


src/push/monitor.ts
```diff
@@ -180,6 +180,7 @@ export async function createLightweightMonitors(
     const parsedDoc = parseDocument(content, {
       lineCounter,
       keepSourceTokens: true,
+      merge: true,
```
devcorpio, line 183: Why is this change required?

shahzad31: i will revert this. i was testing yaml parsing.

devcorpio:
Should we support the following format, too?

```
  alert:
    status
      enabled: false
    tls
       enabled: false
```

src/push/monitor.ts
shahzad31, line 183:
```suggestion
```

shahzad31:
> 

i think yes, let me add a use case for it. It's getting so ugly :D 

I wish yaml parser could handle it

vigneshshanmugam (changes requested):
It would be better to add a solution to this problem for all keys instead of just doing for `alert`, this problem exists for all the nested fields in HB and its definitely not scalable solution to keep doing this for individual fields.

src/push/monitor.ts
```diff
@@ -260,42 +260,49 @@ export function buildMonitorFromYaml(
...
   }
-  if (config['alert.tls.enabled'] !== undefined) {
-    const value = config['alert.tls.enabled'];
-    delete config['alert.tls.enabled'];
-    alertConfig['tls'] = {
-      enabled: value,
+  return Object.keys(result).length > 0 ? result : undefined;
+};
+
+export function getAlertKeyValue(
```
vigneshshanmugam, line 278: can you add test for this which would take care of `delete behaviour`.

vigneshshanmugam (approved):
LGTM pending comments.

PR #797, opened 2023-07-04
fix: remove match/tags from push command

### LINKED ISSUE #780: Make tags behave consistently
There has been some confusion over the current implementation of tags because it differs from what is expected (for example as raised in https://github.com/elastic/synthetics/issues/775).

Synthetics needs to be updated to consolidate how `tags` (and `match` in the CLI) work, as follows:
- Remove `--tags` and `--match` from `push`
  - If either `--tags` or `--match` is used with `push`, fail the `push` with an error and print a message
    | push cannot be used with either '--tags' or '--match', see [link to forum post] for more on why we removed this broken behavior.
  - Add a deprecation notice to the release note including this change
  - Update the `push --help` text that currently suggests it will filter on `push`
  - Note that `--tags` and `--match` are still valid options for _running_ monitors (but not with `push`)
  - Update the [docs page](https://www.elastic.co/guide/en/observability/current/synthetics-command-reference.html) that `--tags` and `--match` are only relevant when running monitor (not pushing them)
- Tags defined for an **individual monitor**
  - Tags can continue to be defined for an individual monitor via:
    - `monitor.use.tags` of `journey({name:'My Monitor', tags:['foo']}, …`
      - Note that the `journey()` version is not documented or intended for general use (will maybe documented in the future if there's a need)
- Add `monitor.use.tags` support to the `synthetics.config.ts`
  - Precedence order is that this global setting is overridden by any monitor specific setting for `tags`
    - Either `monitor.use.tags` or `journey…` type of definition
  -  i.e. the behaviour is consistent with `locations`
- Tags defined against a monitor in any way are then reflected in the Synthetics UI monitor tags
- Add a sticky forum post for a month or so describing the change

### LINKED ISSUE #787: Consolidate the Private Locations configuration for lightweight project monitors
When configuring a private location setting for a lightweight project monitor, it currently uses a `private_locations: []` configuration, which is different to how they are configured for browser monitors (`privateLocations: []`).  This is different, confusing and not documented.

ACs:
* Include an alias configuration to allow the private location to be configured for lightweight project monitors, using `privateLocations: []`
  * To ensure it’s not a breaking change, keep support for `private_locations: []`
* Update the [documentation](https://www.elastic.co/guide/en/observability/current/synthetics-lightweight.html#synthetics-lightweight-projects) to include how to configure the private location via `privateLocations: []`

vigneshshanmugam:
+ fix #780
+ fix #787 
+ PR removes the ability to add `--tags`and `--match` to the push command and aborts the command if the command is run with any of the CLI flags.
+ Adds the ability to configure tags for the project based browser and lightweight monitors. Will be overridden by the individual monitor.tags or journey.tags if present on the monitors themselves. 
+ `private_locations` and `privateLocations` are supported in the lightweight monitor config. 
+ `alert.status, alert.tls` was not captured from global config file for browser monitors. Fixed that as well. 

### Testing
- Build the project `npm run build`
- Run the Push command 
```sh
SYNTHETICS_API_KEY=<key> node dist/cli.js push --tags "foo*"
// aborts and logs the error
```
- Test with configuring `monitor.tags` in the synthetics.config file and check if its applied on all monitors in the UI.

### DIFF
--- a/src/options.ts
+++ b/src/options.ts
@@ -127,6 +127,11 @@ export function normalizeOptions(
        */
       const monitor = config.monitor;
       for (const key of Object.keys(monitor || {})) {
+        // TODO: screenshots require special handling as the flags are different
+        if (key === 'screenshot') {
+          options.screenshots = options.screenshots ?? monitor[key];
+          continue;
+        }
         options[key] = options[key] ?? monitor[key];
       }
       break;
--- a/src/options.ts
+++ b/src/options.ts
@@ -136,7 +136,9 @@ export function normalizeOptions(
 
 export function validatePushOptions(opts: PushOptions & { match?: string }) {
   if (opts.tags || opts.match) {
-    throw error('Aborted. Invalid flags: tags or match are not supported.');
+    throw error(`Aborted. Invalid CLI flags.
+
+Tags and Match are not supported in push command.`);
   }
 }
 
--- a/src/options.ts
+++ b/src/options.ts
@@ -26,8 +26,8 @@
 import merge from 'deepmerge';
 import { createOption } from 'commander';
 import { readConfig } from './config';
-import type { CliArgs, RunOptions } from './common_types';
-import { THROTTLING_WARNING_MSG, warn } from './helpers';
+import type { CliArgs, PushOptions, RunOptions } from './common_types';
+import { THROTTLING_WARNING_MSG, error, warn } from './helpers';
 
 type Mode = 'run' | 'push';
 
@@ -120,20 +120,24 @@ export function normalizeOptions(
       options.screenshots = cliArgs.screenshots ?? 'on';
       break;
     case 'push':
+      validatePushOptions(options as PushOptions);
       /**
        * Merge the default monitor config from synthetics.config.ts file
        * with the CLI options passed via push command
        */
       const monitor = config.monitor;
-      options.screenshots = options.screenshots ?? monitor?.screenshot;
-      options.schedule = options.schedule ?? monitor?.schedule;
-      options.locations = options.locations ?? monitor?.locations;
-      options.privateLocations =
-        options.privateLocations ?? monitor?.privateLocations;
-      options.alert = options.alert ?? monitor?.alert;
+      for (const key of Object.keys(monitor || {})) {
+        options[key] = options[key] ?? monitor[key];
+      }
       break;
   }
   return options;
+}
+
+export function validatePushOptions(opts: PushOptions & { match?: string }) {
+  if (opts.tags || opts.match) {
+    throw error('Aborted. Invalid flags: tags or match are not supported.');
+  }
 }
 
 /* eslint-disable-next-line @typescript-eslint/no-unused-vars */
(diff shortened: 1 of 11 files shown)


src/options.ts
```diff
@@ -120,22 +120,33 @@ export function normalizeOptions(
...
        */
       const monitor = config.monitor;
-      options.screenshots = options.screenshots ?? monitor?.screenshot;
-      options.schedule = options.schedule ?? monitor?.schedule;
-      options.locations = options.locations ?? monitor?.locations;
-      options.privateLocations =
-        options.privateLocations ?? monitor?.privateLocations;
-      options.alert = options.alert ?? monitor?.alert;
+      for (const key of Object.keys(monitor || {})) {
+        // TODO: screenshots require special handling as the flags are different
```
devcorpio, line 130: Does this TODO mean that you will revisit this area later on?

devcorpio (commented):
**I believe something is wrong with the push command and the tags**

My understanding is that tags will never filter out monitors when pushing, regardless of where those tags are defined (synthetics.config.file, monitor.use.tag, and journey({...})

**Scenario tested:**

I defined a tag within the **synthetics.config.ts** file

Screenshot 2023-07-06 at 12 02 42

These are the monitors I wanted to push (**4** monitors):

Screenshot 2023-07-06 at 12 01 08

After that, I executed the push command:

```SYNTHETICS_API_KEY=mykey npm run push```

My surprise here has been to see that **just one monitor** has been pushed. The only one that has been pushed is the lightweight one (the tag was also visible in the UI). My expectations were to see the 4 monitors pushed and visible in the UI with the corresponding tag.

The filtering (that just affect browser monitors) takes place [here](https://github.com/elastic/synthetics/blob/main/src/core/runner.ts#L403)

This is the output of my logs (you can see 3 entries there)

Screenshot 2023-07-06 at 12 12 17


---

On the other hand, things that I can confirm that work greatly:




--

`SYNTHETICS_API_KEY=mylocalkey npm run push -- --match 'foo*'`

**Output**:

Screenshot 2023-07-06 at 11 24 41


--

`SYNTHETICS_API_KEY=mylocalkey npm run push -- --tags 'baz'`

**Output**:

Screenshot 2023-07-06 at 11 24 41

--

> Test with configuring monitor. tags in the synthetics.config file and check if it's applied on all monitors in the UI.

Screenshot 2023-07-06 at 11 27 54


--
> Will be overridden by the individual monitor.tags or journey.tags if present on the monitors themselves.

**Scenario**: tag defined in **synthetics.config.file** + **journey** 

The tag defined in the **journey** wins.

**Scenario**: tag defined in **synthetics.config.file** + **journey** + **monitor.use.tag**

The tag defined in the **monitor.use.tag** wins.

src/options.ts
vigneshshanmugam: Nope, I will remove TODO here.

vigneshshanmugam:
> My surprise here has been to see that just one monitor has been pushed. The only one that has been pushed is the lightweight one (the tag was also visible in the UI). My expectations were to see the 4 monitors pushed and visible in the UI with the corresponding tag.

Good catch @devcorpio , Fixed it now and added test to capture this. Also found another bug with alerts not being added to global config.

devcorpio (approved):
LGTM! 🎉  everything works now

Issue #802, opened 2023-07-12
Synthetics `push` command doesn't detect changes in imported files


awahab07:
**Agent Version:**  1.3.0

### Summary
When using the `push` command, Synthetics CLI doesn't react to code changes in the imported files, e.g.

```typescript
import { journey, step, monitor, expect } from '@elastic/synthetics';
import { prepare } from '../helpers/helpers';

journey('Journey Name', ({ page, params }) => {
  step('Step 01', async () => {
    await prepare(page);
  });

  ...
});
```

for the above journey, any changes in the `prepare` will not mark the monitor as edited and `push` will not push the updated monitor to Kibana, unless the journey file itself is modified.

#### Steps to reproduce
1. Create an example journey in a Synthetics CLI project.
2. In one of the steps, make a valid operation inside a helper method imported from an outer file e.g. `../helpers/helpers'`.
3. Push the monitor via
    ```
    SYNTHETICS_API_KEY=<API-KEY> npm run push
    ```
4. Make a change in the imported helper method and push again.
5. You'll notice that the changes in the helper method won't be pushed as the modification is not detected by the CLI. If there were no other modifications, CLI will show:
    ```
     > Monitor Diff: Added(0) Updated(0) Removed(0) Unchanged(1)
    ```
    
#### Expected behavior
Any change in any imported file should mark that journey file as modified, even if the journey file isn't changed. Or if possible, mark only those journeys as modified which are consumers of imported components that have been modified.

justinkambic:
Post-FF Testing LGTM

Added an import from an external file to my example project monitor file:

<img width="533" alt="image" src="https://github.com/elastic/synthetics/assets/18429259/b3ea3947-16ca-4d3e-b26c-e9da57670fbf">

<img width="840" alt="image" src="https://github.com/elastic/synthetics/assets/18429259/72d9200a-ce45-4fd0-aaac-fae194dd9989">

PR #804, opened 2023-07-18
fix: update hash logic to account for bundling

shahzad31:
fixes https://github.com/elastic/synthetics/issues/802

update hash logic to account for bundling.

Earlier it was doing difference based on hash calculated before bundling,  in updated logic diff will be done after hash calculated based on content of bundle.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -121,3 +121,5 @@ __tests__/e2e/junit.xml
 seccomp/build
 
 test-projects/
+
+.DS_Store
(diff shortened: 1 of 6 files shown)


vigneshshanmugam (commented):
There is a risk with this approach, Making the hash based on the bundle would be problematic in few scenarios
- If any of the journey changes in a single file, then all journeys are affected
- Bundling code gets affected with esbuild library changes. 

In any case, we need to fix this issue as its not the right behaviour, but lets think through other approaches and decide.

vigneshshanmugam (approved):
LGTM

We will proceed with this approach for now to fix the bug and work on a efficient way to bundle the journeys with the approach discussed here -https://github.com/elastic/synthetics/pull/809

PR #805, opened 2023-07-18
Update base reporter to emit screenshot as .jpg instead of json format

shahzad31:
Update base reporter to omit screenshot as .jpg instead of json format !!

image

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -119,5 +119,3 @@ __tests__/e2e/junit.xml
 
 .idea/
 seccomp/build
-
-test-projects/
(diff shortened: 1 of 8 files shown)


vigneshshanmugam (commented):
We discussed offline, the plan is to come up with these 
- Test Result Interface output with attachments or similar API to get these screenshots 
- Writing to `.synthetics` by default to be in safer side for permission model inside HB
- Enhance the reporters to work on that testoutput.
- Allow overriding the path for emitting the output.

vigneshshanmugam:
Closing in favor of https://github.com/elastic/synthetics/issues/878

Issue #700, opened 2023-02-08
support playwright test assertions in synthetics


vigneshshanmugam:
+ We rely on the default `expect` library assertions which does not have the complete list of assertions that are supported by the Playwright test library. This would be confusing for users who are trying to run these playwright tests in the elastic globally managed infrastructure. 

+ Using the `plawyright-test/expect` assertions would require us to replace the current [expect](https://github.com/elastic/synthetics/blob/cbf4c4e388acb25c86a94e3c99ad516aee9f13e6/src/core/expect.ts) and keep the playwright version up to date #614

jasonwhetton:
please support the standard playwright assertions this is really helpful not least for migrating existing tests.

leweafan:
You are right we can't use our standard tests.
`TypeError: expect(...).toHaveCount is not a function`

jasonwhetton:
hi, is there any plan to pick this up? having web first assertions is pretty much a necessity for using playwright in a stable way.... it seems odd that this isn't prioritised.

paulb-elastic:
@jasonwhetton this is on our backlog and something we will look to add in the future, but we're just not able to focus on it right now.

PR #808, opened 2023-07-25
feat: support playwright locator assertions

shahzad31:
fixes https://github.com/elastic/synthetics/issues/700
support playwright locator assertions

1. Export playwright expect 
2. Export types related to LocationAssertions
3. Export devices from playwright devices

Will add some tests after getting feedback on the approach

### DIFF
--- a/__tests__/cli.test.ts
+++ b/__tests__/cli.test.ts
@@ -319,7 +319,7 @@ describe('CLI', () => {
     }
   });
 
-  it('run expect assetions with type check', async () => {
+  it('run expect assertions with type check', async () => {
     // flag turns on type checking
     process.env['TS_NODE_TYPE_CHECK'] = 'true';
     const cli = new CLIMock()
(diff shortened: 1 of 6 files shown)


vigneshshanmugam (commented):
We caught up on the Sync and decided
- We will split the devices issue where users might need to do `import {devices} from "playwright-core" which might result in pulling all of the PW code and hitting the size limit. Instead of this PR which exports it from `@playwight/test`, we will expose it from Synthetics and `export it from playwright-core` instead of pulling in additional deps. 
- Will do a spike on importing `expect` from `@playwright/test` as the runner might interfere with ours. This could also be done in 
    - simple way - Importing the bundled playwright/expcect library and bundling them on the synthetics side.
    - Solid way - Move to `@playwright/test` and integrate with our agent -https://github.com/vigneshshanmugam/synthetics-pw-runner

vigneshshanmugam:
We will use the approach on this PR https://github.com/elastic/synthetics/pull/812 to make the assertions seamless.

vigneshshanmugam:
Closing as per the above comment.

PR #812, opened 2023-07-28
feat: bundle expect from @playwright/test library

vigneshshanmugam:
+ fix [#808 ](https://github.com/elastic/synthetics/issues/700)
+ Bundles the PW test library expect matcher with the synthetics and replaces the expect from jest in favor of the PW test matchers. 
+ The PR uses the `expect` library from the `@playwright/test` without bringing in the full runner which we do not require as we have our own Synthetics runner which is build with different syntax. This also replaces the `expect` from jest in favor of the PW test matchers which under the hood is built on top of expect.
+ With bundling approach, we get the benefit of only importing the necessary modules from the PW test and core and still utilize all the available matchers except for the matchers that does comparision on the files - `toMatchSnapshot` and `toHaveScreenshot`. 

### Folowup
+ [ ] Update the docs around expect library - https://www.elastic.co/guide/en/observability/8.9/synthetics-create-test.html#synthetics-assertions-methods
+ [x] Callout `toHaveScreenshot` and `toMatchSnapshot` wont work. 
+ [ ] Stacktrace fix to polish the errors from Expect - #824

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,6 +9,7 @@
     "prepublish": "npm run clean && npm run build",
     "build": "tsc",
     "build:locations": "sh utils/update_locations.sh",
+    "build:bundles": "cd bundles && npm run build",
     "watch": "tsc -w",
     "lint": "eslint . --rulesdir utils/eslint-rules",
     "lint:fix": "npm run lint -- --fix",
--- a/package.json
+++ b/package.json
@@ -6,10 +6,10 @@
   "types": "dist/index.d.ts",
   "scripts": {
     "clean": "rimraf dist",
-    "prepublish": "npm run clean && npm run build:bundles && npm run build",
-    "build": "tsc",
+    "prepublish": "npm run clean && npm run build",
+    "build:lib": "tsc",
     "build:locations": "sh utils/update_locations.sh",
-    "build:bundles": "cd bundles && npm run build",
+    "build": "node utils/build.js",
     "watch": "tsc -w",
     "lint": "eslint . --rulesdir utils/eslint-rules",
     "lint:fix": "npm run lint -- --fix",
--- a/package.json
+++ b/package.json
@@ -6,7 +6,7 @@
   "types": "dist/index.d.ts",
   "scripts": {
     "clean": "rimraf dist",
-    "prepublish": "npm run clean && npm run build",
+    "prepublish": "npm run clean && npm run build:bundles && npm run build",
     "build": "tsc",
     "build:locations": "sh utils/update_locations.sh",
     "build:bundles": "cd bundles && npm run build",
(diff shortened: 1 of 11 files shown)


package.json
```diff
@@ -6,9 +6,10 @@
   "types": "dist/index.d.ts",
   "scripts": {
     "clean": "rimraf dist",
-    "prepublish": "npm run clean && npm run build",
+    "prepublish": "npm run clean && npm run build:bundles && npm run build",
```
shahzad31, line 9: seems like it will run npm run build twice in same directory?

vigneshshanmugam: Fixed.

devcorpio (approved):
this looks great! besides the code is clean and easy to follow! 🚀 

just a question related to the PR's description:

You mentioned there that "toHaveScreenshot and toMatchSnapshot" will not work. But when testing your PR, I was able to see them within the playwright expect module:

playwright-expect

I'm sure that what you mentioned is for a particular reason. Just out of curiosity, why are they not going to work?

shahzad31:
> LGTM! This looks great! besides, the code is clean and easy to follow! 🚀
> 
> just a question related to the PR's description:
> 
> You mentioned that "toHaveScreenshot and toMatchSnapshot" will not work. But when testing your PR, I was able to see them within the playwright expect module:
> 
> playwright-expect
> I'm sure that what you mentioned is for a particular reason. Just out of curiosity, why are they not going to work?

we need some typescript magic to remove those assertions or some other magic to throw an error when user tries to use those assertions.

shahzad31 (approved):
LGTM, 

Looks great, i pushed a small change for non supported assertion !!

shahzad31:
image

devcorpio:
> Looks great, i pushed a small change for non supported assertion !!

Nice idea! 🎉 

> Somehow for me types aren't properly highlighted for used assertions


hmm, in vscode seems to work properly. Is perhaps a webstorm "problem"? Do you need to restart or something?

vigneshshanmugam:
> Somehow for me types aren't properly highlighted for used assertions

You need to nuke some node_modules files and do build again for it to work correctly with new types. 

> Looks great, i pushed a small change for non supported assertion !!

Thanks, I forgot to push it. I am working on a small TS change which might show it might not work. I will take your change if its not working.

devcorpio:
> I am not sure how its working for you. 

I meant that the IDE IntelliSense with typescript was working correctly for me(Shahzad could write things like expect(page).toHaveInventedAssertion())

vigneshshanmugam:
I see, typings would work, but the method itself wont work which Is why we have to exclude them or show errors.

paulb-elastic:
https://github.com/elastic/observability-docs/issues/3206 raised for the docs update

Issue #718, opened 2023-03-23
Make `--url` optional for `locations` command


paulb-elastic:
When using `push`, the `--url` parameter is not needed because it’s contained in the `synthetics.config.ts` file.

However, when using the `locations` command, the `--url` param is needed or it only returns the [internally defined locations](https://github.com/elastic/synthetics/blob/main/src/locations/public-locations.ts).

It would be useful to only need to specify the `--auth` (or have it predefined in the environment) and if there is a `project.url` defined in `synthetics.config.ts`, then use that by default (there by outputting any private locations too).  It should still be possible to override this using `--url`, but it should behave more consistently with `push`.

ACs:
* Running `locations` on it’s own, returns the [internally defined locations](https://github.com/elastic/synthetics/blob/main/src/locations/public-locations.ts)
* Running `locations --auth ...`
  * If `project.url` is defined in `synthetics.config.ts`, CLI returns the locations from that Kibana instance (for public and private locations)
  * If `project.url` is NOT defined in `synthetics.config.ts`, CLI return the [internally defined locations](https://github.com/elastic/synthetics/blob/main/src/locations/public-locations.ts)
* Running `locations --auth ... --url ...` returns the locations from that Kibana instance (for public and private locations)
  * Effectively ignoring any `project.url` definition in `synthetics.config.ts`

shahzad31:
it's working for me without adding url option

```
➜  test-projects git:(locations-url) npx @elastic/synthetics locations
Available locations: 
   * japan
   * india
   * singapore
   * australia_east
   * united_kingdom
   * germany
   * canada_east
   * brazil
   * us_east
   * us_west
   
Set default location for monitors via
  - Synthetics config file 'monitors.locations' | 'monitors.privateLocations' field
  - Monitor API 'monitor.use({ locations: ["japan"], privateLocations: ["custom-location"] }')

```

paulb-elastic:
To clarify @shahzad31's point - that's only for the main locations (generated internally by Synthetics from https://github.com/elastic/synthetics/blob/main/src/locations/public-locations.ts), but doesn't include the private locations without `--url` being specified (even though it's in the `synthetics.config.ts` file).

PR #814, opened 2023-08-01
fix: align cli options between push and locations for url/auth

shahzad31:
fixes https://github.com/elastic/synthetics/issues/718
align cli options between push and locations for url/auth !!

image

### DIFF
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -150,21 +150,19 @@ program
       tearDown();
     }
   });
-
+const authOption = new Option(
+  '--auth <auth>',
+  'API key used for Kibana authentication(https://www.elastic.co/guide/en/kibana/master/api-keys.html).'
+)
+  .env('SYNTHETICS_API_KEY')
+  .makeOptionMandatory(true);
 // Push command
 program
   .command('push')
   .description(
     'Push all journeys in the current directory to create monitors within the Kibana monitor management UI'
   )
-  .addOption(
-    new Option(
-      '--auth <auth>',
-      'API key used for Kibana authentication(https://www.elastic.co/guide/en/kibana/master/api-keys.html).'
-    )
-      .env('SYNTHETICS_API_KEY')
-      .makeOptionMandatory(true)
-  )
+  .addOption(authOption)
   .option(
     '--schedule <time-in-minutes>',
     "schedule in minutes for the pushed monitors. Setting `10`, for example, configures monitors which don't have an interval defined to run every 10 minutes.",
@@ -245,21 +243,36 @@ program
     `List all locations to run the synthetics monitors. Pass optional '--url' and '--auth' to list private locations.`
   )
   .option('--url <url>', 'Kibana URL to fetch all public and private locations')
-  .option(
-    '--auth <auth>',
-    'API key used for Kibana authentication(https://www.elastic.co/guide/en/kibana/master/api-keys.html).'
-  )
+  .addOption(authOption)
   .action(async (cmdOpts: LocationCmdOptions) => {
+    const workDir = cwd();
+    let url = cmdOpts.url;
+    let tearDown = () => {};
     try {
-      let locations = Object.keys(LocationsMap);
-      if (cmdOpts.auth && cmdOpts.url) {
-        const allLocations = await getLocations(cmdOpts);
-        locations = formatLocations(allLocations);
+      if (!url) {
+        tearDown = await globalSetup({ inline: false, ...program.opts() }, [
+          workDir,
+        ]);
+        const settings = await loadSettings();
+        url = settings.url;
       }
-      renderLocations(locations);
+    } catch (e) {}
+
+    try {
+      if (url && cmdOpts.auth) {
+        const allLocations = await getLocations({
+          url,
+          auth: cmdOpts.auth,
+        });
+        renderLocations(formatLocations(allLocations));
+      } else {
+        renderLocations({ publicLocations: Object.keys(LocationsMap) });
+      }
     } catch (e) {
       e && error(e);
       process.exit(1);
+    } finally {
+      tearDown();
     }
   });
 
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -56,7 +56,8 @@ import { installTransform } from './core/transform';
 /* eslint-disable-next-line @typescript-eslint/no-var-requires */
 const { name, version } = require('../package.json');
 
-const { params, pattern, playwrightOpts, auth } = getCommonCommandOpts();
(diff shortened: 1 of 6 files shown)


src/cli.ts
```diff
@@ -145,6 +145,7 @@ program
       }
     } catch (e) {
       console.error(e);
+      console.log('what the hell', e);
```
vigneshshanmugam, line 148: needs tobe removed.

src/cli.ts
```diff
@@ -245,21 +246,42 @@ program
...
+          url,
+          auth: cmdOpts.auth,
+        });
+        renderLocations(formatLocations(allLocations));
+      } else {
+        renderLocations({ publicLocations: Object.keys(LocationsMap) });
       }
-      renderLocations(locations);
     } catch (e) {
+      console.log('what the hell', e);
```
vigneshshanmugam, line 280: needs tobe removed.

src/cli.ts
```diff
@@ -245,21 +246,42 @@ program
     `List all locations to run the synthetics monitors. Pass optional '--url' and '--auth' to list private locations.`
   )
   .option('--url <url>', 'Kibana URL to fetch all public and private locations')
-  .option(
-    '--auth <auth>',
-    'API key used for Kibana authentication(https://www.elastic.co/guide/en/kibana/master/api-keys.html).'
+  .addOption(
```
vigneshshanmugam, line 249: can we reuse the auth option from the push command and add mandatory only on push.

src/cli.ts
```diff
@@ -245,21 +246,42 @@ program
...
+    const workDir = cwd();
+    let url = cmdOpts.url;
+    let tearDown = () => {};
     try {
-      let locations = Object.keys(LocationsMap);
-      if (cmdOpts.auth && cmdOpts.url) {
-        const allLocations = await getLocations(cmdOpts);
-        locations = formatLocations(allLocations);
+      if (!url) {
+        tearDown = await globalSetup({ inline: false, ...program.opts() }, [
```
vigneshshanmugam, line 261: Move this out of the try block, not required. we don't need the full setup since setup also involves loading test files which are not required. You can just do transform. 
```ts
 const revert = installTransform();
 finally {
 revert()
}

```

src/locations/index.ts
```diff
@@ -54,39 +54,59 @@ export async function getLocations(options: LocationCmdOptions) {
 }
 
 export function formatLocations(locations: Array<LocationMetadata>) {
-  const formatted: Array<string> = [];
+  const publicLocations: Array<string> = [];
```
vigneshshanmugam, line 57: I dont understand the use of this split given we already append `private` keyword for all private locations. 

If the purpose of this is to display them separately, then please change the grouping logic and rendering logic to account for empty private location array.

vigneshshanmugam (changes requested):

src/locations/index.ts
```diff
@@ -54,39 +54,64 @@ export async function getLocations(options: LocationCmdOptions) {
...
+    publicLocations,
+    privateLocations,
+    allLocations: [
+      ...publicLocations,
+      ...privateLocations.map(loc => `${loc} ${PRIVATE_KEYWORD}`),
+    ],
+  };
 }
 
 export function groupLocations(locations: Array<string>) {
```
vigneshshanmugam, line 79: this function only existed because we were merging both locations together, now that you have splitted the locations in to distinct categories, we dont need to split and join again with Private Keyword.

src/locations/index.ts
```diff
@@ -54,39 +54,64 @@ export async function getLocations(options: LocationCmdOptions) {
...
-      name = `${name}${PRIVATE_KEYWORD}`;
+      privateLocations.push(name);
     }
-    formatted.push(name);
   }
-  return formatted;
+  return {
+    publicLocations,
+    privateLocations,
+    allLocations: [
```
vigneshshanmugam, line 72: use concat.

src/cli.ts
```diff
@@ -151,20 +152,16 @@ program
...
   )
-  .addOption(
-    new Option(
-      '--auth <auth>',
-      'API key used for Kibana authentication(https://www.elastic.co/guide/en/kibana/master/api-keys.html).'
-    )
-      .env('SYNTHETICS_API_KEY')
-      .makeOptionMandatory(true)
-  )
+  .addOption(new Option('--auth <auth>', apiDocsLink).makeOptionMandatory(true))
```
vigneshshanmugam, line 164: can you move this option to the shared list here - https://github.com/elastic/synthetics/blob/c560bef082718c9799b93da72a3af95bba4c225f/src/options.ts#L171

src/cli.ts
```diff
@@ -245,21 +242,29 @@ program
...
+  .option('--auth <auth>', apiDocsLink)
   .action(async (cmdOpts: LocationCmdOptions) => {
+    const revert = installTransform();
+    let url = cmdOpts.url;
     try {
-      let locations = Object.keys(LocationsMap);
-      if (cmdOpts.auth && cmdOpts.url) {
-        const allLocations = await getLocations(cmdOpts);
-        locations = formatLocations(allLocations);
+      url = (await loadSettings())?.url;
```
vigneshshanmugam, line 250: loadSettings will print error on stderr and cannot be caught, We need to fix this inorder to not print the message.

src/locations/index.ts
vigneshshanmugam: Had a chat offline, this was done to make `locations` command look nicer with different types of locations, we still will use the keyword on init.

src/locations/index.ts
shahzad31: Is using spread operator wrong in this ?

src/cli.ts
shahzad31: actually we can't reuse it, since once it's made mandatory, it will becomes mandatory for auth as well

src/cli.ts
shahzad31: not sure , what you mean? should i fix the function?

src/locations/index.ts
vigneshshanmugam: unnecessary copy. would like to keep it simpler than using spread when not required.

src/cli.ts
vigneshshanmugam: if you run locations command now, you will see error printed on console saying no `synthetics.config.ts` file.

src/cli.ts
vigneshshanmugam: But we can enable/disable auth when adding. Pushed the change

PR #817, opened 2023-08-17
feat: collect logs for successful journey as well

shahzad31:
collect logs for successful journey as well !!

For successful journey we collect last 10 errors, last 5 warns, infos and logs

### DIFF
--- a/src/helpers.ts
+++ b/src/helpers.ts
@@ -36,6 +36,8 @@ import {
   NetworkConditions,
   Location,
   ThrottlingOptions,
+  StatusValue,
+  BrowserMessage,
 } from './common_types';
 import micromatch from 'micromatch';
 
@@ -426,4 +428,22 @@ export function isMatch(
 export function tagsMatch(tags, pattern) {
   const matchess = micromatch(tags || ['*'], pattern);
   return matchess.length > 0;
+}
+
+export function filterBrowserConsoles(
+  messages: BrowserMessage[],
+  status: StatusValue
+) {
+  if (status === 'failed') {
+    return messages;
+  }
+  if (status === 'succeeded') {
+    // we collect total 100 messages from the browser console, giving errors and warnings priority
+    const errors = messages.filter(msg => msg.type === 'error').slice(-50);
+    const warnings = messages.filter(msg => msg.type === 'warn').slice(-50);
+    const logs = messages.filter(msg => msg.type === 'warn').slice(-50);
+
+    return [...errors, ...warnings, ...logs].slice(100);
+  }
+  return [];
 }
--- a/src/helpers.ts
+++ b/src/helpers.ts
@@ -36,8 +36,6 @@ import {
   NetworkConditions,
   Location,
   ThrottlingOptions,
-  StatusValue,
-  BrowserMessage,
 } from './common_types';
 import micromatch from 'micromatch';
 
@@ -428,22 +426,4 @@ export function isMatch(
 export function tagsMatch(tags, pattern) {
   const matchess = micromatch(tags || ['*'], pattern);
   return matchess.length > 0;
-}
-
-export function filterBrowserConsoles(
-  messages: BrowserMessage[],
-  status: StatusValue
-) {
-  if (status === 'failed' || messages.length < 100) {
-    return messages;
-  }
-  if (status === 'succeeded') {
-    // we collect total 100 messages from the browser console, giving errors and warnings priority
-    const errors = messages.filter(msg => msg.type === 'error').slice(-50);
-    const warnings = messages.filter(msg => msg.type === 'warn').slice(-50);
-    const logs = messages.filter(msg => msg.type === 'warn').slice(-50);
-
-    return [...errors, ...warnings, ...logs].slice(100);
-  }
-  return [];
 }
--- a/src/helpers.ts
+++ b/src/helpers.ts
@@ -434,7 +434,7 @@ export function filterBrowserConsoles(
   messages: BrowserMessage[],
   status: StatusValue
 ) {
-  if (status === 'failed') {
+  if (status === 'failed' || messages.length < 100) {
     return messages;
   }
   if (status === 'succeeded') {
--- a/src/plugins/browser-console.ts
+++ b/src/plugins/browser-console.ts
@@ -31,6 +31,8 @@ import { getTimestamp } from '../helpers';
 
 const defaultMessageLimit = 1000;
 
+const allowedTypes = ['error', 'warning', 'log'];
+
 export class BrowserConsole {
   private messages: BrowserMessage[] = [];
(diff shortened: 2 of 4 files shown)


src/plugins/browser-console.ts
```diff
@@ -83,10 +85,33 @@ export class BrowserConsole {
...
+
+    return this.filterOutput(status);
+  }
+
+  filterOutput(status?: StatusValue) {
+    if (status === 'failed') {
+      return this.messages;
+    }
+    if (status === 'succeeded') {
+      // we collect last 10 error, last 5 logs, infos and last 5 warnings and return those
```
vigneshshanmugam, line 101: it would be better to give more priority to errors/warnings and than to log when we know they are present. 

The algorithm could be something like
- Set a Max of 50/100 messages for a journey
- collect all errors, warnings, then get logs if applicable. 

Right now, this would yield different results for every run as the errors/warn/log could happen at different times.

vigneshshanmugam (changes requested):
Other than the comments, Plugins should not act based on journey results. It would be better to keep the core logic on the runner and have a utility that does the filtering. 

Dont expose the `status` on stop.

src/plugins/browser-console.ts
```diff
@@ -73,7 +75,13 @@ export class BrowserConsole {
 
   private enforceMessagesLimit() {
     if (this.messages.length > defaultMessageLimit) {
-      this.messages.splice(0, 1);
+      // drop logs type first if they exceed the limit
+      const lastLogIndex = this.messages.findIndex(m => m.type === 'log');
```
vigneshshanmugam, line 79: I don't think we need to do this. With 1000 messages on the line which should be more than enough for successful journeys to be filtered on later, this is not necessary and an expensive operation for each console log statements.

src/helpers.ts
```diff
@@ -427,3 +429,21 @@ export function tagsMatch(tags, pattern) {
...
 }
+
+export function filterBrowserConsoles(
+  messages: BrowserMessage[],
+  status: StatusValue
+) {
+  if (status === 'failed') {
+    return messages;
+  }
+  if (status === 'succeeded') {
```
vigneshshanmugam, line 440: couple of optimizations we can do here.
1. if message is less than 100, return without doing unnecessary work.
2. do one pass across the messages and based on types, construct the final messages list with giving priority to error's and warnings.

shahzad31: first make sense, but isn't second micro optimisation?
i feel readability is more important compared to micro optimisation

src/plugins/browser-console.ts
shahzad31: Done

src/helpers.ts
shahzad31: I have done the first

src/helpers.ts
```diff
@@ -427,3 +429,21 @@ export function tagsMatch(tags, pattern) {
...
+  status: StatusValue
+) {
+  if (status === 'failed' || messages.length < 100) {
+    return messages;
+  }
+  if (status === 'succeeded') {
+    // we collect total 100 messages from the browser console, giving errors and warnings priority
+    const errors = messages.filter(msg => msg.type === 'error').slice(-50);
+    const warnings = messages.filter(msg => msg.type === 'warn').slice(-50);
+    const logs = messages.filter(msg => msg.type === 'warn').slice(-50);
```
vigneshshanmugam, line 444:
```suggestion
    const logs = messages.filter(msg => msg.type === 'info').slice(-50);
```

src/helpers.ts
```diff
@@ -427,3 +429,21 @@ export function tagsMatch(tags, pattern) {
...
+  messages: BrowserMessage[],
+  status: StatusValue
+) {
+  if (status === 'failed' || messages.length < 100) {
+    return messages;
+  }
+  if (status === 'succeeded') {
+    // we collect total 100 messages from the browser console, giving errors and warnings priority
+    const errors = messages.filter(msg => msg.type === 'error').slice(-50);
+    const warnings = messages.filter(msg => msg.type === 'warn').slice(-50);
```
vigneshshanmugam, line 443:
```suggestion
    const warnings = messages.filter(msg => msg.type === 'warning').slice(-50);
```

src/helpers.ts
```diff
@@ -427,3 +429,21 @@ export function tagsMatch(tags, pattern) {
...
+    const logs = messages.filter(msg => msg.type === 'warn').slice(-50);
+
+    return [...errors, ...warnings, ...logs].slice(100);
```
vigneshshanmugam, line 446:
```suggestion
    return [...errors, ...warnings, ...logs].slice(-100);
```

src/plugins/browser-console.ts
```diff
@@ -73,7 +75,7 @@ export class BrowserConsole {
 
   private enforceMessagesLimit() {
     if (this.messages.length > defaultMessageLimit) {
-      this.messages.splice(0, 1);
+      this.messages = this.messages.splice(0, 1);
```
vigneshshanmugam, line 78: not sure why this is required. splice does inplace replacement.

vigneshshanmugam (changes requested):
the filtering code deserves a test.

src/helpers.ts
```diff
@@ -427,3 +429,21 @@ export function tagsMatch(tags, pattern) {
   const matchess = micromatch(tags || ['*'], pattern);
   return matchess.length > 0;
 }
+
+export function filterBrowserConsoles(
+  messages: BrowserMessage[],
+  status: StatusValue
+) {
+  if (status === 'failed' || messages.length < 100) {
```
vigneshshanmugam, line 437: skipped should not capture messages.

vigneshshanmugam:
@shahzad31 I pushed the fixes and also added tests that were missing, Will get it merged once you cross check.

src/plugins/browser-console.ts
```diff
@@ -72,8 +75,8 @@ export class BrowserConsole {
   };
 
   private enforceMessagesLimit() {
-    if (this.messages.length > defaultMessageLimit) {
-      this.messages.splice(0, 1);
+    if (this.messages.length > DEFAULT_MSG_LIMIt) {
+      this.messages = this.messages.splice(0, 1);
```
shahzad31, line 79:
```suggestion
      this.messages.splice(0, 1);
```

src/plugins/browser-console.ts
```diff
@@ -87,6 +90,34 @@ export class BrowserConsole {
     this.driver.context.off('console', this.consoleEventListener);
     this.driver.page.off('pageerror', this.pageErrorEventListener);
     log(`Plugins: stopped collecting console events`);
+
```
shahzad31, line 93:
```suggestion
```

vigneshshanmugam (approved):

PR #818, opened 2023-08-17
fix: use correct type for kibana version

shahzad31:
Kibana api returns version as string

`"version":{"number":"8.7.1",`


devcorpio (approved):
LGTM

PR #819, opened 2023-08-22
feat: added screenshot info in journey payload

shahzad31:
Journey payload will contain if it has screenshots and also it doesn't make any sense to record screenshot if page is never visited. 

This will be useful for UI to not load screenshots in journey if there are none.


src/core/runner.ts
```diff
@@ -237,7 +237,11 @@ export default class Runner {
       const page = pages[pages.length - 1];
       if (page) {
         data.url ??= page.url();
-        if (screenshots && screenshots !== 'off') {
+        if (
+          screenshots &&
+          screenshots !== 'off' &&
+          data.url !== 'about:blank'
```
devcorpio, line 243: a few questions:

- Why did you add `about:blank`? Have you reproduced any bugs?
- Although I'm assuming that you don't want to screenshot "about:blank" pages.
-  Does it make sense to expect customers to have any scenario where they want to open **about:blank** for any reason? In that case, maybe keep doing the screenshot it might make sense since that is their choice. I understand that seems a weird scenario, I'm just sharing a thought.

Either if the "about:blank" is because of a bug or because of the sake of efficiency/performance improvement, I would add a comment saying the why. Otherwise, it might be a source of questions in the future (for us and for future contributors) 

A test could also be a good way to document this

devcorpio: I read again your PR's description, perhaps I would add the comment adding "why" based on that comment.

__tests__/reporters/__snapshots__/json.test.ts.snap
```diff
@@ -210,6 +210,6 @@ exports[`json reporter writes each step as NDJSON to the FD 1`] = `
...
 {"type":"journey/browserconsole","@timestamp":1600300800000000,"journey":{"name":"j1","id":"j1"},"step":{"name":"step-name","index":0},"root_fields":{"os":{"platform":"darwin"},"package":{"name":"@elastic/synthetics","version":"0.0.1"}},"payload":{"text":"Boom","type":"error"},"error":{"name":"Error","message":"boom","stack":""},"package_version":"0.0.1"}
-{"type":"journey/end","@timestamp":1600300800000000,"journey":{"name":"j1","id":"j1","status":"succeeded"},"root_fields":{"os":{"platform":"darwin"},"package":{"name":"@elastic/synthetics","version":"0.0.1"}},"payload":{"start":0,"end":11,"status":"succeeded"},"package_version":"0.0.1"}
+{"type":"journey/end","@timestamp":1600300800000000,"journey":{"name":"j1","id":"j1","status":"succeeded"},"root_fields":{"os":{"platform":"darwin"},"package":{"name":"@elastic/synthetics","version":"0.0.1"}},"payload":{"start":0,"end":11,"status":"succeeded","has_screenshots":false},"package_version":"0.0.1"}
```
emilioalvap, line 213: Do we already have an issue to process these on HB side?

shahzad31: isn't whole payload get's processed as is from what i remember?

emilioalvap: Specifically on HB, yes. I'm more worried about integration mappings, I think we'll need to update those

vigneshshanmugam (commented):
I am not sure why we need this `has_screenshots` check. 

Wouldnt it be possible to extract this on the UI from the journey results? 
```ts
const hasScreenshots = screenshots === "on" || (screesnshots === "only-on-failure" && journey.status === "failed")
```

__tests__/reporters/__snapshots__/json.test.ts.snap
shahzad31: we don't need mapping for this field, summary document will be fetched and on client side you can see the field as part of document

vigneshshanmugam:
Closing this PR as we would like to solve the `No Screenshots problem` in a different way for API only tests.

PR #823, opened 2023-08-29
feat: read config file async

shahzad31:
Add ability to declare config file as async 

```
export default async env => {
  const params = await loadParams();
    return config;
  }
};
```

Should work as well.

### DIFF
--- a/__tests__/config.test.ts
+++ b/__tests__/config.test.ts
@@ -33,7 +33,7 @@ const FIXTURES_DIR = join(__dirname, 'fixtures');
 describe('Config', () => {
   it('read config based on environment', async () => {
     const configPath = join(FIXTURES_DIR, 'synthetics.config.ts');
-    expect(readConfig('development', configPath)).toMatchObject({
+    expect(await readConfig('development', configPath)).toMatchObject({
       params: {
         url: 'dev',
       },
@@ -41,7 +41,7 @@ describe('Config', () => {
         ...devices['Galaxy S9+'],
       },
     });
-    expect(readConfig('testing', configPath)).toMatchObject({
+    expect(await readConfig('testing', configPath)).toMatchObject({
       params: {
         url: 'non-dev',
       },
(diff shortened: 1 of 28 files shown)


__tests__/config.test.ts
```diff
@@ -50,15 +50,19 @@ describe('Config', () => {
...
     const tempPath = generateTempPath();
-    function getConfig() {
-      readConfig('development', tempPath);
+    async function getConfig() {
+      await readConfig('development', tempPath);
+    }
+    try {
+      await getConfig();
+    } catch (e) {
+      expect(e).toEqual(
```
vigneshshanmugam, line 59: we need to use .rejects to make sure its throwing error - https://jestjs.io/docs/expect#rejects

__tests__/config.test.ts
```diff
@@ -33,15 +33,15 @@ const FIXTURES_DIR = join(__dirname, 'fixtures');
 describe('Config', () => {
   it('read config based on environment', async () => {
     const configPath = join(FIXTURES_DIR, 'synthetics.config.ts');
-    expect(readConfig('development', configPath)).toMatchObject({
+    expect(await readConfig('development', configPath)).toMatchObject({
```
vigneshshanmugam, line 36: We need a new test where fetch these configs remotely and merge it inside params and also catching these errors correctly.

src/config.ts
```diff
@@ -47,6 +50,9 @@ export function readConfig(env: string, config?: string): SyntheticsConfig {
   }
   if (typeof options === 'function') {
     options = options(env);
+    if (options instanceof Promise) {
```
vigneshshanmugam, line 53: We need something like this
```
    options != null
    typeof val.then === 'function' &&
    typeof val.catch === 'function'
```

Promises are transpiled in different ways based on the Native promises and ES6 promises. This would be the safest way to check for Promise like function.

src/options.ts
```diff
@@ -85,7 +85,7 @@ export function normalizeOptions(
    */
   const config =
     cliArgs.config || !cliArgs.inline
-      ? readConfig(options.environment, cliArgs.config)
+      ? await readConfig(options.environment, cliArgs.config)
```
vigneshshanmugam, line 88: Needs a test and report errors correctly when journeys are run/pushed.

vigneshshanmugam (changes requested):

src/common_types.ts
```diff
@@ -43,7 +43,7 @@ export type Location = {
   column: number;
 };
 
-export type Params = Record<string, any>;
+export type Params = Record<string, string | number>;
```
vigneshshanmugam, line 46:
```suggestion
export type Params = Record<string, any>;
```

vigneshshanmugam:
@shahzad31 I pushed few fixes for the code and tests. Please check it once, I will get it merged once I get answer for this one https://github.com/elastic/synthetics/pull/823#discussion_r1317856400 , I am curious if its intended or just for testing. Thanks.

Issue #745, opened 2023-04-24
Add default reporter for `--dry-run` option where none is specified


carlyrichmond:
A user has raised a [discuss issue](https://discuss.elastic.co/t/no-tests-found-when-using-dry-run/330458) concerning the `--dry-run` option where it comes back with `No tests found` despite being run correctly from within the `journeys` folder:

Example usage:

`npx @elastic/synthetics . --dry-run `

Output:

```zsh
npx @elastic/synthetics . --dry-run
No tests found! (435 ms)
```

Internal discussion on the synthetics Slack group has found that a `--reporter` option needs to be specified, such as the `json` reporter:

`npx @elastic/synthetics . --dry-run --reporter json`

To simplify the API, we suggest using the default reporter where the `--reporter` option isn't specified.

ACs:

- `npx @elastic/synthetics . --dry-run ` returns a list of tests without executing them, using the default reporter
- `npx @elastic/synthetics . --dry-run --reporter json` returns a json collection of tests within executing them, using the json reporter
- Update the [docs](https://www.elastic.co/guide/en/observability/current/synthetics-command-reference.html) to include what `--dry-run` does

PR #827, opened 2023-08-31
fix: fix dry run for base reporter

shahzad31:
Fixed dry run mode for base reporter 

fixes https://github.com/elastic/synthetics/issues/745

image

### DIFF
--- a/src/core/runner.ts
+++ b/src/core/runner.ts
@@ -451,11 +451,11 @@ export default class Runner {
       env: options.environment,
       params: options.params,
     });
-    await this.reset();
+    await this.reset(options);
     return result;
   }
 
-  async reset() {
+  async reset(options: RunOptions) {
     this.#currentJourney = null;
     this.journeys = [];
     this.#active = false;
@@ -464,6 +464,6 @@ export default class Runner {
      * the current synthetic agent run
      */
     await rm(CACHE_PATH, { recursive: true, force: true });
-    await this.#reporter?.onEnd?.();
+    await this.#reporter?.onEnd?.(options.dryRun);
   }
 }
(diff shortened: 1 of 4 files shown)


src/core/runner.ts
```diff
@@ -451,11 +451,11 @@ export default class Runner {
       env: options.environment,
       params: options.params,
     });
-    await this.reset();
+    await this.reset(options);
     return result;
   }
 
-  async reset() {
+  async reset(options?: RunOptions) {
```
devcorpio, line 458: [just a point of view]

I don't tend to be a fan of passing args in cascade. For instance, `reset` doesn't need `dry-run` flag, but now it "depends" on it.

The way I see it is if a reporter behaves differently depending on something (in this case the dry-run flag) I would add that something as a reporter option: https://github.com/elastic/synthetics/blob/main/src/reporters/base.ts#L76 and https://github.com/elastic/synthetics/blob/main/src/core/runner.ts#L155, so everything stays in the "reporter realm"

With that:

- no more cascade args passing
- If other methods within a reporter need to behave differently because of a dry-run, no need to change their signature.

devcorpio:
LGTM, although I added a comment, let me know your thoughts.

btw, I tested it and works well 🎉

src/reporters/base.ts
```diff
@@ -66,14 +66,17 @@ function renderDuration(durationMs) {
 export default class BaseReporter implements Reporter {
   stream: SonicBoom;
   fd: number;
+  dryRun?: boolean;
```
devcorpio, line 69: nit: not sure if question mark is required anymore now

devcorpio (approved):
LGTM!

PR #828, opened 2023-09-05
ci: check pr title with commitlint

reakaleek:
## Context

We want to release using semantic release. This change is a preparation to ensure we have conventional commit messages on the main branch.

## Changes

This change creates a workflow to lint the PR title with commitlint when a PR is opened, changes are pushed, or if the title is edited.
That should be sufficient because we are using squash merges where the PR title is used as commit message.

## How to test

Tested whilst changing the event type to `pull_request` in https://github.com/elastic/synthetics/actions/runs/6085290115/job/16509016926?pr=828

### DIFF
--- a/.github/workflows/pr-title-lint.yml
+++ b/.github/workflows/pr-title-lint.yml
@@ -11,14 +11,13 @@ jobs:
   pr-title-lint:
     runs-on: ubuntu-latest
     steps:
+      - uses: actions/checkout@v3
       - uses: actions/setup-node@v3
-
-      - name: Setup commitlint
-        run: |
-          npm install -g @commitlint/cli@17.7.1 @commitlint/config-conventional@17.7.0
-          echo "module.exports = {extends: ['@commitlint/config-conventional']}" > commitlint.config.js
-
+        with:
+          cache: npm
+          node-version-file: .nvmrc
+      - run: npm ci
       - name: Lint PR title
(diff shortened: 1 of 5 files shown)


.github/workflows/pr-title-lint.yml
```diff
@@ -0,0 +1,24 @@
...
+      - edited
+
+jobs:
+  pr-title-lint:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/setup-node@v3
+
+      - name: Setup commitlint
+        run: |
```
v1v, line 17: **NIT**:

What do you think this particular step is added to the https://github.com/elastic/synthetics/blob/main/package.json ?

This could help with running the same command locally and in the CI.

v1v (approved):
LGTM

.github/workflows/pr-title-lint.yml
reakaleek: I was thinking about this too.

I had the assumption that a developer won't we able to lint the PR title locally anyway.

But you are right, it would still make it more debuggable.

reakaleek: https://github.com/elastic/synthetics/pull/828/commits/626208f353fbfde9baf431714c5a21264f627601

v1v (approved):

vigneshshanmugam (approved):
LGTM

PR #834, opened 2023-09-14
feat: upgrade PW to 1.38.0

vigneshshanmugam:
+ Splitting this PR #832 in to PW updrage and followup with weberror change. 
+ PW dropped the `pw/test` package in favor of using an internal `playwright` package which contains all the runner/reporters and other matcher related utils. 
+ PR migrates the `@playwright/test` to `playwright` package that has the expect matcher library which we use on the synthetics agent, Plus we have to depend on the playwright package itself to workaround the resolution issue - https://github.com/microsoft/playwright/blob/3170963f42f3d564c13acbbfdc5514b18b57282a/packages/playwright/src/util.ts#L308 which is at the top level. 
+ This leads to unnessary imports of the PW package but that would be the easiest way to get around the Node.js resolution instead of dealing with other kinds of resolution hacks. 
+ Fool proof solution would be to have the PW package expose these matchers - https://github.com/microsoft/playwright/issues/24501


devcorpio:
The error "Error: Cannot find module 'playwright/lib/transform/esmLoader'" that you can see in the CI (and also locally) can be fixed by installing `playwright@1.38.0` in the "main" package.json. (it's one of the solutions at least, that's why I did locally while I was reviewing previous PR)

__tests__/cli.test.ts
```diff
@@ -24,7 +24,7 @@
  */
 
 import { join } from 'path';
-import { devices } from 'playwright-chromium';
+import { devices } from 'playwright-core';
```
vigneshshanmugam, line 27: PW chromium uses pw-core types except installs the chromium browser.

src/plugins/network.ts
```diff
@@ -93,6 +93,17 @@ export class NetworkManager {
     race.then(() => this._barrierPromises.delete(race));
   }
 
+  private _nullableFrameBarrier(req: Request): Frame | null {
```
vigneshshanmugam, line 96: Changed in PW 1.38 where the error would be thrown when accessing frame that are not created.

package.json
```diff
@@ -50,8 +50,9 @@
     "kleur": "^4.1.5",
     "micromatch": "^4.0.5",
     "pirates": "^4.0.5",
-    "playwright-chromium": "=1.37.0",
-    "playwright-core": "=1.37.0",
+    "playwright": "=1.38.0",
```
devcorpio, line 53: I see that in the end installing playwright in the main package.json has been the way to avoid the rabbit hole related to "cannot find module 'playwright/lib/transformer/esmLoader', I believe is a good choice, the alternative(s) not sure if were worth the effort

devcorpio (approved):
LGTM! 🎉

package.json
vigneshshanmugam: I don't prefer this approach personally given we are including the PW test runner in the final packed code and having to deal with the bundles ourselves. Would have been a lot easier if matcher is exported.

Issue #815, opened 2023-08-04
capture unhandled exceptions across all pages


vigneshshanmugam:
+ Current we capture all errors that happens on the overall browser context, through `context.on("error", () => {})`, But there is no equivalent API on the PW side for capturing the Unhandled exceptions which is important to see all the errors that gets thrown on the list of pages that the Synthetics journeys are run through. 

+ It is not ideal to rely on `page.on("pageerror") events` as the we cannot work with the async nature of events, we have to go through `context.("page") and then page.on("pageerror")`, It would be easier if we can get these events on the Browser context from the PW API. 

```ts
const context = await browser.newContext();

context.on("pageerror", (error, page?) => {
   // error -> unhandled exception
});

```

Relevant issue - https://github.com/microsoft/playwright/issues/24466

PR #835, opened 2023-09-14
feat: capture unhandled rejections across all pages

vigneshshanmugam:
+ fix https://github.com/elastic/synthetics/issues/815
+ Available from Playwright 1.38.0 - PR updates the PW to the latest version and captures the unhandled rejections via the new `weberror` event from browser contest instead of relying on the page level `pageerror` event which does not provide the rejections across multiple pages.
+ Docs - https://playwright.dev/docs/api/class-weberror


devcorpio (approved):
LGTM

PR #842, opened 2023-09-28
chore: upgrade sharp to latest

vigneshshanmugam:
+ Upgrades `sharp` and removes the types package which are already shipped with sharp.


justinkambic (approved):
LGTM!

PR #853, opened 2023-10-13
chore: update undici to latest

vigneshshanmugam:
+ Regular maintenance update. Updates to the latest undici package and fix types.

### DIFF
--- a/src/push/request.ts
+++ b/src/push/request.ts
@@ -57,7 +57,9 @@ export async function sendReqAndHandleError<T>(
   options: APIRequestOptions
 ): Promise<T> {
   const { statusCode, body } = await sendRequest(options);
-  return (await (await handleError(statusCode, options.url, body)).json()) as T;
+  return (
+    await handleError(statusCode, options.url, body)
+  ).json() as Promise<T>;
 }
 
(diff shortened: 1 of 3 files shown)


src/push/request.ts
```diff
@@ -57,9 +57,14 @@ export async function sendReqAndHandleError<T>(
   options: APIRequestOptions
 ): Promise<T> {
   const { statusCode, body } = await sendRequest(options);
-  return await (await handleError(statusCode, options.url, body)).json();
+  return (await (await handleError(statusCode, options.url, body)).json()) as T;
```
justinkambic, line 60:
```suggestion
  return (await handleError(statusCode, options.url, body)).json() as Promise<T>;
```

Could we do this to avoid the double `await`? It is not beautiful.

justinkambic (approved):
LGTM, a small nit

src/push/request.ts
vigneshshanmugam: I have no clue, why it was this way. thanks for spotting. I will push this manually to satifsy the prettier.

justinkambic (approved):

PR #856, opened 2023-10-23
fix: capture non error type in json reporter

vigneshshanmugam:
+ Throwing non-error type in the steps results in weird `String split error` as we were trying to split on `error.message` which is not the correct thing to do. This was already handled in the Base reporter. 
+ PR address the same by capturing the throwed primitive values/objects from the Journey/Steps and capturing them as error.message which Kibana uses to visualize the error when running inside Heartbeat. 

```ts
step('Test', async () => {
  throw 'error';
});


```


emilioalvap (approved):
LGTM

PR #861, opened 2023-10-27
fix: normalize monitor names when bundling

### LINKED ISSUE #860: Having special characters in journey name causes different errors
Error sources vary, sometimes it's local during *bundle* phase, sometimes during `push` to Kibana.

Environment:
- @elastic/synthetics: 1.5.1
-  Kibana: 8.8.1
- ES: 8.8.1

### Local error
Steps to reproduce - set journey name to `https://example.com` and push. Same name causes no errors if added from Kibana UI.
```ts
journey("https://example.com", ({ page }) => {
  monitor.use({
    ...config
  });

  step("Test step", async () => {
    await page.goto("https://example.com");
  });
});
```
```bash
$ npx @elastic/synthetics push

> Pushing monitors for project: xxx
> bundling 1 monitors
[Error: ENOENT: no such file or directory, open '/.../.synthetics/bundles/https:/example.com.zip'] {
  errno: -2,
  code: 'ENOENT',
  syscall: 'open',
  path: '/.../.synthetics/bundles/https:/example.com.zip'
}
```

### Remote error
Steps to reproduce - set journey name to `foo: bar` and push. Interestingly this will cause same error when added from Kibana UI.
```ts
journey("foo: bar", ({ page }) => {
  monitor.use({
    ...config
  });

  step("Test step", async () => {
    await page.goto("https://example.com");
  });
});
```
```bash
$ npx @elastic/synthetics push

> Pushing monitors for project: xxx
> bundling 1 monitors
> Monitor Diff: Added(0) Updated(1) Removed(0) Unchanged(0)
✖ Erroring or updating 1 monitors (4505ms)
   > incomplete explicit mapping pair; a key node is missed; or followed by a non-tabulated empty line at line 3, column 10:
       name: foo: bar
                ^
       Failed to update journey: xxx
```

vigneshshanmugam:
+ fix #860 
+ PR adds normalization technique that deals with path traversal abusal that would eventually lead to doing lookups while encoding bundled journeys when pushing project based browser monitors.

### DIFF
--- a/src/push/monitor.ts
+++ b/src/push/monitor.ts
@@ -37,7 +37,7 @@ import {
   MonitorConfig,
 } from '../dsl/monitor';
 import { PushOptions } from '../common_types';
-import { isParamOptionSupported } from './utils';
+import { isParamOptionSupported, normalizeMonitorName } from './utils';
 
 // Allowed extensions for lightweight monitor files
 const ALLOWED_LW_EXTENSIONS = ['.yml', '.yaml'];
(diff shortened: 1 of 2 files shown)


devcorpio (approved):
LGTM

PR #862, opened 2023-11-01
fix: remove top level errors from pagerror events

vigneshshanmugam:
+ Errors that are captured from the `journey/browserconsole` gets added to the root documents that causes HB to capture them as the Journey error. Even though this is wrong from the HB perspective, These top level erros from the Page are not useful as the error itself contains stack trace to minified code which are not useful for the user when looking at these browser console messages.

+ This PR addresses the issue by removing the `error` field from the docs but still keeping the error.messages and other details which will be shown under the Step console events.


justinkambic (approved):
Code LGTM, I did not test this e2e

PR #868, opened 2023-11-10
chore: keep pw version same as other pw deps

vigneshshanmugam:
+ This was slipped when we did the change last time during the PW upgrade - https://github.com/elastic/synthetics/commit/f75c4705b30f30a2b33b2d20e82b18a65adfa0e6

+ Fixing it so we dont need to do anything separate in the Recorder - https://github.com/elastic/synthetics-recorder/pull/480


justinkambic (approved):
LGTM

PR #869, opened 2023-11-10
fix: upgrade deps to trigger a patch release

vigneshshanmugam:
+ We need to do a patch version bump after the PW mismatch fix to make the release workflow trigger correctly.


justinkambic (approved):
LGTM

PR #873, opened 2023-12-08
feat: introduce kibana interal origin header for serverless requests

justinkambic:
Kibana will reject external requests to `internal` routes by default in Serverless.

We flag our requests as `internal` by passing the `x-elastic-internal-origin` header with requests. This patch includes this header in the requests we send to Kibana, which will allow us to query the project monitor API without making it a documented public API.


vigneshshanmugam (approved):
LGTM, I believe this doesn't not impact the ESS stacks.

justinkambic:
>  I believe this doesn't not impact the ESS stacks.

I did test this against a CFT stack and was able to push projects, so I think we are good.

PR #880, opened 2024-01-04
feat: added journey and step annotations

### LINKED ISSUE #404: Enable journey.skip to skip tests
During development, or when encountering a flaky test in CI, there are currently two ways to skip the test.

The first way is by using the filters option to specify the test that should not be run with a negate glob.

The second way is by commenting out the whole test.

We should provide a more streamlined way of skipping a test, by enabling the `journey.skip`. The `.skip` syntax is familiar to many developers using jest, though admittedly I'm unclear how other testing frameworks may handle this.

shahzad31:
+ fix #404 
Support syntax like

Following annotation have been added

`journey.only` will only run this journey
`journey.skip` will skip this journey
`step.skip`
`step.soft` failure of this step will not abort rest of the journey
`step.only`

`-f --force` if journey is skipped , then while pushing it will error out, but can be ignore with this flag

```
  step.skip('launch application again', async () => {
    expect(page.textContent('body')).toContain('Google');
  });
```

example result 

image

### DIFF
--- a/__tests__/core/runner.test.ts
+++ b/__tests__/core/runner.test.ts
@@ -819,6 +819,29 @@ describe('runner', () => {
       expect(data).toContain('1 passed');
     });
 
+    it('only journey', async () => {
+      runner.addJourney(
+        journey.only('j1', async () => {
+          step('step1', async () => {});
+        })
+      );
+      runner.addJourney(
+        journey('j2', async () => {
+          step('step1', async () => {});
+        })
+      );
+      const result = await runner.run(defaultRunOptions);
+      expect(result).toEqual({
+        j1: {
+          status: 'succeeded',
+          stepResults: [],
+        },
+        j2: {
+          status: 'skipped',
+        },
+      });
+    });
+
     it('skip step', async () => {
       runner.addJourney(
         journey('j1', async () => {
--- a/__tests__/core/runner.test.ts
+++ b/__tests__/core/runner.test.ts
@@ -130,10 +130,23 @@ describe('runner', () => {
       throw error;
     });
     const result = await runner.runJourney(journey, defaultRunOptions);
-    expect(result).toEqual({
-      status: 'failed',
-      error,
-    });
+    expect(result).toMatchInlineSnapshot(`
+      {
+        "error": [Error: Broken step 2],
+        "status": "failed",
+        "steps": [
+          {
+            "status": "succeeded",
+            "url": "about:blank",
+          },
+          {
+            "error": [Error: Broken step 2],
+            "status": "failed",
+            "url": "about:blank",
+          },
+        ],
+      }
+    `);
   });
 
   it('run journey - with hooks', async () => {
@@ -141,9 +154,12 @@ describe('runner', () => {
     journey.addHook('before', noop);
     journey.addHook('after', noop);
     const result = await runner.runJourney(journey, defaultRunOptions);
-    expect(result).toEqual({
-      status: 'succeeded',
-    });
+    expect(result).toMatchInlineSnapshot(`
+      {
+        "status": "succeeded",
+        "steps": [],
+      }
+    `);
   });
 
   it('run journey - failed when hooks errors', async () => {
@@ -154,10 +170,13 @@ describe('runner', () => {
       throw error;
     });
     const result = await runner.runJourney(journey, defaultRunOptions);
-    expect(result).toEqual({
-      status: 'failed',
-      error,
-    });
+    expect(result).toMatchInlineSnapshot(`
+      {
+        "error": [Error: Broken after hook],
+        "status": "failed",
+        "steps": [],
+      }
+    `);
   });
 
   it('run journey - failed on beforeAll', async () => {
@@ -318,9 +337,14 @@ describe('runner', () => {
     const journey = new Journey({ name }, noop);
     runner.addJourney(journey);
     const result = await runner.run(defaultRunOptions);
-    expect(result).toEqual({
-      [name]: { status: 'succeeded' },
-    });
+    expect(result).toMatchInlineSnapshot(`
+      {
+        "test-journey": {
+          "status": "succeeded",
+          "steps": [],
+        },
+      }
+    `);
   });
 
   it('run api - match journey name explicit', async () => {
@@ -331,9 +355,14 @@ describe('runner', () => {
         ...defaultRunOptions,
         match: 'j2',
       })
-    ).toEqual({
-      j2: { status: 'succeeded' },
-    });
+    ).toMatchInlineSnapshot(`
+      {
+        "j2": {
+          "status": "succeeded",
+          "steps": [],
+        },
+      }
+    `);
   });
 
   it('run api - match journey name and tag globs', async () => {
@@ -344,10 +373,18 @@ describe('runner', () => {
         ...defaultRunOptions,
         match: 'j*',
       })
-    ).toEqual({
-      j1: { status: 'succeeded' },
-      tagj2: { status: 'succeeded' },
-    });
+    ).toMatchInlineSnapshot(`
+      {
+        "j1": {
+          "status": "succeeded",
+          "steps": [],
+        },
+        "tagj2": {
+          "status": "succeeded",
+          "steps": [],
+        },
+      }
+    `);
   });
 
   it('run api - prefer tags glob matching', async () => {
@@ -362,11 +399,22 @@ describe('runner', () => {
         tags: ['foo*'],
         match: 'j*',
       })
-    ).toEqual({
-      j1: { status: 'succeeded' },
-      j3: { status: 'succeeded' },
-      j5: { status: 'succeeded' },
-    });
+    ).toMatchInlineSnapshot(`
+      {
+        "j1": {
+          "status": "succeeded",
+          "steps": [],
+        },
+        "j3": {
+          "status": "succeeded",
+          "steps": [],
+        },
+        "j5": {
+          "status": "succeeded",
+          "steps": [],
+        },
+      }
+    `);
   });
 
   it('run api - support multiple tags', async () => {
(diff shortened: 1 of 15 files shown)


paulb-elastic:
@shahzad31 what's the use case for skipping a step like this (as opposed to deleting it, commenting it out)?  Is it to keep the structure as it was before a change (so step n is always step n, even if step n-1 is removed)?  Do you envisage this being a temporary thing a user would do, or permanent?

shahzad31:
> @shahzad31 what's the use case for skipping a step like this (as opposed to deleting it, commenting it out)? Is it to keep the structure as it was before a change (so step n is always step n, even if step n-1 is removed)? Do you envisage this being a temporary thing a user would do, or permanent?

@paulb-elastic it's the most familiar way to in testing frameworks to provide an option like this instead of commenting out , it's especially useful when testing journeys locally. It's a very familiar syntax for especially jest developers.

src/core/index.ts
```diff
@@ -57,13 +57,26 @@ export const journey = wrapFnWithLocation(
   }
 );
 
-export const step = wrapFnWithLocation(
+type StepType = (name: string, callback: VoidCallback) => Step;
+
+interface StepWithSkipType extends StepType {
```
vigneshshanmugam, line 62: would move it inside step and name it as `StepWithModifiers` or `StepWithAnnotations`.

src/core/index.ts
```diff
@@ -57,13 +57,26 @@ export const journey = wrapFnWithLocation(
...
+}
+
+export const step: StepWithSkipType = wrapFnWithLocation(
   (location: Location, name: string, callback: VoidCallback) => {
     log(`Step register: ${name}`);
     return runner.currentJourney?.addStep(name, callback, location);
   }
 );
 
+step.skip = wrapFnWithLocation(
```
vigneshshanmugam, line 73: If we are going to resemble PW, it also has support for skipped based on a callback based function. 

```ts
step.skip((env)=> env === "development", 'Skipping dev tests')
```

It would be nice to add support for this as well.

src/dsl/step.ts
```diff
@@ -30,16 +30,19 @@ export class Step {
   index: number;
   callback: VoidCallback;
   location?: Location;
+  isSkipped = false;
```
vigneshshanmugam, line 33: would be simpler to keep this as static variable and use it from the addStep. We might add more mofidiers/annotations later which might increase the args length.

vigneshshanmugam (commented):
Left some comments. 

- What do you think about adding support for journey.skip which is being asked here - https://github.com/elastic/synthetics/issues/404. We need to handle the skipped journeys from being pushed in project based monitor workflow
- Need tests for the skipped steps

src/core/index.ts
shahzad31: let's do that in a follow up

__tests__/core/runner.test.ts
```diff
@@ -793,4 +793,144 @@ describe('runner', () => {
       });
     });
   });
+
+  describe.only('journey and step annotations', () => {
```
vigneshshanmugam, line 797:
```suggestion
  describe('journey and step annotations', () => {
```

__tests__/core/runner.test.ts
```diff
@@ -793,4 +793,144 @@ describe('runner', () => {
...
+
+    it('only journey', async () => {
+      runner.addJourney(
+        journey.only('j1', async () => {
+          step('step1', async () => {});
+        })
+      );
+      runner.addJourney(
+        journey('j2', async () => {
+          step('step1', async () => {});
```
vigneshshanmugam, line 830: use noop in all other tests as well.

__tests__/core/runner.test.ts
```diff
@@ -793,4 +793,144 @@ describe('runner', () => {
...
+        })
+      );
+      await runner.run(defaultRunOptions);
+      const data = getBufferData();
```
vigneshshanmugam, line 817: Would be better to check explicity which ran and which skipped - We could use `readAndCloseStreamJson` and check the json response.

src/cli.ts
```diff
@@ -185,6 +185,7 @@ program
     'the target Kibana spaces for the pushed monitors — spaces help you organise pushed monitors.'
   )
   .option('-y, --yes', 'skip all questions and run non-interactively')
+  .option('-f, --force', 'Ignore warnings and force push')
```
vigneshshanmugam, line 188: I would reuse the `-y`, we already use this for ignoring some of the users changes on local and forcefully pushing the project monitors. Lets not introduce two things to do the same thing.

src/core/index.ts
```diff
@@ -57,13 +63,80 @@ export const journey = wrapFnWithLocation(
   }
 );
 
-export const step = wrapFnWithLocation(
+journey.skip = wrapFnWithLocation(
```
vigneshshanmugam, line 66: Pretty much duplicated. Both of the methods .skip and .only can be invoked by having a common function that does the same thing. 
```ts
func addAnnotation(name: 'only' | 'skip' , args) { 
    const j = new Journey(...args);
    j[name]= true;
    runner.addJourney(j);
}

journey.only = addAnnotation('only')
journey.skip = addAnnotation('skip')

```

src/core/index.ts
```diff
@@ -57,13 +63,80 @@ export const journey = wrapFnWithLocation(
...
+    return runner.currentJourney?.addStep(name, callback, location, {
+      soft: true,
+    });
+  }
+);
+
+/**
+ * Only run this step and skip rest of the steps
+ */
+step.only = wrapFnWithLocation(
```
vigneshshanmugam, line 131: same as the comment above for journey annotation.

src/core/index.ts
```diff
@@ -57,13 +63,80 @@ export const journey = wrapFnWithLocation(
...
+  (location: Location, name: string, callback: VoidCallback) => {
+    log(`Step register: ${name}`);
+    return runner.currentJourney?.addStep(name, callback, location, {
+      only: true,
```
vigneshshanmugam, line 135: Lets try to follow the same notion `step.only = true` like for journey.

src/core/index.ts
```diff
@@ -57,13 +63,80 @@ export const journey = wrapFnWithLocation(
...
+    return runner.currentJourney?.addStep(name, callback, location, {
+      skip: true,
+    });
+  }
+);
+
+/**
+ * Failure of soft step will not skip rest of the steps
+ */
+step.soft = wrapFnWithLocation(
```
vigneshshanmugam, line 119: Been thinking about this today, I am not convinced that this makes sense. Feels like we are trying to work around the `expect.soft` problem. Lets re-explore this in a separate PR. Mainly suggesting because once we have the expect.soft support this becomes useless. 

If the devs want to complete a step without failing, It would be more ideal to either comment out then failing method `expect.blah` or change that call to a soft assertion. I dont think anyone would have a step just for a assertion and would call `.soft` on that whole step instead.

src/core/runner.ts
```diff
@@ -366,6 +368,7 @@ export default class Runner {
           result.error = stepResult.error;
         }
       }
+      result.steps = stepResults;
```
vigneshshanmugam, line 371: I would remove it and test using the custom reporter hooks since its mainly used to make the testing easier.

src/dsl/journey.ts
```diff
@@ -39,6 +39,7 @@ export type JourneyOptions = {
   name: string;
   id?: string;
   tags?: string[];
+  skip?: boolean;
```
vigneshshanmugam, line 42: is this required?

src/core/runner.ts
```diff
@@ -417,6 +420,11 @@ export default class Runner {
       journey.callback({ params: options.params } as any);
       journey.monitor.update(this.monitor?.config);
       journey.monitor.validate();
+      if (journey.skip && !options.force) {
```
vigneshshanmugam, line 423: can be moved up to error out sooner.

src/core/runner.ts
```diff
@@ -417,6 +420,11 @@ export default class Runner {
...
+        throw new Error(
+          `Journey ${journey.name} is skipped. Remove the skip method or use -f, --force to ignore this.`
```
vigneshshanmugam, line 425: It looks like you are trying to push skipped monitors (journey.name), Please remove the .skip annotation and try again.

src/cli.ts
vigneshshanmugam: Thinking through this again. This would end up costing user and doing harm than good. Lets just error in case of skipped journey and figure out a solution if there is a need.

vigneshshanmugam (changes requested):

src/cli.ts
shahzad31: i have remove the -f

src/core/index.ts
shahzad31: done

src/core/runner.ts
shahzad31: this will also enhance creating custom reporter detail reporting.

src/core/index.ts
vigneshshanmugam: Thanks Shahzad, I am really not convinced on the soft usecase for step. Would you suggest some usecase?

A step can error only if there are any expect statements, otherwise it cant unless its a JS bug which users can figure out if they run locally

if users want to soft assert, they can do expect.soft or so if we cannot add support via the PW expect. Lets do it as followup call, unless there is a need for the step.soft. We would be in a better position to add it later than to deprecate it based on what I said above.

src/core/runner.ts
vigneshshanmugam: Sure, makes sense.

vigneshshanmugam (commented):
Had a look today, There are some typings issue and reporter issues. Will push a fix later today.

vigneshshanmugam:
@shahzad31 I pushed a fix that deals with few things
- Multiple .only on journey/steps work properly
- Reporter had issues and was displaying displayed journeys where our counts are based on steps. Ignored journeys are never reported to keep it simpler. 
- Step level only journeys are run even if they are failling. Last error will be added to the journey level error.

vigneshshanmugam (approved):
LGTM

justinkambic (approved):

PR #882, opened 2024-01-05
feat: double monitor bundle size to 1500Kb

vigneshshanmugam:
+ As a interim solution, Increase the recommended bundle size limit to 1500Kb from 800kb. We will look in to minification and other sorts of optimization techniques as a separate step.


shahzad31 (approved):
will also need to follow up with increasing route payload limits.

PR #884, opened 2024-01-06
fix: keep size limit at 1500kb for project monitors

vigneshshanmugam:
+ Fix that was Incorrectly pushed to set the size at 200kb.


emilioalvap (approved):
LGTM

PR #886, opened 2024-01-10
feat: minify whitespaces when bundling monitors

### LINKED ISSUE #883: Add minification support for project based monitors before pushed to Kibana
### Summary

When synthetics agent pushes the journeys/monitors to Kibana, we perform bundling (concatenation) and we gzip the result source before unpacking the code on the Heartbeat side. This is good enough, but to get extra savings, we would like to enable the `minification` for the journey/monitor code to get even more savings. 

### Proposal

When running with minification, we shave off 200Kb from the `pdp-parse` module - https://www.npmjs.com/package/pdf-parse, This would be easier change by enabling the `minification` on the ESbuild side - The bundler we rely on while performing the push operation. 

However minification might break our sourcemaps, We apply sourcemaps extraction and unpacking at different levels. More details are on this PR -  https://github.com/elastic/synthetics/pull/687. The outcome of this minifation should not affect the sourcemaps and we should ensure the step functions, journey scripts, error stacks are not pointing to the original source even after we enable the minify flag.

vigneshshanmugam:
+ fix #883 
+ Updated esbuild to latest `0.19.11` to work with certain modules. 
+ Minifying the bundles turned out to not yield significant savings than minifying just the whitespaces and also required more work on the runner side when reconstructing the source for running the journeys. Wrote detailed information on the original issue - https://github.com/elastic/synthetics/issues/883#issuecomment-1885918654

### DIFF
--- a/__tests__/push/bundler.test.ts
+++ b/__tests__/push/bundler.test.ts
@@ -90,9 +90,8 @@ async function validateZip(content) {
   expect(files).toEqual([partialPath]);
 
   // Verify if files are bundled together
-  expect(contents).toContain('node_modules/is-positive/index.js');
-  expect(contents).toContain('test-bundler/bundle.journey.ts');
-  expect(contents).toContain('test-bundler/utils.ts');
+  expect(contents).toContain('import_is_positive');
+  expect(contents).toContain('utils_default');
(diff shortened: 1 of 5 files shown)


devcorpio (approved):
LGTM

PR #887, opened 2024-01-12
ci(github-action): use hardcoded package name

v1v:
We have agreed to use this approach:

> Given the name of the package never changes and I dont expect it to change anytime in the future. We can replace the value of the .name with @elastic/synthetics since the release is specific to the synthetics package.

This will help with solving the issue defined in [here](https://securitylab.github.com/research/github-actions-untrusted-input/#remediation)


reakaleek (approved):

vigneshshanmugam (approved):
Thanks @v1v

PR #891, opened 2024-02-01
fix(deps): update pw to 1.41.2

vigneshshanmugam:
+ There was a regression which was fixed in 1.41.2, bumping before we do another synthetics release.


justinkambic (approved):
Code LGTM 👍

PR #896, opened 2024-02-20
fix: allow overriding enabled/retestOnFailure via global config

### LINKED ISSUE #894: Global monitor options enabled and retestOnFailure are ignored
Not all global `synthetics.config.ts` settings for [MonitorConfig](https://github.com/elastic/synthetics/blob/main/src/dsl/monitor.ts#L56-L79) get applied to individual monitors.

Example valid `synthetics.config.ts` for cases where by default all monitors (`enabled`), all alerts (`alert`) and retest on failure (`retestOnFailure`) should be disabled:
```ts
import type { SyntheticsConfig } from "@elastic/synthetics";

export default (env: string) => {
    params: {},
    playwrightOptions: {},
    /**
     * Configure global monitor settings
     */
    monitor: {
      schedule: 10,
      locations: [],
      privateLocations: [],
      alert: {
        status: {
          enabled: false,
        },
        tls: {
          enabled: false,
        },
      },
      enabled: false,
      throttling: false,
      retestOnFailure: false,
    },
    /**
     * Project monitors settings
     */
    project: {
      id: "foo",
      url: "https://example.com/",
      space: "default",
    },
  };
  if (env !== "development") {
    /**
     * Override configuration specific to environment
     * Ex: config.params.url = ""
     */
  }
  return config;
};

```

Expected monitor config would be something like this:
```ts
const j1 = new Journey({ name: 'j1', tags: ['foo*'] }, () => {});
j1.updateMonitor({
  id: 'test-j1',
  schedule: 3,
  locations: ['united_kingdom'],
  privateLocations: ['spain'],
});
runner.addJourney(j1);

const monitors = runner.buildMonitors( /* put global settings here */ );
expect(monitors.length).toBe(1);
console.dir(monitors[0].config);
expect(monitors[0].config).toEqual({
  // from local config
  id: 'test-j1',
  name: 'j1',
  type: 'browser',
  tags: ['foo*'],
  locations: ['united_kingdom'],
  privateLocations: ['spain'],
  schedule: 3,
  // Params and PlaywrightOptions, not in scope of this issue
  playwrightOptions: {},
  params: {},
  // from global config
  enabled: false,
  retestOnFailure: false,
  throttling: false,
  alert: { status: { enabled: false }, tls: { enabled: false } },
  screenshot: undefined,
});
```
But instead you're getting `MonitorConfig`:
```ts
{
  throttling: false,
  schedule: 3,
  locations: [ 'united_kingdom' ],
  privateLocations: [ 'spain' ],
  params: {},
  playwrightOptions: {},
  screenshot: undefined,
  tags: [ 'foo*' ],
  alert: { status: { enabled: false }, tls: { enabled: false } },
  name: 'j1',
  id: 'test-j1',
  type: 'browser'
}
```
with `enabled` and `retestOnFailure` settings completely ignored.
Tests for this also seem to be wrong, because [runner.updateMonitor](https://github.com/elastic/synthetics/blob/main/__tests__/core/runner.test.ts#L757) gets executed by test with global settings and `runner.buildMonitors(options)` is called with local settings which is wrong, `runner.buildMonitors(options)` should be called with global settings and `runner.updateMonitor` should not be executed at all, because it's invoked by `runner.buildMonitors` [here](https://github.com/elastic/synthetics/blob/main/src/core/runner.ts#L410).

vigneshshanmugam:
+ fix #894 
+ enabled: false and retestOnFailure: false on the synthetics global config did not get passed to Kibana which effectively meant the users can only enable/disable them via individual monitor config. This PR fixes the problem and allows all monitors to be overridden with correct value.

### DIFF
--- a/__tests__/core/runner.test.ts
+++ b/__tests__/core/runner.test.ts
@@ -754,6 +754,7 @@ describe('runner', () => {
     });
 
     it('runner - build monitors with global config', async () => {
+      // Faking monitor.use across file level
       runner.updateMonitor({
         schedule: 5,
         locations: ['us_east'],
@@ -772,16 +773,22 @@ describe('runner', () => {
         schedule: 3,
         locations: ['united_kingdom'],
         privateLocations: ['spain'],
+        retestOnFailure: true,
       });
-      j2.updateMonitor({ throttling: { latency: 1000 } });
+      j2.updateMonitor({ throttling: { latency: 1000 }, enabled: true });
       runner.addJourney(j1);
       runner.addJourney(j2);
 
-      const monitors = runner.buildMonitors(options);
+      const monitors = runner.buildMonitors({
+        ...options,
+        enabled: false,
+        retestOnFailure: false,
+      });
       expect(monitors.length).toBe(2);
       expect(monitors[0].config).toEqual({
         id: 'test-j1',
         name: 'j1',
+        enabled: false,
         type: 'browser',
         tags: ['foo*'],
         locations: ['united_kingdom'],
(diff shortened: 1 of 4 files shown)


devcorpio (approved):
LGTM

--
**before the change:**

Screenshot before the change


**after the change:**

Screenshot after the change

PR #903, opened 2024-03-11
feat: support yml anchors during push command

### LINKED ISSUE #888: yaml anchor support for lightweight monitors
We have a large list of similar http lightweight monitors, that need e.g. the same checks. To be DRY it would be nice if yaml anchors would be supported to be able to do something like this:

```yaml
defaults: &defaults
  type: http
  schedule: '@every 5m'
  timeout: 60s
  private_locations:
    - some-location-dev
    - some-location-staging
  check.response:
    json:
      - description: is active
        expression: 'status == "active"'
      - description: is CapabilityStatement
        expression: 'resourceType == "CapabilityStatement"'

heartbeat.monitors:
- name: Monitors are awesome
  id: awesomemonitor1
  urls: "https://some-unique-url/endpoint1"
  << : *defaults
- name: Another awesome monitors
  id: awesomemonitor2
  urls: "https://some-other-unique-url/and-different-endpoint"
  << : *defaults

```

vigneshshanmugam:
+ fix #888 
+ YAML anchors help with DIY principle when same multiple monitors have similar structure. When users are maintaining a large number of monitors, this becomes super handy.

### DIFF
--- a/__tests__/push/monitor.test.ts
+++ b/__tests__/push/monitor.test.ts
@@ -365,14 +365,17 @@ heartbeat.monitors:
   schedule: '@every 1m'
   <<: *http-type
   <<: *team-tag
+  urls: "https://blog.elastic.co"
 - id: http2
   name: "http2"
   <<: *http-type
   <<: *schedule-5
+  urls: "https://elastic.co"
 - id: tcp1
   name: "tcp1"
   type: tcp
   <<: *schedule-5
+  hosts: ["elastic.co:443"]
       `);
       const [mon1, mon2, mon3] = await createLightweightMonitors(
         PROJECT_DIR,
@@ -384,18 +387,21 @@ heartbeat.monitors:
         schedule: 1,
         type: 'http',
         tags: ['team1', 'team2'],
+        urls: 'https://blog.elastic.co',
       });
       expect(mon2.config).toMatchObject({
         id: 'http2',
         name: 'http2',
         schedule: 5,
         type: 'http',
+        urls: 'https://elastic.co',
       });
       expect(mon3.config).toMatchObject({
         id: 'tcp1',
         name: 'tcp1',
         schedule: 5,
         type: 'tcp',
+        hosts: ['elastic.co:443'],
       });
     });
   });
--- a/__tests__/push/monitor.test.ts
+++ b/__tests__/push/monitor.test.ts
@@ -347,6 +347,57 @@ heartbeat.monitors:
         },
       });
     });
+
+    it('support anchors', async () => {
+      await writeHBFile(`
+http-type: &http-type
+  type: http
+schedule5: &schedule-5
+  schedule: '@every 5m'
+team-tag: &team-tag
+  tags:
+    - team1
+    - team2
+
+heartbeat.monitors:
+- id: http1
+  name: "http1"
+  schedule: '@every 1m'
+  <<: *http-type
+  <<: *team-tag
+- id: http2
+  name: "http2"
+  <<: *http-type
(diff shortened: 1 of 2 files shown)


devcorpio (approved):
code LGTM

Edit: I will test it manually and let you know

devcorpio:
TL;DR: testing panned out 🚀, so LGTM 🎉 

btw, at the end you will see an observations section I added, let me know what you think:


## Details:

YML used:

```
http-type: &http-type
  type: http
schedule5: &schedule-1
  schedule: '@every 1m'
team-tag: &team-tag
  tags:
    - team1
    - team2

heartbeat.monitors:
- id: http1
  name: "http1"
  schedule: '@every 1m'
  <<: *http-type
  <<: *team-tag
  urls: "https://elastic.github.io/synthetics-demo/"
- id: http2
  name: "http2"
  <<: *http-type
  <<: *schedule-1
  urls: "https://elastic.co"
- id: tcp1
  name: "tcp1"
  type: tcp
  <<: *schedule-1
  hosts: ["www.elastic.co:443"]
```


**case 1: results of using the YML anchor in a branch without the new code changes**

Screenshot 2024-03-14 at 12 08 32


I tested out of curiosity. Above, you can see what the CLI displayed.

The interesting thing is that in Kibana I was able to see two monitors (although they have been created as `journeys` rather than `http`), see below:

Screenshot 2024-03-14 at 12 25 54

**case 2: results of using the YML anchor in the PR's branch**

since http1 and http2 were created as journeys in the previous step. Now we can see the following when executing the command:

Screenshot 2024-03-14 at 12 28 25

This is a good signal since that tells us the new logic is involved 🚀 

Now, after removing all monitors and starting afresh, the CLI shows this:

Screenshot 2024-03-14 at 12 31 18

Besides, on this Kibana page we can see clearly how things have been set as expected:

Screenshot 2024-03-14 at 12 24 02

I also tested pushing the same YML again (without any changes):

Screenshot 2024-03-14 at 12 34 09

all good, too.

I also tested pushing the YML with a little variation (changing the URL of the TCP monitor):

Screenshot 2024-03-14 at 12 35 22

everything is okay as well.

## OBSERVATIONS

1. Perhaps not super important, but when writing the YML, VS code was complaining about duplicated keys:

Screenshot 2024-03-14 at 12 39 27

Display related to:

Screenshot 2024-03-14 at 12 39 58

Should we document this or mention it somehow? Maybe it's not important I'm not sure, but wanted to highlight it. It doesn't affect the functionality at all.


2. uploading a TCP monitor with a host that is "malformed" such as "https://elastic.co" (it should be elastic.co:443) makes the monitor fail, which make sense, but what caught my attention is that in the UI it says that the URL is unavailable:

Screenshot 2024-03-14 at 12 45 22

which seems weird, at the beginning I thought that the YML wasn't working and that the URL was empty.

But after updating to the proper value (elastic.co:443), I started seeing the URL set again:

Screenshot 2024-03-14 at 12 49 32

again, just something that wanted to highlight, perhaps is an intended behaviour


3. I will update the unit test, although it works, I believe that we should use all the required fields for each monitor type. For instance, `urls` and `hosts`, saying mainly because tests can also be a good place where to extract examples from, just a nitpick thingy 😄

devcorpio (approved):

vigneshshanmugam:
@devcorpio Thanks for the through tests 🎉 

Regarding the VScode thing, we really don't control that and its the YML formatting/extension that throws these lint errors. We can discard them as its up to the library to handle these.

emilioalvap (approved):
LGTM

PR #904, opened 2024-03-19
fix: redact sensitive req/response headers by default

### LINKED ISSUE #468: Support suppressing of sensitive parameters / URLs
In some situations users must test an app that has sensitive URLs. While putting sensitive data in URLs is generally considered a bad idea (for a list of the very narrow exceptions see [this blog post](https://neilmadden.blog/2019/01/16/can-you-ever-safely-include-credentials-in-a-url/), it would make sense to suppress portions of sensitive URLs.

I'm not sure what the cleanest way to do this would be in terms of an API, maybe others could suggest an approach?

vigneshshanmugam:
+ fix #468 
+ Redacts some of the well known sensitive headers and other Access token related keys that might be accidentally captured by the agent when capturing network information. 
+ Cookies are already filtered out by Playwright, But we still get rid of Cookie and Set cookie for the request/response cycle to be on safer side.

### DIFF
--- a/src/reporters/json.ts
+++ b/src/reporters/json.ts
@@ -167,9 +167,10 @@ const SANITIZE_HEADER_KEYS = [
   /^.*token.*$/i,
   /^.*session.*$/i,
   /^.*auth.*$/i,
+  /^cookie$/i,
   /^set\x2dcookie$/i,
 ];
-export function redactKeys(obj: Record<string, string>) {
+export function redactKeys(obj: Record<string, string> = {}) {
   const result = {};
   for (const key of Object.keys(obj)) {
     const shouldRedact = SANITIZE_HEADER_KEYS.some(regex => regex.test(key));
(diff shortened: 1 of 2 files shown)


devcorpio:
LGTM 🎉 

Testing went well:

#### before the PR changes screenshot
before changes


#### after the PR changes screenshot

after changes

src/reporters/json.ts
```diff
@@ -154,8 +154,41 @@ function formatTLS(tls: SecurityDetails) {
...
 }
 
+/**
+ * List of wildcard header keys that should be sanitized/redacted from the request/response headers
+ *
+ * Spec followed by APM agent is used as a reference.
+ * https://github.com/elastic/apm/blob/36a5abd49ff156c80cf0c9e2e1eac919873cb18b/specs/agents/sanitization.md?plain=1#L27
+ */
+const SANITIZE_HEADER_KEYS = [
+  /^password$/i,
```
devcorpio, line 164: [question]

Is there any reason for not adding all the "default" keys available in the APM document?

default keys

devcorpio (approved):

src/reporters/json.ts
vigneshshanmugam: Mainly because it was an exhaustive set of list and we can introduce these extra headers like credit, card etc incase its required. Also frontend headers differ a lot between the backend transactions which might involve order processing system and other services which makes sense for backend agents.

Issue #906, opened 2024-03-27
Hide API key during CLI init dialogue  and optionally export to env, combine Cloud ID and Kibana URL prompt


justinkambic:
When the user creates a new project, Synthetics will prompt them for their API key.

Today, this requires pasting the API key (and including it in plaintext). This looks like:

```shell
➜ npx @elastic/synthetics init
> Initializing Synthetics project in '.'
✔ Do you use Elastic Cloud (Y/n) · false
✔ What is the url of your Kibana instance · http://127.0.0.1:5601
✔ What is your API key · <blah>
```

After the project scaffolding is created, `npm run push` will not work unless the user also exports `SYNTHETICS_API_KEY` or includes it as an `env` arg when running the command.

For tasks like demoing Synthetics or running this setup in any kind of communal way, this requires showing the API key in plaintext. It would be better if, when the user pasted their API key, the prompt did not display it (which is standard for typing/pasting passwords in CLI prompts).

It would also be a useful addition to have the option to have Synthetics export the API key to the path at the end of the dialogue so the `npm run push` command will work straight away without additional configuration. This is something I, as a frequent user, forget to do plenty of the time and is probably confusing for users.

Additionally, we should combine the question about the user's Cloud ID and their Kibana URL into a single question and translate the Cloud ID to a Kibana URL for them. This lets us skip a question and streamlines the flow.

vigneshshanmugam:
I am generally against the idea of changing the Environment variables via scripts unless we know absolutely what we are doing. 

In the synthetics case doing so would make it harder to debug issues if the user has multiple projects setup that requires different authkeys based on the project settings they are pushing. But I also get the sense, this is really frustrating that even after giving the APIKey, we still expect users to pass it when doing `npm run push`. To fix that, I propose we improve the push command generation and add the `--auth <value>` at the end of project scaffolding and so the `npm run push` command will work straight away without additional configuration

vigneshshanmugam:
`--auth` idea wont work as we it would write the APiKey in plain text on the package.json script which is not good from the security perspective. 

Only thing we could do is to change the documentation and mention the users to export it themselves to be in safer side.

vigneshshanmugam:
The best solution to solve both of the concerns raised is - https://github.com/elastic/synthetics/issues/603, Which works for both multiple projects and easier onboarding.

Issue #907, opened 2024-03-27
Synthetics `init` command takes args to skip dialogue steps


justinkambic:
Today, when creating a new project via `npx @elastic/synthetics init`, I am given a number of prompts. This includes my Kibana API key, locations, monitor duration, Kibana spaces, etc.

I would like to skip these and get right to managing my project.

The `init` command should allow arguments for these fields, so if we create a project quickstart flow in Kibana we could publish a pre-filled command like `npx @elastic/synthetics init --apiKey {...} --kibanaUrl http://127.0.0.1:5601/ --location us-east --location us-west --period 5 --kibanaSpace default --exportApiKey yes` that the user could then paste to their shell and skip the dialogue altogether.

With a feature like this, they could either `npm run push` straight away, or make modifications as needed and push afterwards.

We only really require API Key, Kibana URL, and Location input from the user, all the other fields are default-able, so if they're unspecified we can simply supply the default with this usage. For no arguments, we could use the existing wizard dialogue instead.

justinkambic:
An additional nice feature would be, if the API key is unspecified but `SYNTHETICS_API_KEY` is defined in the env we could optionally pick it up without displaying, but I'm not sure if that'd be confusing.

borism:
Hi, we also would be quite interested in this feature so we can use it to automatically onboard new projects via Github action in our Synthetics-as-a-code monorepo. Doesn't seem to work like this yet in 1.9.0.

shahzad31:
Related multi-env `init` scaffold: #1182 (parent #1175).

PR #911, opened 2024-03-29
feat: improve project scaffolding experience

vigneshshanmugam:
+ fix https://github.com/elastic/synthetics/issues/906
+ Part of https://github.com/elastic/synthetics/issues/907
+ Adds support for validating both the Kibana URL or cloud.id and adds validation step if the given URL is invalid. 
+ Hides API key on the CLI prompt and adds the API Key to the push command

### DIFF
--- a/src/generator/index.ts
+++ b/src/generator/index.ts
@@ -287,12 +287,10 @@ All set, you can run below commands inside: ${this.projectDir}:
 
   Run synthetic tests: ${cyan(runCommand(this.pkgManager, 'test'))}
 
-  Push monitors to Kibana: ${cyan(
-    'SYNTHETICS_API_KEY=<value> ' + runCommand(this.pkgManager, 'push')
-  )}
+  Push monitors to Kibana: ${cyan(runCommand(this.pkgManager, 'push'))}
 
   ${yellow(
-    'Make sure to configure the SYNTHETICS_API_KEY before pushing monitors to Kibana.'
+    'Modify the API Key via `SYNTHETICS_API_KEY` env variable or --auth CLI flag.'
   )}
 
 Visit https://www.elastic.co/guide/en/observability/current/synthetic-run-tests.html to learn more.
--- a/src/generator/index.ts
+++ b/src/generator/index.ts
@@ -45,7 +45,6 @@ import type { ProjectSettings } from '../common_types';
 const templateDir = join(__dirname, '..', '..', 'templates');
 
 type PromptOptions = ProjectSettings & {
-  auth: string;
   locations: Array<string>;
   privateLocations: Array<string>;
   schedule: number;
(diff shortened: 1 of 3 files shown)


src/generator/index.ts
```diff
@@ -80,29 +83,30 @@ export class Generator {
...
-        ? 'What is your cloud.id'
-        : 'What is the url of your Kibana instance',
       name: 'url',
+      message: 'Enter Elastic Kibana URL or Cloud ID',
       required: true,
-      result(value) {
-        return onCloud ? cloudIDToKibanaURL(value) : value;
+      validate(value) {
+        if (!IS_URL.test(value)) {
+          value = cloudIDToKibanaURL(value);
```
emilioalvap, line 92: Can we include this statement in the try/catch statement? `cloudIDToKibanaURL` is likely to throw if gibberish is passed

vigneshshanmugam: Good point.

src/generator/index.ts
```diff
@@ -80,29 +83,30 @@ export class Generator {
...
-        return onCloud ? cloudIDToKibanaURL(value) : value;
+      validate(value) {
+        try {
+          if (!IS_URL.test(value)) {
+            value = cloudIDToKibanaURL(value);
+          }
+          new URL(value);
+          return true;
+        } catch (e) {
+          return 'Invalid URL or Cloud ID';
```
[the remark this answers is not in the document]
vigneshshanmugam: RegExp was wrong, I forgot to push it. Thanks for catching it.

src/generator/index.ts
justinkambic: Same train of thought here, the UX is kind of bad in the case you enter garbage. For instance, this is what I see:

```
➜ npx @elastic/synthetics init leafeon
> Initializing Synthetics project in 'leafeon'
? Enter Elastic Kibana URL or Cloud ID › garbage  

❯ Invalid URL or Cloud ID
```

Once this happens, all of my keystrokes are just met with a beep and I can't try again. I had to kill the process, which is actually more annoying than if it just blew up and crashed on its own. How feasible is it to let us try again after seeing this error?

justinkambic (approved):
Left one comment on the UX. It'd be nice to improve as part of this patch, but if it can't be accomplished quickly I am ok if we don't block this, and I can update the issue to reflect the need for a future patch.

Functionality LGTM and I tested this pushing to a cloud project via cloud ID and URL methods, both worked well.

src/generator/index.ts
vigneshshanmugam: Thats weird, I was able to use backspace/delete and re-enter again to go to the next step.

vigneshshanmugam: Please try again and let me know how it goes, Happy to remove the validation later if you still get hit by this.

justinkambic: Just an error on my part, everything is good here!

Issue #910, opened 2024-03-28
Nice to have features for Control Plane team


gsantoro:
1. During the bootstrap phase the `init` command asks to provide the Elastic cloud Id and the Kibana URL. I believe the only thing required is the Kibana URL. Removing the Cloud ID requirement could open up the possibility of using serverless clusters. Which could be a big requirement for the Control Plane team - https://github.com/elastic/synthetics/pull/911
2. Update docs #909 
3. Show which monitors were deleted, added, ... (any state change) #803 
4. Ask for confirmation when deleting monitors/journeys. Confirmation is only asked if deletions are the only actions to be performed. If there are additions or just changes, no confirmation is required. This is partially covered by #778
5. Group journeys using the same tag and then only push the journeys with those tags. This would partially revert #797. The reasoning for this is that most of the journeys at https://github.com/elastic/synthetics-monitors/tree/main/qa/qa_serverless_monitors/journeys are exactly the same as https://github.com/elastic/synthetics-monitors/tree/main/production/production_serverless_monitors/journeys. The only difference is the URL for Kibana and other params which are already in a separate config file. If I were able to only push journeys by tags I would expect to be able to run the following commands for qa and production environments

```
npm run push -- --config qa.synthetics.config.ts --tags env:qa
npm run push -- --config synthetics.config.ts --tags env:prod
```

those commands could be put in a package.json like

```
{
  "name": "cp-synthetics-lab",
  "version": "1.0.0",
  "description": "",
  "main": "index.js",
  "scripts": {
    "push": "npx @elastic/synthetics push",
    "qa:push": "npm run push -- --config qa.synthetics.config.ts --tags env:qa",
    "prod:push": "npm run push -- --config synthetics.config.ts --tags env:prod"
  },
  "keywords": [],
  "author": "",
  "license": "ISC",
  "dependencies": {
    "@elastic/synthetics": "^1.8.0"
  }
}
```

so that pushing all tests for the qa environment could be as easy as

```
npm run qa:push
npm run prod:push
```

Journeys will need to be tagged as this one

```
journey(
  {
    name: "httpbin - Check Response Code",
    tags: ["httpbin", "env:qa"],
    id: "serverless-httpbin-response-code",
  },
  ({ page, params }) => {
    checkResponseCode(page, params.httpBinUrl);
  }
);
```

PR #913, opened 2024-04-03
feat: support tags/match filtering in push

vigneshshanmugam:
+ Part of https://github.com/elastic/synthetics/issues/910
+ Filtering in push via tags/match has been asked a lot both externally and internally as it makes it easier for the users to control the monitors getting pushed to Kibana. We have [intentionally made the decision](https://github.com/elastic/synthetics/issues/780) to not support it, mainly because monitors were deleted behind the scenes when users pushed monitors with different tags. We [improved deletion flow](https://github.com/elastic/synthetics/pull/912) and requires explicit approval from the user before all the monitors gets deleted, Also the monitors deleted are shown to the user if `DEBUG=synthetics` flag is enabled. 
+ PR adds support for filtering monitors via `--tags and --match` CLI flags in addition to the undocumented `--pattern` field. It behaves similar to the journey test runner which is used for filtering the journeys when running locally. 

```sh
// filter monitor names 
npx @elastic/synthetics --match "foo*"

// filter monitor with tags

npx @elastic/synthetics --tags "env:qa, env:staging"
```

+ Users can tag the monitors/journeys in different ways and its still supported and backwards compatible. 
```ts
// tag on the journey level
journey({name: "using journey param", tags: ["env:qa"], async(){})

// tag using monitor.use API
journey("using monitor API", async(){
  monitor.use({ tags: ["env:qa"]})
  step("blah", () => {})
})

// global tags via Synthetics.config.ts
export default () => {
  return {
    monitor: {
      tags: ["env:dev"]
    }
  }
};
```

Lightweight monitors
```yaml
heartbeat.monitors:
  - type: http
    enabled: true
    name: google.com
    id: google-1
    urls: ["google.com."]
    tags: ["env:qa"]
```

### DIFF
--- a/__tests__/core/runner.test.ts
+++ b/__tests__/core/runner.test.ts
@@ -338,7 +338,7 @@ describe('runner', () => {
     expect(
       await runner.run({
         ...defaultRunOptions,
-        filter: { match: 'j2' },
+        grepOpts: { match: 'j2' },
       })
     ).toMatchObject({
       j2: { status: 'succeeded' },
@@ -351,7 +351,7 @@ describe('runner', () => {
     expect(
       await runner.run({
         ...defaultRunOptions,
-        filter: { match: 'j*' },
+        grepOpts: { match: 'j*' },
       })
     ).toMatchObject({
       j1: { status: 'succeeded' },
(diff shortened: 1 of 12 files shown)


ale-neto:
nice bro, I'm currently going through a problem, and I believe this might be the solution.

gsantoro:
It works as expected!

Thanks

devcorpio (approved):
LGTM

tested both lightweight and browser.

Example of what the user will see if a deletion is triggered. I also set DEBUG=synthetics

Screenshot


**Note for users**: monitors created manually (this is to say, not using [Project monitors](https://www.elastic.co/guide/en/observability/current/synthetics-get-started-project.html)) will not be deleted if their tags don't match the ones used in the push command. The reason for this is that they are independent monitors.

PR #914, opened 2024-04-05
docs: align cli flags with synthetics docs

vigneshshanmugam:
+ Following some of the changes here - https://github.com/elastic/observability-docs/pull/3728
+ fix https://github.com/elastic/synthetics/issues/910

### DIFF
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -80,11 +80,14 @@ program
       Object.keys(reporters)
     )
   )
-  .option('--inline', 'Run inline journeys from heartbeat')
+  .option('--inline', 'read journeys from stdin instead of reading from files')
   .option('-r, --require <modules...>', 'module(s) to preload')
   .option('--sandbox', 'enable chromium sand-boxing')
-  .option('--rich-events', 'Mimics a heartbeat run')
-  .option('--no-headless', 'run browser in head-full mode')
+  .option(
+    '--rich-events',
+    'preset flag used when running monitors directly via Heartbeat'
+  )
+  .option('--no-headless', 'run with the browser in head-full mode')
   .option(
     '--capability <features...>',
     'Enable capabilities through feature flags'
@@ -92,7 +95,7 @@ program
   .addOption(
     new Option(
       '--screenshots [flag]',
-      'take screenshots at end of each step'
+      'Control whether or not to capture screenshots at the end of each step'
     ).choices(['on', 'off', 'only-on-failure'])
   )
   .option(
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -87,7 +87,7 @@ program
     '--rich-events',
     'preset flag used when running monitors directly via Heartbeat'
   )
-  .option('--no-headless', 'run with the browser in head-full mode')
+  .option('--no-headless', 'run with the browser in headful mode')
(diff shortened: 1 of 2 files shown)


src/cli.ts
```diff
@@ -80,19 +80,22 @@ program
...
+  .option('--inline', 'read journeys from stdin instead of reading from files')
   .option('-r, --require <modules...>', 'module(s) to preload')
   .option('--sandbox', 'enable chromium sand-boxing')
-  .option('--rich-events', 'Mimics a heartbeat run')
-  .option('--no-headless', 'run browser in head-full mode')
+  .option(
+    '--rich-events',
+    'preset flag used when running monitors directly via Heartbeat'
+  )
+  .option('--no-headless', 'run with the browser in head-full mode')
```
justinkambic, line 90:
```suggestion
  .option('--no-headless', 'run with the browser in headful mode')
```

This is really jargon-y but I think `headful` is an accepted term for this meaning. I realize it deviates from what we had before but I think this is allowed and nicer.

src/cli.ts
```diff
@@ -80,19 +80,22 @@ program
...
   .option(
     '--capability <features...>',
     'Enable capabilities through feature flags'
   )
   .addOption(
     new Option(
       '--screenshots [flag]',
-      'take screenshots at end of each step'
+      'Control whether or not to capture screenshots at the end of each step'
```
justinkambic, line 98:
```suggestion
      'Control whether to capture screenshots at the end of each step'
```

`whether` does all the heavy lifting, no need to add more conditional wording.

src/options.ts
```diff
@@ -214,16 +214,16 @@ export function getCommonCommandOpts() {
...
   const configOpt = createOption(
     '-c, --config <path>',
-    'configuration path (default: synthetics.config.js)'
+    'path to the configuration file (default: synthetics.config.(js|ts))'
   );
 
   const tags = createOption(
     '--tags <name...>',
-    'run/push tests with a tag that matches the glob'
+    'run/push tests with the tag(s) that matches a pattern'
```
justinkambic, line 222:
```suggestion
    'run/push tests with the tag(s) matching a pattern'
```

IMO this reads better because it is correct for `tag|tags`.

src/options.ts
```diff
@@ -214,16 +214,16 @@ export function getCommonCommandOpts() {
...
   );
   const match = createOption(
     '--match <name>',
-    'run/push tests with a name or tags that matches the glob'
+    'run/push tests with a name or tags that matches a pattern'
```
justinkambic, line 226: As a developer I personally find the usage of `glob` to be more specific and helpful than `pattern`, but doesn't matter too much I guess.

justinkambic (approved):
I didn't do any practical testing in this review, as it looks like the flags themselves are unchanged. I made some copy recommendations. I'm not a technical writer, these are all optional. If revision to the docs is also required I'm fine with skipping the recommended changes.

LGTM.

src/options.ts
vigneshshanmugam: ++ I mainly changed it glob felt more technical than pattern?

justinkambic (approved):

PR #921, opened 2024-04-29
fix: avoid consuming kibana response twice

### LINKED ISSUE #920: TypeError: unusable when creating a new project
Hello,

I'm unable to create a new project. When running `npx @elastic/synthetics init projects-test`, after entering the Kibana URL and the API key I get the error `TypeError: unusable`. The API key is for private locations, Elasticsearch and Kibana are self-hosted and behind a VPN but I'm connected to the VPN.

Elasticsearch version: 8.11.3
Kibana version: 8.11.3
Synthetics version: 1.9.0
Node version: 21.7.3

vigneshshanmugam:
+ fix #920 
+ Underlying error was triggered from Undici HTTP client when we consume the Response body twice during the error scenario, This hides the error message and throws the default TypeError instead. Copying docs from undici
```
Note: Once a mixin has been called then the body cannot be reused, thus calling additional mixins on .body, e.g. .body.json(); .body.text() will result in an error TypeError: unusable being thrown and returned through the Promise rejection.


```

### DIFF
--- a/src/push/request.ts
+++ b/src/push/request.ts
@@ -80,11 +80,11 @@ export async function handleError(
     throw formatNotFoundError(url, await body.text());
   } else if (!ok(statusCode)) {
     let parsed: APIError;
-    const resp = await body.text();
     try {
+      const resp = await body.text();
       parsed = JSON.parse(resp) as APIError;
     } catch (e) {
-      throw formatAPIError(statusCode, 'unexpected non-JSON error', resp);
+      throw formatAPIError(statusCode, 'unexpected error', e.message);
     }
     throw formatAPIError(statusCode, parsed.error, parsed.message);
   }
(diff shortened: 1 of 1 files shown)


src/push/request.ts
```diff
@@ -80,14 +80,11 @@ export async function handleError(
     throw formatNotFoundError(url, await body.text());
   } else if (!ok(statusCode)) {
     let parsed: APIError;
+    const resp = await body.text();
     try {
```
emilioalvap, line 84:
```suggestion

    const resp = await body.text();

```

emilioalvap: We should guard against `body.text()` rejections

vigneshshanmugam: Good catch. I will move the error message as generic as well.

emilioalvap (approved):
LGTM

Issue #917, opened 2024-04-19
Typescript definitions not available for expect


ivosh:
Dear,

Please could you fix the following line in 
https://github.com/elastic/synthetics/blob/main/src/core/expect.ts#L42

so that the Typescript definitions are actually available for `expect`.
The original line points to a non-existing location.

Original:
```
export const expect: typeof import('../../bundles/node_modules/playwright/types/test').expect =
```

Fixed:
```
export const expect: typeof import('../../node_modules/playwright/types/test').expect =
```

The impact of this issue is that the Typescript definitions for `expect` are not available for IDE, `eslint` etc.
This gives a lot of false warnings causing developers confusion.

vigneshshanmugam:
Thanks for reporting @ivosh, Good find.

vigneshshanmugam:
Closed by #923

ivosh:
@vigneshshanmugam Thank you for fixing this (small) issue and for releasing a fix in a couple of days. I confirm linting now works!

vigneshshanmugam:
@ivosh Thanks again for testing and reporting the issue.

Issue #614, opened 2022-09-26
keep Playwright versions up to date on every release


vigneshshanmugam:
+ Dependabot was recently disabled due to the noise it generated when upgrading the dependency versions.Main reason for disabling was since we relied on the exact versions of Playwright. PW version bump itself would need to be mapped to exact versions and has to be tested with HB to ensure they are smooth and safe to be merged.

+ We can reenable dependabot to get notified for the PW releases and also open a PR, which we could work on and manually fix if failing tests to make it workable. 

**NOTE: We may also need to automate the HB testing along with this change as we would require extensive testing with HB images when upgrading PW versions**

vigneshshanmugam:
Add a pre publishing/release script that checks if the PW version is up to date and fail if we are behind. 

In case of API breakage from the PW core, we should have the ability to forcefully release.

paulb-elastic:
@vigneshshanmugam does https://github.com/elastic/synthetics/issues/781 suffice for now, should this be automation we look at in the future?

vigneshshanmugam:
Haven't looked in to this one yet @paulb-elastic, But yeah 781 should do it for this release. But we need some automation around it as its easy to slip.

PR #923, opened 2024-05-01
fix: update pw to 1.43.1 and fix expect types

vigneshshanmugam:
+ Updates PW to the latest release `1.43.1` as part of the regular cadence #614 
+ PR removes some of the build techniques that we used for importing expect library matcher from `test` package, Instead since the expect package has been includes in the `playwright` library itself, we no longer need that workaround since `test-runner` is separated already to `@playwright/test` package. 
+ Fix the expect types that was raised in https://github.com/elastic/synthetics/issues/917
+ To workaround the module resolution issues, we resolve to the absolute path and use Node.js resolution algorithm to find the package that was installed along with the Synthetics library. This is safer than manually recursing.


devcorpio (approved):
LGTM

Important to highlight: before and after the fix my IDE was autocompleting everything related to `expect` properly. (I restarted IDE several times while switching branches and removing/installing node_modules).

Let me know if you want me to test something very specific, but again, I don't see any difference with the autocompletion of expect's methods when writing monitors.

vigneshshanmugam:
@devcorpio It might have worked before because of the NPM synthetics linking, if you create a fresh project and make sure there is no links to your local `@elastic/synthetics` repo. Then `expect` types wont work since the resolution of import would break as the issue pointed out. 

With the new change, This is not the case anymore as we are using the Node.js Import resolution which would find it in the node_modules directory. Probably a mistake from my part when adding the expect support.

Issue #930, opened 2024-06-03
Add support for more frequencies to Lightweight and Browser monitors


vigneshshanmugam:
+ Allow users to configure monitor frequencies less than `<1minute` for Lightweight monitors. Kibana already [added support](https://github.com/elastic/kibana/pull/184380) for sub minute resolutions, we need to expose them on the agent. 
+ We need to make sure the Types for the schedules align with this new format and restrict <1m resolution for the Browser monitors as we don't support them on our global monitoring locations. 
+ Add support for `2m`  monitor frequency to all browser + Lightweight monitors

PR #932, opened 2024-06-11
feat: support every 10 and 30 seconds intervals for lightweight monitors

shahzad31:
fixes https://github.com/elastic/synthetics/issues/930
support 10 and s30 seconds intervals for lightweight monitors !!

Some changes were also needed in kibana https://github.com/elastic/kibana/pull/185920

### DIFF
--- a/__tests__/push/__snapshots__/index.test.ts.snap
+++ b/__tests__/push/__snapshots__/index.test.ts.snap
@@ -25,7 +25,7 @@ Run 'npx @elastic/synthetics init' to create project with default settings.
 exports[`Push error on invalid schedule 1`] = `
 "Aborted. Invalid synthetics project settings.
 
-Set default schedule(12) to one of the allowed values - 1,3,5,10,15,20,30,60,120,240
+Set default schedule(12) to one of the allowed values - 1,2,3,5,10,15,20,30,60,120,240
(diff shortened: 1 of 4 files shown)


vigneshshanmugam (commented):
Thanks @shahzad31 Code looks good. Few thoughts. 

- What happens if users push these changes to an older version of Kibana? Does the API reject the monitors? 
- Can we add a follow up doc issue and add the relevant info for Lightweight and Browser monitor and also suggest from what stack we start supporting these new schedules.

shahzad31:
> Thanks @shahzad31 Code looks good. Few thoughts.
> 
> * What happens if users push these changes to an older version of Kibana? Does the API reject the monitors?
> * Can we add a follow up doc issue and add the relevant info for Lightweight and Browser monitor and also suggest from what stack we start supporting these new schedules.

Usage against old kibana will throw an error, since we do validation there.

vigneshshanmugam (approved):

MakoWish:
What version was this supposed to work on? We are on 8.14.1, and 1 minute is still the quickest interval we can choose.

shahzad31:
@MakoWish it will be shipped in 8.15.0, which is just around the corner.

PR #938, opened 2024-07-08
fix: update playwright to 1.45.1

vigneshshanmugam:
+ Part of #614 
+ Use the Node.js version `18.20.3` which is distributed on the Heartbeat image.


emilioalvap (approved):
LGTM

PR #939, opened 2024-07-09
fix: push only newly created/updated monitors

### LINKED ISSUE #936: Fix push command to only push monitors that were modified/deleted
+ Currently we push all the monitors when pushing monitors, even when the monitor configuration is unchanged which causes confusion to all users. We can filter down and push the modified monitors (Added/Updated/Deleted) ones instead of pushing the all the unchanged monitors as well - https://github.com/elastic/synthetics/blob/21b4b2f4b9243be04bad8105048ef398970a83f7/src/push/index.ts#L93

+ Same is also applicable to deleted monitor list.

vigneshshanmugam:
+ fix #936

### DIFF
--- a/src/push/index.ts
+++ b/src/push/index.ts
@@ -288,7 +288,7 @@ export async function pushLegacy(monitors: Monitor[], options: PushOptions) {
 
   let schemas: MonitorSchema[] = [];
   if (monitors.length > 0) {
-    progress(`bundling ${monitors.length} monitors`);
+    progress(`preparing ${monitors.length} monitors`);
     schemas = await buildMonitorSchema(monitors, false);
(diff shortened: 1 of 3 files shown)


src/push/index.ts
```diff
@@ -78,7 +78,7 @@ export async function push(monitors: Monitor[], options: PushOptions) {
 
   const { monitors: remote } = await bulkGetMonitors(options);
 
-  progress(`bundling ${monitors.length} monitors`);
+  progress(`preparing ${monitors.length} monitors`);
```
vigneshshanmugam, line 81: reworded this as this was confusing as we also include LW monitors here which doesn't get bundled.

emilioalvap (approved):
LGTM

PR #942, opened 2024-07-26
fix: log screenshot errors in debug mode

vigneshshanmugam:
+ Sometimes the page gets crashed during the screenshot events, making it really hard to debug whats happening with the Journey/Steps. This is fixed already in the PW v1.45 - https://github.com/microsoft/playwright/issues/28995#issuecomment-2167261979. PR just adds a debug log event when screenshot fails to get captured for some reason and makes it easier to see whats happening during page crash events.


emilioalvap (approved):
LGTM

PR #948, opened 2024-08-27
ci: use obltmachine for tagging and creating the release in the release workflow

reakaleek:
Due to the introduction of tag rulesets, the github-actions[bot] cannot bypass the rule and thus cannot push a tag.
Hence, we use @obltmachine, who is on the bypasser list, to create the release.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -27,7 +27,11 @@ jobs:
       - uses: actions/checkout@v4
         with:
           fetch-depth: 0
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -15,7 +15,6 @@ on:
         default: false
 
 permissions:
-  contents: write # needed to push the tag and create the release
```
reakaleek, line 18: This is not needed anymore because we use the PAT of @obltmachine

emilioalvap (approved):
LGTM

PR #957, opened 2024-09-13
feat: add support for Time based OTP

vigneshshanmugam:
+ Part of https://github.com/elastic/synthetics-dev/issues/367
+ Add first class MFA support with a prebundled TOTP library `otpauth`. Its actively maintained and works on all runtimes.  
+ Synthetics library exposes a `mfa.totp` function that can be used to generate a One time token during the journey run. 
```ts
import { journey, step, mfa} from '@elastic/synthetics';

journey('2FA', ({ page }) => {
  step('Login using 2FA', async () => {
    // login using username and pass and go to 2FA in next page
    const token = mfa.token("NB2W45DFOIZA");
    await page.getByPlaceholder("token-input").fill(token)
  });
});
```
+ Users can also configure it for Inline and other Project based journeys by specifying these MFA secrets from params which would allow for generating the token at runtime instead of generating from the UI which would cause authentication issues as journeys would run past the period the token is valid for. 

```ts
    const token = mfa.token(params.MFA_GH_SECRET);
```

### DIFF
--- a/src/core/mfa.ts
+++ b/src/core/mfa.ts
@@ -23,15 +23,43 @@
  *
  */
 
-import { createTOTP } from '../../src/core/totp';
+import { TOTP } from "otpauth";
 
-describe('TOTP', () => {
-  it('generate one time token', () => {
-    const totp = createTOTP({ secret: "FLIIOLP3IR3W" });
-    expect(totp.generate().length).toBe(6);
-    expect(totp.toString()).toEqual('otpauth://totp/SyntheticsTOTP?secret=FLIIOLP3IR3Q&algorithm=SHA1&digits=6&period=30')
-  });
-});
+type TOTPOptions = {
+  /**
+   * Provider or Service the secret is associated with
+   */
+  issuer?: string
+  /**
+   * Account Identifier.
+   * @default 'SyntheticsTOTP'
+   */
+  label?: string
+  /**
+   * Include issuer prefix in label.
+   */
+  issuerInLabel?: boolean
+  /**
+   * The encoded secret key used to generate the TOTP.
+   */
+  secret?: string
+  /**
+   * The algorithm used to generate the TOTP.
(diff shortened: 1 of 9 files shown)


shahzad31 (approved):
LGTM !!

Tested with https://otplib.yeojz.dev/ 

and it seems to work as expected !!

src/core/mfa.ts
```diff
@@ -0,0 +1,65 @@
...
+   */
+  digits?: number
+  /**
+   * Validity period in seconds for the token.
+   * @default 30
+   */
+  period?: number
+};
+
+export function totp(secret?: string, options: TOTPOptions = {}) {
```
shahzad31, line 63: how come secret is optional? isn't that always required?

vigneshshanmugam: No, if we dont pass anything, then it gets auto generated. Just a connivence if all we want is a 6 digit code.

shahzad31 (approved):
LGTM !!

src/core/mfa.ts
emilioalvap: nit: should this make `issuerInLabel: true` explicit here? Instead of relying on the library default

vigneshshanmugam: We could do it though, however different platforms GH / Cloudflare differs here. GH explicitly checks it whereas cloudflare doesnt.

emilioalvap (approved):
LGTM

kvesthy:
@vigneshshanmugam There is discrepancy about the script below **mfa.token**

const token = mfa.token(params.MFA_GH_SECRET) 

It should be 

const token = **mfa.totp(**params.MFA_GH_SECRET)

Can you please review?

vigneshshanmugam:
@kvesthy Good catch, Sent out a fix to the docs. Thank you.

Issue #729, opened 2023-03-02
[Synthetics] - add meta information to monitoring data


jalogisch:

dejongm:
We're migrating to Synthetics from Heartbeat and this feature is very much a requirement for us to move forward without implementing some messy workarounds. Can we get a status, regarding progress or timeline on this?

shahzad31:
@dejongm @jalogisch do you expect those meta fields to be insert at the root of the documents? or would it suffice if they are nested let's say under a field `meta : {area: website}` it might be bit difficult for us to append at root , since that may have mappings repercussions.  Appending under `meta` or lets say `labels` would be pretty easy and can be done in near term.

philippkahr:
I would rather do `labels` as this is in ECS and I have seen it more commonly in the wild with like `labels.team: abc`.

jalogisch:
I second the ECS Compliant solution.

herrBez:
If you need to rename the labels to the expected fields, you can also use the Ingest Pipeline `synthetics@custom`.

dejongm:
I agree to adhere with the ECS and nest them under `labels`.

PR #959, opened 2024-09-19
feat: add labels meta

shahzad31:
Allows the ability to add meta via fields.* fields, data will be appended under labels.* in heartbeat documents !!

fixes https://github.com/elastic/synthetics/issues/729 !!

This will only work with kibana 8.16.0  with PR  https://github.com/elastic/kibana/pull/193250 !!

```

  monitor.use({
    id: 'example-monitor',
    schedule: 10,
    fields: {
      foo: 'bar',
      team: 'synthetics',
    }
  });
```
### lightweight 
```

heartbeat.monitors:
- type: http
  name: Todos Lightweight
  id: "todos- light  weightt
  enabled: true
  urls: "https://elastic.github.io/synthetics-demo"
  schedule: '@every 3m10s'
  timeout: 16s
  alert.status.enabled: true
  fields.environment: "production"
  fields.team: "todos"
```

### DIFF
--- a/__tests__/push/monitor.test.ts
+++ b/__tests__/push/monitor.test.ts
@@ -105,9 +105,13 @@ describe('Monitors', () => {
       },
       fields: { area: 'website' },
     });
-    monitor.update({ locations: ['brazil'] });
+    monitor.update({ locations: ['brazil'], fields: { env: 'dev' } });
     const schema1 = await buildMonitorSchema([monitor], true);
     expect(schema1[0].hash).not.toEqual(schema[0].hash);
+    expect(schema1[0].fields).toEqual({
+      area: 'website',
+      env: 'dev',
+    });
   });
 
   it('parse @every schedule format', async () => {
@@ -475,7 +479,7 @@ heartbeat.monitors:
       });
     });
 
-    it('supports labels in config', async () => {
+    it('supports fields in config', async () => {
       await writeHBFile(`
 heartbeat.monitors:
 - type: icmp
--- a/__tests__/push/monitor.test.ts
+++ b/__tests__/push/monitor.test.ts
@@ -83,6 +83,7 @@ describe('Monitors', () => {
       hash: expect.any(String),
       locations: ['europe-west2-a', 'australia-southeast1-a'],
       privateLocations: ['germany'],
+      fields: { area: 'website' },
     });
   });
 
@@ -102,6 +103,7 @@ describe('Monitors', () => {
       filter: {
         match: 'test',
       },
+      fields: { area: 'website' },
     });
     monitor.update({ locations: ['brazil'] });
     const schema1 = await buildMonitorSchema([monitor], true);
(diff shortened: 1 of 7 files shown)


emilioalvap:
@shahzad31 I'm worried this will confusing since the lightweight representation is not a valid heartbeat monitor and effectively undocumented, wdyt?

shahzad31:
> @shahzad31 I'm worried this will confusing since the lightweight representation is not a valid heartbeat monitor and effectively undocumented, wdyt?

@emilioalvap i will created a docs issue in a follow up to document this.

src/push/monitor.ts
```diff
@@ -268,6 +268,7 @@ export function buildMonitorFromYaml(
     enabled: config.enabled ?? options.enabled,
     locations: options.locations,
     tags: options.tags,
+    labels: parseLabels(config, options.labels),
```
emilioalvap, line 271: Please add this new field to some of the unit tests validating lightweight monitors

shahzad31: i have added here [`200e52f` (#959)](https://github.com/elastic/synthetics/pull/959/commits/200e52f7622ad28985d46c42c4723399ddd54533)

src/common_types.ts
```diff
@@ -255,6 +255,7 @@ export type PushOptions = Partial<ProjectSettings> &
     kibanaVersion?: string;
     yes?: boolean;
     tags?: Array<string>;
+    labels?: Record<string, string>;
```
vigneshshanmugam, line 258: Let's rename this to fields to map with the heartbeat top level fields name and under the hood in Kibana, you can map to labels. Having labels as top level is confusing - 

Also let's write JS doc annotation so it's visible on TS config.

__tests__/push/monitor.test.ts
```diff
@@ -472,6 +472,50 @@ heartbeat.monitors:
         hosts: ['elastic.co:443'],
       });
     });
+
+    it('supports labels in config', async () => {
```
vigneshshanmugam, line 476: Let's add test for browser monitor as well.

paulb-elastic:
@shahzad31 
>This will only work with kibana 8.16.0

If using < 8.16.0, are these just ignored (i.e. it won't fail to `push`)?

shahzad31:
> @shahzad31 
> >This will only work with kibana 8.16.0
> 
> If using < 8.16.0, are these just ignored (i.e. it won't fail to `push`)?

@paulb-elastic validation will fail the push. User will need to remove labels.

__tests__/push/monitor.test.ts
```diff
@@ -472,6 +474,50 @@ heartbeat.monitors:
...
```
vigneshshanmugam, line 478:
```suggestion
    it('supports fields in config', async () => {
```

src/common_types.ts
```diff
@@ -255,6 +255,7 @@ export type PushOptions = Partial<ProjectSettings> &
     kibanaVersion?: string;
     yes?: boolean;
     tags?: Array<string>;
+    fields?: Record<string, string>;
```
vigneshshanmugam, line 258: nit: MonitorConfig['fields']

vigneshshanmugam (commented):
LGTM overall, Can you add a test for Browser monitor and supplying both global and monitor fields. Thanks

vigneshshanmugam (approved):
LGTM

PR #960, opened 2024-09-21
feat: expose runner info in hooks

### LINKED ISSUE #954: Expose journey/step info details on the hooks
+ Use case is to run specific logic based on the status of the journey (whether its failed or succeeded).

vigneshshanmugam:
+ fix #954 
+ Apologies for the big PR. Had to move a lot of things and type the classes appropriately to lay the ground work for exposing the runner information to the hooks. Please have a look at `runner.ts` as its the main file that deals with exposing them. 
+ Took the liberty to cleanup some of the things with launching browsers and attaching/disposing context during journey runs only when necessary instead of starting/stop browsers all the time even during skipped, unmatched journeys. 
+ Runner information is exposed on the all the hooks during journey run, The API and interface looks like this, docs will be updated to reflect this new API through separate PR. 
```ts
interface RunnerInfo {
  /**
   * Processed configuration from the CLI args and the config file
   */
  readonly config: RunOptions;
  /**
   * Currently active journey
   */
  readonly journey: Journey | undefined;
  /**
   * All registerd journeys
   */
  readonly journeys: Journey[];
}

journey({ name: 'test journey', tags: ['ltag', "hello"], }, ({ page }) => {
  before(({ info }) => {
    // info.config.params
  });
  after(({ info }) => {
    // info.journey.status 
    // info.journey.duration
    // info.journey.steps 
  });
  step('step 1', async () => {
    await page.goto("https://example.com")
  });
});
```


src/common_types.ts
```diff
@@ -282,12 +284,10 @@ export type SyntheticsConfig = {
...
 
 /** Runner Payload types */
-export type JourneyResult = {
-  status: StatusValue;
-  error?: Error;
+export type JourneyResult = Partial<Journey> & {
   networkinfo?: PluginOutput['networkinfo'];
   browserconsole?: PluginOutput['browserconsole'];
-  steps?: Array<StepResult>;
+  stepsresults?: Array<StepResult>;
```
shahzad31, line 290: what's the history behind keeping these lower case?

src/reporters/junit.ts
```diff
@@ -80,7 +80,7 @@ export default class JUnitReporter extends BaseReporter {
   override onStepEnd(
     journey: Journey,
     step: Step,
-    { status, error, start, end }: StepEndResult
+    { }: StepEndResult
```
shahzad31, line 83: should this be removed if it's not needed, are you keeping it for backward compatibility?

vigneshshanmugam: This can be removed, forgot to clean up.

src/common_types.ts
vigneshshanmugam: Nothing I can think so, sort of kept it compatible with the Plugin types - https://github.com/elastic/synthetics/blob/03c74b85f8e76c103f2b55e981b275d6ac6fcd46/src/plugins/plugin-manager.ts#L36

src/dsl/step.ts
```diff
@@ -23,28 +23,37 @@
...
+  readonly name: string;
+  readonly index: number;
+  readonly location?: Location;
+  #cb: VoidCallback;
   skip = false;
   soft = false;
   only = false;
+  _startTime = 0;
+  duration = -1;
+  status: StatusValue = 'succeeded';
```
emilioalvap, line 38: nit: I feel something in the lines of `pending` would be a more sensible default for this.

shahzad31:
i did some basic testing, will do more, i do have one question, we are only doing this with hook callbacks, i think it would be nice if we can do the same with step call back. what do you think?

src/core/gatherer.ts
```diff
@@ -43,25 +43,37 @@ const DEFAULT_TIMEOUT = 50000;
  */
 export class Gatherer {
   static browser: ChromiumBrowser;
+  static pluginManager: PluginManager;
 
-  static async setupDriver(options: RunOptions): Promise<Driver> {
-    log('Gatherer: setup driver');
-    const { wsEndpoint, playwrightOptions } = options;
+  static async launchBrowser(options: RunOptions) {
+    if (Gatherer.browser != null) {
```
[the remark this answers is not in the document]
vigneshshanmugam: Yes thats always the case even before. During local runs, there is a single browser instance.

__tests__/core/runner.test.ts
```diff
@@ -171,33 +167,51 @@ describe('runner', () => {
...
       j1: { status: 'failed', error },
       j2: { status: 'failed', error },
     });
   });
 
+  it('run journey - expose info in hooks', async () => {
+    const error = new Error('Broken step');
+    const j1 = journey('fail-journey-with-hooks', () => {
+      before(({ info }) => {
+        expect(info.journey?.status).toBe('succeeded');
```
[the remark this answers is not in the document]
vigneshshanmugam: The current running journey in the execution context.

src/dsl/step.ts
shahzad31: should be fine for UI

vigneshshanmugam:
> i do have one question, we are only doing this with hook callbacks, i think it would be nice if we can do the same with step call back. what do you think?

We chatted offline. The info will be exposed also on journeys, so you can get that in step and hooks if needed. But provide the types also on hooks as its natural to do in hooks and works in beforeAll etc.

__tests__/core/runner.test.ts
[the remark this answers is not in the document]
emilioalvap: 👍 Ties to my suggestion on the default value, feels a bit weird to have a step succeed _before_ execution.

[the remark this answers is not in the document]
vigneshshanmugam: Shahazad confirmed it shouldn't break the UI - https://github.com/elastic/synthetics/pull/960#discussion_r1781369515. Lets keep the journey and step in `pending` status as per your suggestion.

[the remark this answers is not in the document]
vigneshshanmugam: https://github.com/elastic/synthetics/pull/960/commits/479970ce2e2bfb70e1926de6e58c5728a7d6cce8

src/dsl/step.ts
vigneshshanmugam: Done https://github.com/elastic/synthetics/pull/960/commits/479970ce2e2bfb70e1926de6e58c5728a7d6cce8

shahzad31 (approved):
LGTM !!

Tested with bunch of journeys and things looks normal !!

emilioalvap:
++ LGTM, spent some time comparing HB output with this PR, everything seems consistent

src/reporters/base.ts
```diff
@@ -84,15 +80,15 @@ export default class BaseReporter implements Reporter {
   }
 
   /* eslint-disable @typescript-eslint/no-unused-vars */
-  onJourneyEnd(_: Journey, { error }: JourneyEndResult) {
+  onJourneyEnd(journey: Journey, { }: JourneyEndResult) {
```
shahzad31, line 83: status is part of both journey and JourneyEndResult , should we change that?

shahzad31: but i don't think it's available as value in JourneyEndResult

src/core/runner.ts
```diff
@@ -277,142 +270,133 @@ export default class Runner {
...
+      browserDelay: this.#browserDelay,
       timestamp: getTimestamp(),
       options,
-      ...pluginOutput,
-      browserconsole: filterBrowserMessages(
-        pluginOutput.browserconsole,
-        status
-      ),
+      networkinfo: pOutput.networkinfo,
+      browserconsole: bConsole,
```
shahzad31, line 324: status is removed as value but is still typed in JourneyEndResult.

vigneshshanmugam: Its because of the `Partial<Journey>`, Have to clean this up and use the `RunnerInfo`. Will do that when we do the cleanup for Multi Step API tests.

PR #967, opened 2024-10-17
ci: use pull-request event

v1v:
We enabled this to validate the PR listing in the title always.

However, I cannot think of what's the reason we cannot protect ourselves and use `Require approval for all outside collaborators` as we do in the other workflows.

Any genuine reasons, we need to run always regardless the code is not tested until the PR has been approved?


vigneshshanmugam (approved):
LGTM, we should be good to run this workflow after the approval.

PR #974, opened 2024-10-25
feat: support service.name property for tests

andrewvc:
This PR allows the association of APM service names with monitors, as allowed by plain heartbeat. I think something isn't quite right though as trying this against a real cluster I didn't see that value set. Marking this as a draft for now.

### DIFF
--- a/__tests__/push/monitor.test.ts
+++ b/__tests__/push/monitor.test.ts
@@ -82,6 +82,7 @@ describe('Monitors', () => {
       enabled: true,
       hash: expect.any(String),
       locations: ['europe-west2-a', 'australia-southeast1-a'],
+      'service.name': 'test-service',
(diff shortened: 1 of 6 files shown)


vigneshshanmugam (commented):
@andrewvc Do we know if Kibana API supports this field when creating monitors?

andrewvc:
It is documented here: https://www.elastic.co/guide/en/kibana/current/add-monitor-api.html as the dotted `service.name`, @shahzad31 do you know if it would work in the bulk push?

shahzad31:
kibana PR is here https://github.com/elastic/kibana/pull/212229

shahzad31 (approved):
LGTM !!

andrewvc:
Closing in favor of https://github.com/elastic/kibana/pull/212229

PR #975, opened 2024-10-25
feat: introduce totp command for OTP token

vigneshshanmugam:
+ Part of https://github.com/elastic/synthetics-dev/issues/367
+ Goes together with https://github.com/elastic/synthetics/pull/957 and makes the onboarding to the MFA based auth easier. 
+ Introduces a `totp` command that lets users generate a OTP token based on the provided secret that can be used to onboard the Synthetics TOTP generator to other providers. 

```ts
npx @elastic/synthetics totp secret --issuer <issuer> --label <label>

// prints
OTP Token: 123456

```

Documentation will be covered as part of https://github.com/elastic/observability-docs/issues/4334

### DIFF
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -27,6 +27,8 @@
 
 import { program, Option } from 'commander';
 import { cwd } from 'process';
+import { bold } from 'kleur/colors';
+import { resolve } from 'path';
 import { CliArgs, PushOptions } from './common_types';
 import { reporters } from './reporters';
 import {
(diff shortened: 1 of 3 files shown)


src/cli.ts
```diff
@@ -274,4 +275,20 @@ program
...
+  .command('totp <secret>')
+  .description('Generate a Time-based One-Time token using the provided secret.')
+  .option('--issuer <issuer>', 'Provider or Service the secret is associated with.')
+  .option('--label <label>', 'Account Identifier (default: SyntheticsTOTP)')
+  .action((secret, cmdOpts: TOTPCmdOptions) => {
+    try {
+      const token = totp(secret, cmdOpts)
+      write(bold(`OTP Token: ${token}`));
+    } catch (e) {
+      e && error(e);
```
emilioalvap, line 289: What is the benefit of filtering out the falsy values in this case? It might be confusing if the program exits on error but no error is reported

emilioalvap (approved):
Small comment, otherwise LGTM

src/cli.ts
emilioalvap: I'd rather we print the error, whatever that might end up being
```suggestion
      error(e);
```

emilioalvap (approved):
LGTM

PR #977, opened 2024-10-31
fix: adding warning when pushing monitors from yaml files

### LINKED ISSUE #976: Add warnings for silent failures during lightweight synthetics pushes
When the push process encounters a lightweight synthetic `.yml` file with an unexpected encoding it will fail to parse the heartbeat.monitors sequence and fail silently.  There should be a warning for when a lightweight synthetic monitor file is encoded wrong and will be ignored.  See code below with continue statement.

```javascript
 const content = await (0, promises_1.readFile)(file, 'utf-8');
        const lineCounter = new yaml_1.LineCounter();
        const parsedDoc = (0, yaml_1.parseDocument)(content, {
            lineCounter,
            merge: true,
            keepSourceTokens: true,
        });
        // Skip other yml files that are not relevant
        const monitorSeq = parsedDoc.get('heartbeat.monitors');
        if (!monitorSeq) {
            continue;
        }
```

Ideally there would be a warning letting the user know that their project contains files that may not be processed due to encoding issues.

Another useful warning would be one for users (like myself) who converted from using heartbeat to using synthetics.  Heartbeat supported `browser` type monitors, and synthetics does not (makes sense because people should be using journeys.  The issue is that browser type monitors are simply ignored (see code below).  There should be a warning if a browser type monitor is detected in a file.  Especially since the current behavior is to simply ignore them and not produce an error.  

```javascript
            const monitor = mergedConfig[i];
            // Skip browser monitors from the YML files
            if (monitor['type'] === 'browser') {
                continue;
            }
```

As a user of elastic, my team encountered these silent ignores and it took us a significant amount of time engaged with elastic support to identify and fix the problem (support case 01774618).  If these warnings were in place, then we would have been able to identify the problem much sooner.

Spoonsk:
+ fixes #976 
Added warnings to npx push for scenarios where monitors are silently ignored.  See #976

### DIFF
--- a/src/push/monitor.ts
+++ b/src/push/monitor.ts
@@ -28,7 +28,7 @@ import { extname, join } from 'path';
 import { LineCounter, parseDocument, Document, YAMLSeq, YAMLMap } from 'yaml';
 import { bold, red } from 'kleur/colors';
 import { Bundler } from './bundler';
-import utf8Validate from 'utf-8-validate';
+import NodeBuffer from 'node:buffer';
 import { SYNTHETICS_PATH, totalist, indent, warn } from '../helpers';
 import { LocationsMap } from '../locations/public-locations';
 import {
@@ -187,10 +187,10 @@ export async function createLightweightMonitors(
   for (const file of lwFiles.values()) {
     // First check encoding and warn if any files are not the correct encoding.
     const bufferContent = await readFile(file);
-    const isUtf8 = utf8Validate(bufferContent);
+    const isUtf8 = NodeBuffer.isUtf8(bufferContent);
     if (!isUtf8) {
       warn(
-        `${file} is not UTF-8 encoded. Monitor configurations in this file might be ignored.`
+        `file ${file} is not encoded in utf-8.  utf-8 encoding is expected.  Monitor configurations in this file may be ignored.`
       );
     }
     const content = bufferContent.toString('utf-8');
--- a/src/push/monitor.ts
+++ b/src/push/monitor.ts
@@ -189,9 +189,7 @@ export async function createLightweightMonitors(
     const bufferContent = await readFile(file);
     const isUtf8 = NodeBuffer.isUtf8(bufferContent);
     if (!isUtf8) {
-      warn(
-        `file ${file} is not encoded in utf-8.  utf-8 encoding is expected.  Monitor configurations in this file may be ignored.`
-      );
+      warn(`${file} is not UTF-8 encoded. Monitors might be skipped.`);
     }
     const content = bufferContent.toString('utf-8');
     const lineCounter = new LineCounter();
--- a/src/push/monitor.ts
+++ b/src/push/monitor.ts
@@ -228,7 +228,7 @@ export async function createLightweightMonitors(
       // Skip browser monitors from the YML files
       if (monitor['type'] === 'browser') {
         warn(
-          `a monitor with type 'browser' was detected in file ${file}.  'browser type monitors are not supported.  Monitor will be skipped.`
+          `Browser monitors from ${file} are skipped.`
         );
         continue;
       }
--- a/src/push/monitor.ts
+++ b/src/push/monitor.ts
@@ -28,6 +28,7 @@ import { extname, join } from 'path';
 import { LineCounter, parseDocument, Document, YAMLSeq, YAMLMap } from 'yaml';
 import { bold, red } from 'kleur/colors';
 import { Bundler } from './bundler';
+import utf8Validate from 'utf-8-validate';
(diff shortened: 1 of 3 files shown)


Spoonsk:
> ❌ Author of the following commits did not sign a [Contributor Agreement](https://www.elastic.co/contributor-agreement): [8e1d65c](https://github.com/elastic/synthetics/commit/8e1d65c9003b1d68dcde9caf4cef128775e9e84e), [f855a3c](https://github.com/elastic/synthetics/commit/f855a3c3f1a88ebf5571d6ccdfaaa1e0e1e04241)
> 
> Please, read and sign the above mentioned agreement if you want to contribute to this project

I've since signed the agreement.  Let me know if you need anything else here.

src/push/monitor.ts
```diff
@@ -184,7 +185,15 @@ export async function createLightweightMonitors(
   let warnOnce = false;
   const monitors: Monitor[] = [];
   for (const file of lwFiles.values()) {
-    const content = await readFile(file, 'utf-8');
+    // First check encoding and warn if any files are not the correct encoding.
+    const bufferContent = await readFile(file);
+    const isUtf8 = utf8Validate(bufferContent);
```
vigneshshanmugam, line 190: is this necessary?, I believe you can add a try catch on `readFile` and log the failures?

Spoonsk: @vigneshshanmugam in my case with a 'utf-16le' `readFile(file, 'utf-8')`  read the file without errors, however the resulting string was jumbled.  This caused the yaml to not parse and the file to be skipped.

vigneshshanmugam: Gotcha, can you remove the dep and use the built in Buffer.isUtf8 https://nodejs.org/api/buffer.html#bufferisutf8input

src/push/monitor.ts
```diff
@@ -218,6 +227,9 @@ export async function createLightweightMonitors(
       const monitor = mergedConfig[i];
       // Skip browser monitors from the YML files
       if (monitor['type'] === 'browser') {
+        warn(
+          `a monitor with type 'browser' was detected in file ${file}.  'browser type monitors are not supported.  Monitor will be skipped.`
```
vigneshshanmugam, line 231:
```suggestion
          `Browser monitors from ${file} are skipped.`
```

src/push/monitor.ts
```diff
@@ -184,7 +185,15 @@ export async function createLightweightMonitors(
   let warnOnce = false;
   const monitors: Monitor[] = [];
   for (const file of lwFiles.values()) {
-    const content = await readFile(file, 'utf-8');
+    // First check encoding and warn if any files are not the correct encoding.
+    const bufferContent = await readFile(file);
+    const isUtf8 = utf8Validate(bufferContent);
+    if (!isUtf8) {
+      warn(
+        `file ${file} is not encoded in utf-8.  utf-8 encoding is expected.  Monitor configurations in this file may be ignored.`
```
vigneshshanmugam, line 193:
```suggestion
        `${file} is not UTF-8 encoded. Monitor configurations in this file might be ignored.`
```

src/push/monitor.ts
Spoonsk: @vigneshshanmugam I updated the code to use the standard library function.  I also updated the `@types@node` dev dependencies.  The types update was needed to get the definition for the utf8 function, however it was also needed simply to be in lockstep with the engine requirement.

vigneshshanmugam (approved):

vigneshshanmugam:
@Spoonsk Thank you for the contribution :)

vigneshshanmugam:
@Spoonsk Can you sign the CLA again? Thanks.

Spoonsk:
> @Spoonsk Can you sign the CLA again? Thanks.

I filled out the CLA again with my github name and we are still seeing the CLA failure.  Is there something we need to do to retrigger the CLA evaluation?

shahzad31:
@Spoonsk i see commits from 2 users names, Joe Gast and you, maybe Joe Gast also need to sign?

JosephGast:
> @Spoonsk i see commits from 2 users names, Joe Gast and you, maybe Joe Gast also need to sign?

I just signed.

Spoonsk:
I see the issue.  We have some commits on here that were created by a user with an email that isn't associated with a github account.  I will try to re-author those commits on this Branch, if that doesn't work.  Then I may need to create another branch with another PR and we can abandon this one.

Spoonsk:
There we go.  I was able to re-author the Joe Gast commits.  Should CLA should be happy now.

vigneshshanmugam (approved):
Thank you @Spoonsk

justinkambic (approved):
LGTM!!

PR #986, opened 2024-12-11
fix: omit playwright-core from bundles !!

shahzad31:
we have usage of playwright-core in internal monitors, which ends up bundling along.

### DIFF
--- a/src/push/bundler.ts
+++ b/src/push/bundler.ts
@@ -51,7 +51,7 @@ export class Bundler {
         write: false,
         minifyWhitespace: true,
         sourcemap: 'inline',
-        external: ['@elastic/synthetics'],
+        external: ['@elastic/synthetics', 'playwright-core'],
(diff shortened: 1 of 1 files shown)


src/push/bundler.ts
```diff
@@ -51,7 +51,7 @@ export class Bundler {
         write: false,
         minifyWhitespace: true,
         sourcemap: 'inline',
-        external: ['@elastic/synthetics'],
+        external: ['@elastic/synthetics', 'playwright-core'],
```
vigneshshanmugam, line 54: We could do this, but we would end up creating broken experience for the users if they decided to pull some code from PW. 

Allowing this is better for the UX side than having a broken experience. WDYT?

vigneshshanmugam: Better way would be to disallow it as per your PR and have a symbolic link in the Heartbeat side when we set up project monitors so we dont have a broken experience for users.

PR #987, opened 2024-12-11
feat: export types from playwright-core 

shahzad31:
export types from playwright-core  !!

Tested a simple journey 
```

import { journey, step, monitor, request } from '@elastic/synthetics';

journey(`My Example Journey playwrgiht`, ({ page }) => {
  monitor.use({
    locations: ['us_central'],
    enabled: false,
  });
    const context = await request.newContext();
    await page.goto('http://google.com');
  });
});
```

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -61,7 +61,7 @@ export type {
   BrowserContext,
 } from 'playwright-core';
 
-export type { devices, request } from 'playwright-core';
+export { devices, request } from 'playwright-core';
 
 export type { Serializable } from 'playwright-core/types/structs';
 
--- a/src/index.ts
+++ b/src/index.ts
@@ -57,13 +57,8 @@ export type {
   APIRequestContext,
   APIResponse,
   APIRequest,
-  Browser,
-  BrowserContext,
 } from 'playwright-core';
-
 export { devices, request } from 'playwright-core';
-
-export type { Serializable } from 'playwright-core/types/structs';
 
 /**
  * Export the types necessary to write custom reporters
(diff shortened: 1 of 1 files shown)


src/index.ts
```diff
@@ -72,6 +79,7 @@ export type {
   JourneyResult,
   StepResult,
   StartEvent,
+  Params,
```
vigneshshanmugam, line 82: It can be simply referenced as  `SyntheticsConfig['params']`?

src/index.ts
```diff
@@ -55,9 +55,16 @@ export type {
...
-  devices,
+  APIResponse,
+  APIRequest,
+  Browser,
+  BrowserContext,
 } from 'playwright-core';
 
+export { devices, request } from 'playwright-core';
+
+export type { Serializable } from 'playwright-core/types/structs';
```
vigneshshanmugam, line 66: Do we know the use case for this one?

src/index.ts
```diff
@@ -91,3 +99,5 @@ export type {
   Steps,
   actionTitle,
 } from './formatter/javascript';
+
+export { log } from './core/logger';
```
vigneshshanmugam, line 103: likewise, this is a helper purely for the library as it uses the internals to identify if we have to log or not. I dont like the idea of exposing it and allowing users to rely on this internal details.

src/index.ts
vigneshshanmugam: Removed.

src/index.ts
vigneshshanmugam: Removed.

vigneshshanmugam (approved):
LGTM

axw (approved):

andrewvc (approved):

PR #988, opened 2024-12-12
fix: expose running journey as currentJourney

vigneshshanmugam:
+ Some libraries depend on this private variables, so renaming it to the previous value to make things easier.


emilioalvap (approved):
LGTM

PR #989, opened 2024-12-16
fix: fix backwards compatibility to 1.14

emilioalvap:
- Fix backwards compatibility to 1.14

### DIFF
--- a/src/core/runner.ts
+++ b/src/core/runner.ts
@@ -129,7 +129,7 @@ export default class Runner implements RunnerInfo {
   }
 
   /**
-   * @deprecated Since version 1.17.0. Use _updateMonitor instead.
+   * @deprecated Since version Please do not rely on the internal methods.
    * Alias _addHook for backwards compatibility
    */
   addHook(type: HookType, callback: HooksCallback) {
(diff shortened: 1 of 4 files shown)


src/core/runner.ts
```diff
@@ -131,8 +128,20 @@ export default class Runner implements RunnerInfo {
     this.#hooks[type].push(callback);
   }
 
+  /**
+   * @deprecated Since version 1.17.0. Use _updateMonitor instead.
```
vigneshshanmugam, line 132:
```suggestion
   * @deprecated Since version 1.17.0. Please do not rely on the internal methods. 
```

vigneshshanmugam: WDYT?

vigneshshanmugam (commented):
LGTM, but instead of alias, I would very much in favor of asking users to not rely on these methods or callout saying use it with caution.

src/core/runner.ts
emilioalvap: I'm OK with that, will update.

vigneshshanmugam (approved):

Issue #973, opened 2024-10-22
enable/document local testing using


seanstory:
related to https://github.com/elastic/synthetics/issues/972

Actually deployed synthetic journeys seem to run differently than when you [test locally](https://www.elastic.co/guide/en/observability/current/synthetics-create-test.html#synthetics-test-locally), in that they don't go through the same bundling step. This means that tests can pass locally, but fail when deployed, making them much more difficult to troubleshoot. 

It would be great to have support in the CLI (either baked in, or at least documented) to be able to run a more comparable test locally, without needing to push semi-untested journeys.

Issue #929, opened 2024-06-03
List the Ids of the synthetics journeys added, updated, deleted, and unchanged in a push


gsantoro:
Currently, when we push journeys to Kibana, we can only get a count of the journeys that were added, updated, removed, or unchanged in the current push command.

We would like a more detailed output where you can check the IDs of added, updated, removed, or unchanged journeys.

```
NODE_ENV=prod npx @elastic/synthetics push
> Pushing monitors for 'serverless-prod' project in kibana 'default' space
> bundling 0 monitors
> Monitor Diff: Added(0) Updated(0) Removed(1) Unchanged(0)
```

vigneshshanmugam:
Discussed this offline with the team. We will do the following
- Pushing with `debug` flag will show all the monitors - created/updated in addition to the deleted monitors. 
- Group them in to sections so its easier to parse on the Terminal. 
- Add support for `--dry-run` if possible. This is an enhancement.

drewpost:
@vigneshshanmugam @shahzad31 - Did this get done?

a-finocchiaro:
@vigneshshanmugam Thank you for your response! So I have been trying to run it locally and I am getting this output:

<img width="2391" height="1055" alt="Image" src="https://github.com/user-attachments/assets/e49fe753-c2cc-49b8-ab5b-8210d5c17d36" />
(note: the token here is for a devenv and is ephemeral so it's fine that it was included in the output)

I have also verified that `@elastic/synthetics` is installed globally _and_ in the project that I am trying to use `--dry-run` against. 

```
✔0 aaron@fedora ~/repos/ecp-synthetics-monitors/projects/smoke-tests config$ npm list -g
/home/aaron/.asdf/installs/nodejs/20.16.0/lib
├── @elastic/synthetics@1.22.0
├── corepack@0.28.2
└── npm@10.8.1
```

I posted this follow up because in the past when I had run this I thought I saw behavior that seemed to indicate that `push` with `--dry-run` was still pushing monitors to the Kibana instance, but after running this command it seems like the push portion of this is not doing that which is great! However, the piece that runs the monitors always errors out for me. Is there a way to have this command run _just_ the push with dry run and not attempt to run the monitors?

vigneshshanmugam:
>  However, the piece that runs the monitors always errors out for me. Is there a way to have this command run just the push with dry run and not attempt to run the monitors?

[...]

PR #991, opened 2024-12-19
feat: run pushed monitors locally in dry-run mode

vigneshshanmugam:
+ fix https://github.com/elastic/synthetics/issues/973 and https://github.com/elastic/synthetics/issues/929
+ Setting `DEBUG=synthetics` will provides detailed information like Monitor size, overall project size and monitors that were added/deleted/updated etc. 
+ With `--dry-run`, we have the ability to now run these monitors similar to how it would be run inside the Synthetics Public/Private managed locations. 

Screenshot 2024-12-19 at 4 17 52 PM

### DIFF
--- a/src/push/index.ts
+++ b/src/push/index.ts
@@ -188,7 +188,7 @@ export function formatDuplicateError(monitors: Set<Monitor>) {
   let inner = '';
   for (const monitor of monitors) {
     const { config, source } = monitor;
-    inner += `* ${config.id} - ${source.file}:${source.line}:${source.column} \n`;
+    inner += `* ${config.id} - ${source.file}:${source.line}:${source.column}\n`;
   }
   outer += indent(inner);
   return outer;
--- a/src/push/index.ts
+++ b/src/push/index.ts
@@ -24,7 +24,7 @@
  */
 import { readFile, writeFile } from 'fs/promises';
 import { prompt } from 'enquirer';
-import { bold, grey } from 'kleur/colors';
+import { bold, underline } from 'kleur/colors';
 import {
   getLocalMonitors,
   buildMonitorSchema,
@@ -57,8 +57,11 @@ import {
   isBulkAPISupported,
   isLightweightMonitorSupported,
   logDiff,
+  logGroups,
+  printBytes,
 } from './utils';
-import { log } from '../core/logger';
+import { run } from "./run-local"
+
 
 export async function push(monitors: Monitor[], options: PushOptions) {
   if (parseInt(process.env.CHUNK_SIZE) > 250) {
@@ -84,7 +87,7 @@ export async function push(monitors: Monitor[], options: PushOptions) {
   const { monitors: remote } = await bulkGetMonitors(options);
 
   progress(`preparing ${monitors.length} monitors`);
-  const schemas = await buildMonitorSchema(monitors, true);
+  const { schemas, sizes } = await buildMonitorSchema(monitors, true);
   const local = getLocalMonitors(schemas);
 
   const { newIDs, changedIDs, removedIDs, unchangedIDs } = diffMonitorHashIDs(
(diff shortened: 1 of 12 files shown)


src/push/index.ts
```diff
@@ -106,7 +126,6 @@ export async function push(monitors: Monitor[], options: PushOptions) {
   }
 
   if (removedIDs.size > 0) {
-    log(`deleting monitor ids: ${Array.from(removedIDs.keys()).join(', ')}`);
```
vigneshshanmugam, line 109: covered by the logGroups

src/push/monitor.ts
```diff
@@ -148,6 +151,15 @@ export async function buildMonitorSchema(monitors: Monitor[], isV2: boolean) {
       monitor.setContent(content);
       Object.assign(schema, { content, filter });
     }
+    const size = monitor.size();
```
emilioalvap, line 154: `SIZE_LIMIT_KB` description specifies this is gzipped, whereas `monitor.size()` is not, maybe we should align the description

src/push/monitor.ts
```diff
@@ -148,6 +151,15 @@ export async function buildMonitorSchema(monitors: Monitor[], isV2: boolean) {
...
+    const sizeKB = Math.round(size / 1000);
+    if (sizeKB > SIZE_LIMIT_KB) {
+      let outer = bold(`Aborted: Bundled code ${sizeKB}kB exceeds the recommended ${SIZE_LIMIT_KB}kB limit. Please check the dependencies imported.\n`);
+      const inner = `* ${config.id} - ${source.file}:${source.line}:${source.column}\n`;
+      outer += indent(inner);
+      throw red(outer);
```
emilioalvap, line 160: It might be better DX if we iterate through all monitors and collect all errors before pushing so that we report all at once, instead of just the first, wdyt?

src/push/run-local.ts
```diff
@@ -0,0 +1,132 @@
...
+    });
+  })
+}
+
+async function writePkgJSON(dir: string, synthPath: string) {
+  const packageJsonContent = {
+    name: "project-journey",
+    private: "true",
+    dependencies: {
+      "@elastic/synthetics": `file:${synthPath}`,
```
emilioalvap, line 94: Can I suggest using [pathToFileURL](https://nodejs.org/api/url.html#urlpathtofileurlpath-options) instead?

src/push/run-local.ts
```diff
@@ -0,0 +1,132 @@
...
+    },
+  };
+  await writeFile(`${dir}/package.json`, JSON.stringify(packageJsonContent, null, 2), "utf-8");
```
emilioalvap, line 97: Same here, prefer `path.join()`

src/push/run-local.ts
```diff
@@ -0,0 +1,132 @@
...
+  }
+  const content = schema.content;
+  await writeFile(zipPath, content, "base64");
+  await unzipFile(zipPath, unzipPath);
+}
+
+export async function runLocal(schemas: MonitorSchema[]) {
+  // lookup installed bin path of a node module
+  const resolvedPath = execFileSync('which', ['elastic-synthetics'], { encoding: 'utf8' }).trim();
+  const synthPath = resolvedPath.replace("bin/elastic-synthetics", "lib/node_modules/@elastic/synthetics")
```
emilioalvap, line 113: Same here, prefer `path.join`

src/push/run-local.ts
```diff
@@ -0,0 +1,132 @@
...
+  const rand = Date.now();
+  const zipPath = `/tmp/synthetics-zip-${rand}.zip`;
```
emilioalvap, line 115: Same here, prefer `os.tmpdir`+ `path.join`

src/push/monitor.ts
vigneshshanmugam: It is though, the content extracted is the gzipped size of the monitor. Can you give me hint on where we should align it?

src/push/monitor.ts
vigneshshanmugam: I thought about it, But it might cause the errors to be not meaningful if user decides to pull in a dep like `playwright-core` etc which would get included across all journeys if it was in a helper library (common case in ecp). Throwing it away first avoids those issues. This is why we also kept it this way from the beginning.

emilioalvap (approved):
LGTM

PR #993, opened 2025-01-16
fix: calculate push page size based on monitor size per batch

shahzad31:
decide the no of monitors based on size of monitor size while pushing !!

### DIFF
--- a/src/push/bundler.ts
+++ b/src/push/bundler.ts
@@ -24,25 +24,19 @@
  */
 
 import path from 'path';
-import { stat, unlink, readFile } from 'fs/promises';
+import { unlink, readFile } from 'fs/promises';
 import { createWriteStream } from 'fs';
 import * as esbuild from 'esbuild';
 import archiver from 'archiver';
 import { commonOptions } from '../core/transform';
 import { SyntheticsBundlePlugin } from './plugin';
 
-// 1500KB Max Gzipped limit for bundled code to be pushed as Kibana project monitors.
-const SIZE_LIMIT_KB = 1500;
-
 function relativeToCwd(entry: string) {
   return path.relative(process.cwd(), entry);
 }
 
 export class Bundler {
-  moduleMap = new Map<string, string>();
-  constructor() {}
-
-  async prepare(absPath: string) {
+  async bundle(absPath: string) {
     const options: esbuild.BuildOptions = {
       ...commonOptions(),
       ...{
@@ -59,57 +53,42 @@ export class Bundler {
     if (result.errors.length > 0) {
       throw result.errors;
     }
-    this.moduleMap.set(absPath, result.outputFiles[0].text);
+    return result.outputFiles[0].text;
   }
 
-  async zip(outputPath: string) {
+  async zip(source: string, code: string, dest: string) {
     return new Promise((fulfill, reject) => {
-      const output = createWriteStream(outputPath);
+      const output = createWriteStream(dest);
       const archive = archiver('zip', {
         zlib: { level: 9 },
       });
       archive.on('error', reject);
       output.on('close', fulfill);
       archive.pipe(output);
-      for (const [path, content] of this.moduleMap.entries()) {
-        const relativePath = relativeToCwd(path);
-        // Date is fixed to Unix epoch so the file metadata is
-        // not modified everytime when files are bundled
-        archive.append(content, {
-          name: relativePath,
-          date: new Date('1970-01-01'),
-        });
-      }
+      const relativePath = relativeToCwd(source);
+      // Date is fixed to Unix epoch so the file metadata is
+      // not modified everytime when files are bundled
+      archive.append(code, {
+        name: relativePath,
+        date: new Date('1970-01-01'),
+      });
       archive.finalize();
     });
   }
 
   async build(entry: string, output: string) {
-    await this.prepare(entry);
-    await this.zip(output);
+    const code = await this.bundle(entry);
+    await this.zip(entry, code, output);
     const content = await this.encode(output);
-    const sizeKb = await this.checkSize(output);
     await this.cleanup(output);
-    return { content, sizeKb };
+    return content;
   }
 
   async encode(outputPath: string) {
     return await readFile(outputPath, 'base64');
   }
 
-  async checkSize(outputPath: string) {
-    const { size } = await stat(outputPath);
-    const sizeKb = size / 1024;
-    if (sizeKb > SIZE_LIMIT_KB) {
-      throw new Error(
-        `Bundled monitor code exceeds the recommended ${SIZE_LIMIT_KB}KB limit. Please check your dependencies and try again.`
-      );
-    }
-    return sizeKb;
-  }
-
   async cleanup(outputPath: string) {
-    this.moduleMap = new Map<string, string>();
     await unlink(outputPath);
   }
 }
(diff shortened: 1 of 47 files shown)


vigneshshanmugam (commented):
@shahzad31 Thanks for the PR, Can we change the PR based on what we discussed
```
decide the no of monitors based on size of monitor size while pushing
```

src/push/bundler.ts
```diff
@@ -87,10 +87,10 @@ export class Bundler {
   async build(entry: string, output: string) {
     await this.prepare(entry);
     await this.zip(output);
-    const data = await this.encode(output);
-    await this.checkSize(output);
+    const content = await this.encode(output);
+    const sizeKb = await this.checkSize(output);
     await this.cleanup(output);
-    return data;
+    return { content, sizeKb };
```
emilioalvap, line 93: @shahzad31 Please review, this change of the return type is breaking unit tests

emilioalvap:
@shahzad31 can we adjust the PR description to match the actual intention? This doesn't seem to involve any retries as it is

shahzad31:
> @shahzad31 can we adjust the PR description to match the actual intention? This doesn't seem to involve any retries as it is

yeah i will come back to this PR next week

src/push/request.ts
```diff
@@ -58,6 +58,7 @@ export async function sendReqAndHandleError<T>(
   options: APIRequestOptions
 ): Promise<T> {
   const { statusCode, body } = await sendRequest(options);
+
   return (
     await handleError(statusCode, options.url, body)
```
emilioalvap, line 63: Can we add the chunk size and payload size information on this error? It will be helpful to know how big the payload was and how many monitors per chunk were sent

src/push/index.ts
```diff
@@ -102,7 +104,10 @@ export async function push(monitors: Monitor[], options: PushOptions) {
     for (const value of sizes.values()) {
       totalSize += value;
     }
-    progress('total size of the monitors payload is ' + printBytes(totalSize));
+    progress(
+      `total size of the ${sizes.size} monitors payload is ` +
```
vigneshshanmugam, line 108: We already show the number of monitors, I feel like we are duplicating again.

src/push/request.ts
```diff
@@ -58,8 +58,14 @@ export async function sendReqAndHandleError<T>(
...
 ): Promise<T> {
   const { statusCode, body } = await sendRequest(options);
+
   return (
-    await handleError(statusCode, options.url, body)
+    await handleError(
+      statusCode,
+      options.url,
+      body,
+      `${options.body?.length} number of monitors were sent, with a total size of ${options.body?.length} bytes`
```
vigneshshanmugam, line 67: Feels weird, Why not Kibana API return proper error message when we go over the limits?

src/push/kibana_api.ts
```diff
@@ -37,7 +37,8 @@ import {
 import { generateURL } from './utils';
 
 // Default chunk size for bulk put / delete
-export const CHUNK_SIZE = parseInt(process.env.CHUNK_SIZE) || 100;
+export const CHUNK_SIZE = parseInt(process.env.CHUNK_SIZE) || 250;
+export const MAX_PAYLOAD_SIZE_KIB = (1048576 * 50) / 1024; // 50MB
```
vigneshshanmugam, line 41:
```suggestion
export const MAX_PAYLOAD_SIZE_KIB = 50*1024; // 50MiB
```

src/push/utils.ts
```diff
@@ -90,14 +90,51 @@ export function printBytes(bytes: number) {
...
+  sizes: Map<string, number>,
+  maxChunkSizeKB: number,
+  maxChunkItems: number
+): Array<T[]> {
+  const chunks: Array<T[]> = [];
+  let currentChunk: T[] = [];
+  let currentSize = 0;
+
+  for (const item of arr) {
+    const sizeKb = item.id ? sizes.get(item.id) / 1024 : 1;
```
vigneshshanmugam, line 112: This is size in bytes, Lets round it off, so can be Math.round(sizes.get(item.id) / 1000)

src/push/kibana_api.ts
```diff
@@ -37,7 +37,8 @@ import {
 import { generateURL } from './utils';
 
 // Default chunk size for bulk put / delete
-export const CHUNK_SIZE = parseInt(process.env.CHUNK_SIZE) || 100;
+export const CHUNK_SIZE = parseInt(process.env.CHUNK_SIZE) || 250;
```
vigneshshanmugam, line 40: We did a poor naming for this, lets change the variable name atleast that clarifies its monitors per push request

src/push/utils.ts
```diff
@@ -90,14 +90,51 @@ export function printBytes(bytes: number) {
...
   for (let i = 0; i < arr.length; i += size) {
     chunks.push(arr.slice(i, i + size));
   }
   return chunks;
 }
 
+export function getSizedChunks<T extends { id?: string }>(
+  arr: Array<T>,
+  sizes: Map<string, number>,
+  maxChunkSizeKB: number,
```
vigneshshanmugam, line 104: Maybe lets move this to KB instead of KiB as we based operate on the Kbs in the chunks.

src/push/utils.ts
```diff
@@ -90,14 +90,51 @@ export function printBytes(bytes: number) {
...
+    currentSize += sizeKb;
+  }
+
+  // Push the last chunk if it contains any items
+  if (currentChunk.length > 0) {
+    chunks.push(currentChunk);
+  }
+
+  // log chunk sizes
+  progress(`Monitors will be pushed in ${chunks.length} chunks`);
```
vigneshshanmugam, line 133: Lets try to move away from the `chunks` wording, user something like batches or requests seems appropriate.

vigneshshanmugam (changes requested):
The approach looks good, just needs changes on the wording. Also, can you share how it looks on the CLI? Thanks

src/push/request.ts
shahzad31: kibana doesn't return the payload size

shahzad31:
this is how error looks like now 

image

src/push/request.ts
vigneshshanmugam: In this case, we need to do this only for the 413 status code, Otherwise we are showing this error message for all the error codes which feels incorrect.

shahzad31: done

src/push/utils.ts
```diff
@@ -90,12 +90,51 @@ export function printBytes(bytes: number) {
...
+  sizes: Map<string, number>,
+  maxBatchSizeKiB: number,
+  maxBatchItems: number
+): Array<T[]> {
+  const batches: Array<T[]> = [];
+  let currentBatch: T[] = [];
+  let currentSize = 0;
+
+  for (const item of arr) {
+    const sizeKiB = item.id ? sizes.get(item.id) / 1024 : 1;
```
vigneshshanmugam, line 112: The size is stored in bytes, we need to convert them to kibibytes and round it off - https://github.com/elastic/synthetics/pull/993#discussion_r1966379439

vigneshshanmugam (commented):
Looks good except for the conversion logic.

vigneshshanmugam (approved):

src/push/utils.ts
vigneshshanmugam: fixed https://github.com/elastic/synthetics/pull/993/commits/2523967688afbd9f86b867e8070b753b86887054

PR #996, opened 2025-02-05
chore: print error when parsing synthetics.config.ts

brunofarache:
When `synthetics.config.ts` throws an error, `config.ts` doesn't give any clue about the root cause, it's better if we can print the original thrown `Error`.


shahzad31 (approved):
LGTM !!

vigneshshanmugam (approved):
Thank you.

Issue #931, opened 2024-06-06
capture Network information related to the APIRequestContext


vigneshshanmugam:
### Summary

Currently, synthetics agent only captures the network requests that originates within the Browser Context which means when you write a Journey, all the requests that originates within the browser context (including all pages) will be captured. But all the requests that are sent via `APIRequestContext` are not captured by the agent leaving holes in the waterfall information.

vigneshshanmugam:
> @vigneshshanmugam - would we then render these as individual lines on the network waterfall? Would these be of a long enough duration to actually even be visible there in the waterfall view?

Will depend on what users would be doing inside a step, Like users could send multiple API requests inside a single step and other steps could be empty. We need to account for all these use-cases.

drewpost:
This is related to multi-step API test dependency and we'll pick it up when we tackle that work

jasonrhodes:
Can we close this? (I can't :P )

Issue #900, opened 2024-03-01
[proposal] Lightweight API journeys DSL


vigneshshanmugam:
### Summary

Introduce a way to run Lightweight API based journeys that does not involve launching browsers for every run, instead rely on the [`APIRequest`](https://playwright.dev/docs/api/class-apirequest) API provided by Playwright core to do Web API testing. 

### Proposal

The agent will introduce a new API for registering and running these API based checks. 

#### DSL

```ts
type APIJourneyOptions = {
  name: string;
  id?: string;
  tags?: string[];
};

type APIJourneyCallback = (options: {
  request: APIRequestContext;
  params: Params;
}) => void;

export { apijourney }

```
#### Usage

Users would import this API and configure the tests. 

```ts
import { apijourney, monitor, expect } from "@elastic/synthetics";

apijourney('test example.com', async ({ request, params }) => {
 // monitor configurations would still work as usual
 monitor.use({ schedule: 10, locations: [ "us_east1"] });

  const resp = await request.get('https://example.com/api/getText', {
    params: { id: params.id },
  });
  const data = resp.json();
  expect(data).toEqual({ text: 'Hello, World!' });
});

```

#### difference with `journey` API and request context exposed. 

+ Does not require spinning a new browser instance, instead each api journey will create a separate [APIRequestContext](https://playwright.dev/docs/api/class-apirequest#api-request-new-context)  which have its own cookie storage. 
+ The `request` object exposed by the browser journey API does share the same context configurations meaning the cookies will be shared based between API requests and Page requests.

jasonwhetton:
Eagerly awaiting this feature.

dominiqueclarke:
@vigneshshanmugam Do you have an example of the outputted heartbeat documents?

Should we expect, from a UI perspective, that the schema will be identical to typical browser journeys aside from including screenshots? How does this impact network timing for the waterfall chart? 

Which parameters can you configure for api journeys that you cannot configure for browser journeys? Which parameters for browser journeys can you configure that you can't for api journeys? Just envisioning the UI form in my head and trying to determine how different it will be from our current browser monitor form.

@drewpost 
How do you see us representing this type of UI monitor from a UI perspective? Seems like we'll need to expose a 5th monitor type? 

Which data streams do we use? Do we continue to use `synthetics-browser` datastreams or do we need to introduce 5th datastream for this monitor type?

Issue #137, opened 2020-11-12
Add API Journeys


andrewvc:
Currently our synthetic journeys are all browser based. In the future it'd be good to add support for browser-less journeys for testing APIs or other things that don't need a browser. 

This would also be useful for supporting odd protocols heartbeat doesn't support since with a full JS environment almost anything can be tested.

This issue is a stub and needs further fleshing out (example syntax etc.)(

iorfix:
Is there any update about this enhancement? Since playwright added test-api-testing, is it already possible to use it in synthetic tests?

vigneshshanmugam:
@iorfix Not yet, we need to expose the request object from Playwright which can be used for API testing.

hager-yousri:
When using playwright document to write api test I always get this error while it is working good with me in the playwright project but it doesn't work in synthetic journeys
TypeError: Cannot read properties of undefined (reading 'post')

hager-yousri:
any update about using playwright api testing in synthetic tests?
if it is still not supported when it is expected to be supported plz?

drewpost:
Hi @hager-yousri - thanks for following this issue. This is a capability that is on our roadmap. We're currently pushing towards the 1.0 release of our new Synthetics app in Elastic Observability. In addition to the full browser synthetic testing, we will be delivering the single ping availability checking that's in Uptime today. After our 1.0 launch, we will look to add in multi-step API journeys. It is a high priority for us.

paulmouzas:
Is there any update on this? I would like to use some lightweight checks that require oauth2 authentication involving an initial step of retrieving an access token from a separate endpoint (OAuth 2 client credentials flow). Heartbeat doesn't support this so a multistep API journey would fulfill this requirement.

I see that there is a request object that does not have the full overhead of the browser object, but it does not appear to be intended to be used in this way. The [docs](https://www.elastic.co/guide/en/observability/current/synthetics-create-test.html#synthetics-request-param) state that 

> The request parameter is not intended to be used for writing pure API tests. Instead, it is a way to support writing plain HTTP requests in service of a browser-based test.

paulb-elastic:
@paulmouzas you are correct in that the `request` object is intended to be used as part of a browser monitor (e.g. getting some content/data from an API endpoint that is then used in a browser based user journey).

[...]

PR #997, opened 2025-02-05
feat: multi step api runner

shahzad31:
## Summary

Adds `apiJourney(...)` — runs API-only checks without launching a browser. Same DSL shape as `journey(...)`, but the callback receives an isolated `APIRequestContext` instead of a `Page`/`Browser` driver. Pushes to Kibana as an HTTP-type monitor.

```ts
import { apiJourney, monitor, step, expect } from '@elastic/synthetics';

apiJourney('orders API health', ({ request, params }) => {
  monitor.use({ schedule: 1, locations: ['us_east'] });

  let token: string;
  step('obtain OAuth2 token', async () => {
    const r = await request.post(`${params.authUrl}/oauth2/token`, {
      form: { grant_type: 'client_credentials', client_id: params.clientId, client_secret: params.clientSecret },
    });
    expect(r.status()).toBe(200);
    token = (await r.json()).access_token;
  });

  step('check /orders', async () => {
    const r = await request.get(`${params.apiUrl}/orders`, {
      headers: { Authorization: `Bearer ${token}` },
    });
    expect(r.status()).toBe(200);
  });
});
```

> **Requires Playwright ≥ 1.61.** TLS and remote-address capture use the native `APIResponse.securityDetails()` / `APIResponse.serverAddr()` APIs added in [microsoft/playwright#40932](https://github.com/microsoft/playwright/pull/40932).

## Closes

- #137 — Multi-step lightweight API journeys (5+ year ask)
- #900 — API journeys DSL proposal
- #931 — Network capture for `APIRequestContext`
- Internal customer summary: `elastic/synthetics-dev#464`

## What's in the PR

- **`APIJourney` DSL** — subclass of `Journey`, monitor type `'http'`, full `monitor.use(...)` / `params` / `expect` / `.skip` / `.only` support.
- **No browser when not needed** — runner skips `launchBrowser` if no browser journeys are registered.
- **`APINetworkManager`** — intercepts `APIRequestContext.fetch` (the single funnel) and captures URL, method, status, headers (redacted), body bytes, timings, and step attribution.
- **Native TLS / remote address** — reads `APIResponse.securityDetails()` (cert issuer / subject / `validFrom` / `validTo` / protocol) and `APIResponse.serverAddr()` (remote IP / port) straight off the response from the actual request, mirroring the browser network path. HTTPS *and* plain HTTP surface `server.ip` / `server.port`; cert info follows redirects to the final hop. No extra socket or parallel handshake.
- **Playwright 1.61 bump + codegen vendoring** — 1.61 bundles its server code into `coreBundle.js` and no longer exports the recorder's codegen classes, so the minimal slice `SyntheticsGenerator` extends (`_asLocator` / `_generateActionCall` / `JavaScriptFormatter`) is vendored into `src/formatter/codegen.ts`, reusing Playwright's still-exported `iso` helpers. Recorder snapshots are unchanged.
- **`PluginManager`** — common `NetworkPlugin` interface so API network data isn't dropped at journey-end.
- **`JSONReporter`** — emits `journey.type: 'api'`, omits browser-only fields, surfaces `tls.*` and `server.ip` / `server.port`.

## Tests

All green across:

- `dsl/api-journey`, `core/api-journey-register` — DSL + factory
- `plugins/api-network`, `plugins/plugin-manager` — capture, native TLS / remote-addr, orchestration
- `core/api-runner` — e2e, no-browser, failure path, **empty-step grouping**, **cookie isolation**, HTTPS with TLS / IP / body-bytes
- `reporters/json` — `type: 'api'` emission, server / TLS fields
- `formatter/javascript` — recorder codegen snapshots unchanged against the vendored generator

## Manual verification

Ran the built CLI against `api.github.com`, `httpbin.org`, and a local self-signed HTTPS server. Confirmed: no browser spawned, full `tls.*` / `server.*` / `body.bytes` populated, cookies isolated, redaction still works.

## Docs

- README updated with browser vs. API journey sections.
- `examples/todos/api.journey.ts` runnable example added.

## Follow-ups (separate PRs)

1. **Kibana UI** — handle `journey.type: 'api'` in waterfall, monitor list, data stream. Open question from #900 needs an answer.
2. **mTLS client cert** — to fully solve `elastic/enhancements#25818` (Atos France).
3. **Finer timing breakdown** (`dns`, `connect`, `ssl`, `wait`, `receive`) — `APIResponse` doesn't expose per-phase timings; needs Node `http.Agent` hooks (or a future Playwright API).
4. **HAR export integration** — Playwright already records API timings to HAR (`microsoft/playwright#32613`); could complement the response fields.


src/dsl/api-journey.ts
```diff
@@ -0,0 +1,142 @@
...
+type HookType = 'before' | 'after';
+export type APIHooks = Record<HookType, Array<APIHooksCallback>>;
+type APIJourneyCallbackOpts = {
+  params: Params;
+  request: APIRequestContext;
+  info: APIRunnerInfo;
+};
+export type APIJourneyCallback = (options: APIJourneyCallbackOpts) => void;
+
+export class APIJourney {
```
vigneshshanmugam, line 54: Ideally I would like us to have a base Journey class that both BrowserJourney and APIJourney extends from and the runner should be a single entity that decides how to run each of them. I am not in favor of having multiple runners.

src/common_types.ts
```diff
@@ -324,4 +341,10 @@ export type JourneyEndResult = JourneyStartResult &
     options: RunOptions;
   };
 
+export type APIJourneyEndResult = JourneyStartResult &
+  APIJourneyResult & {
+    browserDelay: number;
```
vigneshshanmugam, line 346: browserDelay is not appropriate for apijourneys?

src/core/runner.ts
```diff
@@ -309,22 +316,24 @@ export default class Runner implements RunnerInfo {
 
   async #startJourney(journey: Journey, options: RunOptions) {
     journey._startTime = monotonicTimeInSeconds();
-    this.#driver = await Gatherer.setupDriver(options);
+    this.#driver = await Gatherer.setupDriver(options, journey.type);
```
vigneshshanmugam, line 319: This feels a bit weird , Maybe a better way would be to check if there are defined browser journeys and then launch the browser context etc based on that.

src/dsl/api-journey.ts
```diff
@@ -0,0 +1,72 @@
...
+export type APIJourneyCallback = (options: APIJourneyCallbackOpts) => void;
+
+type APIJourneyType = (
+  options: string | APIJourneyOptions,
+  callback: APIJourneyCallback
+) => APIJourney;
+
+export class APIJourney extends Journey {
+  #cb: APIJourneyCallback;
+  #driver?: Driver | APIDriver;
```
vigneshshanmugam, line 52: can it have a Driver at any point?

src/plugins/api-network.ts
```diff
@@ -0,0 +1,156 @@
...
+    // Intercept API requests
+    ['fetch', 'get', 'post', 'put', 'delete', 'patch', 'head'].forEach(
+      method => {
+        this._originalMethods[method] = request[method].bind(request);
+        request[method] = this._interceptRequest.bind(this, method);
+      }
+    );
+  }
+
+  async _interceptRequest(method, url, options: any) {
```
vigneshshanmugam, line 100: This needs some careful review as we need some extensive tests for cached, aborted, intercepted requests etc, Will do a detailed review later on when we have tests etc.

src/plugins/plugin-manager.ts
```diff
@@ -45,23 +51,32 @@ export class PluginManager {
...
-        instance = new Tracing(this.driver, options);
-        break;
-      case 'performance':
-        instance = new PerformanceManager(this.driver);
-        break;
-      case 'browserconsole':
-        instance = new BrowserConsole(this.driver);
-        break;
+
+    if ('context' in this.driver) {
```
vigneshshanmugam, line 59: Plugin Manager should be transparent to the underlying journey that was run whether its browser/api based. There will be always a chance when browser/api based journeys coexisting. I would handle these and make them respective plugins transparent.

vigneshshanmugam (commented):
Did a first pass.

src/dsl/api-journey.ts
```diff
@@ -0,0 +1,105 @@
...
+  /**
+   * Override default monitor type so `synthetics push` registers this
+   * journey as an HTTP monitor rather than a browser monitor.
+   */
+  override _updateMonitor(config: MonitorConfig) {
+    this._setMonitor(
+      new Monitor({
+        name: this.name,
+        id: this.id,
+        type: 'http',
```
mcapell: I think we should use another type here (eg: `api`). `http` already exist on the lightweight monitors

src/plugins/api-network.ts
```diff
@@ -0,0 +1,275 @@
...
+      browser: { name: 'api', version: '' },
+      request: {
+        url,
+        method: httpMethod,
+        headers: requestHeaders,
+        bytes: headersBytes(requestHeaders) + requestBodyBytes,
+        body: requestBodyBytes > 0 ? { bytes: requestBodyBytes } : undefined,
+      },
+      response: {
+        status: -1,
```
mcapell, line 129: question: in case of an error, all these values might be emitted, as the entry is added in the `results` array. Should we keep those values as null / undefined instead?

src/plugins/api-network.ts
```diff
@@ -0,0 +1,275 @@
...
+      loadEndTime: -1,
+      responseReceivedTime: -1,
+      resourceSize: 0,
+      transferSize: 0,
+      timings: {
+        blocked: -1,
+        dns: -1,
+        ssl: -1,
+        connect: -1,
+        send: -1,
```
mcapell, line 143: If I'm not wrong, all these timings (besides wait, receive, and total) are not updated anywhere else, and they will remain `-1`. Should we remove them instead?

templates/synthetics.config.ts
```diff
@@ -4,6 +4,10 @@ export default env => {
   const config: SyntheticsConfig = {
     params: {
       url: 'https://elastic.github.io/synthetics-demo/',
+      // Used by the API journey examples. jsonplaceholder is a free,
```
mcapell: nit: I don't think this comment adds a lot of value. Should we remove or simplify it?

src/reporters/json.ts
```diff
@@ -504,18 +547,28 @@ export default class JSONReporter extends BaseReporter {
       });
     }
 
+    /**
```
mcapell: nit / non-blocking: for consistency with the other comments, can we simplify and convert this one into an inline comment? similar to the `process_startup_epoch_us` one

src/reporters/json.ts
```diff
@@ -151,8 +179,8 @@ function formatTLS(tls: SecurityDetails) {
         subject: {
           common_name: tls.subjectName,
         },
-        not_after: new Date(tls.validTo * 1000).toISOString(),
-        not_before: new Date(tls.validFrom * 1000).toISOString(),
+        not_after: epochToIso(tls.validTo),
```
mcapell, line 174: question: is this change needed for the apiJourney? The new implementation looks too defensive

src/reporters/json.ts
```diff
@@ -208,6 +236,7 @@ export function formatNetworkFields(network: NetworkInfo) {
       response,
     },
     tls: formatTLS(response?.securityDetails),
+    server: formatServer(response),
```
mcapell: question: This is also added for browser monitors (see https://github.com/elastic/synthetics/pull/997/changes#diff-0397021fb144e227a6761fcde778498b393bbf8c503d8da094e32cc3dbc18fb4R45-R48). Should we add only for api journeys?

src/plugins/api-network.ts
```diff
@@ -0,0 +1,275 @@
...
+        wait: -1,
+        receive: -1,
+        total: -1,
+      },
+    };
+    this.results.push(entry);
+
+    const wallStart = Date.now();
```
mcapell: question: should we use a monotonic clock here (eg: `now()` [from the helpers file](https://github.com/elastic/synthetics/blob/main/src/helpers.ts#L110-L112))? It would be consistent with other timings such as `getTimestamp()` used early in the file.

mcapell (commented):
Overall looks great. I've added a few questions / comments, but I've been testing it locally and it works as expected.

The only thing we are missing is to update the push monitor feature. There are some `type === "browser"` gates in that logic ([example](https://github.com/elastic/synthetics/blob/main/src/push/monitor.ts#L146)), and I think we need to include the new "api" one.

src/dsl/api-journey.ts
shahzad31: Done — now registered as `api` type. Also updated the push bundling gates (monitor.ts/run-local.ts) and README to match.

src/plugins/api-network.ts
shahzad31: Kept: mirrors the browser `NetworkManager`, which also emits a `status: -1` entry for failed/incomplete requests, so downstream already handles the sentinel.

src/plugins/api-network.ts
shahzad31: Kept for parity with the browser path, which also inits all timings to `-1` and only fills what's measurable (`-1` = unknown).

templates/synthetics.config.ts
shahzad31: Simplified to a single line.

src/reporters/json.ts
shahzad31: Converted to an inline comment.

src/reporters/json.ts
shahzad31: Kept — it guards a real crash: a malformed cert date makes `new Date(...).toISOString()` throw `RangeError` and sink the whole `journey/end` doc (regression test in json.test.ts). Trimmed the comment.

src/reporters/json.ts
shahzad31: Good catch — now emitted for api journeys only; browser output keeps the address under `http.response`. Snapshot updated.

src/plugins/api-network.ts
shahzad31: Done — switched to monotonic `now()` from helpers.

mcapell (approved):
Thanks, LGTM!

PR #999, opened 2025-02-13
chore: update playwright to 1.50.1

shahzad31:
update playwright to 1.50.1 !!

## 1.49.0 Breaking changes !! 

Breaking: channels chrome, msedge and similar switch to new headless
This change affects you if you're using one of the following channels in your playwright.config.ts:

chrome, chrome-dev, chrome-beta, or chrome-canary
msedge, msedge-dev, msedge-beta, or msedge-canary
What do I need to do?
After updating to Playwright v1.49, run your test suite. If it still passes, you're good to go. If not, you will probably need to update your snapshots, and adapt some of your test code around PDF viewers and extensions. See https://github.com/microsoft/playwright/issues/33566 for more details.

Other breaking changes
There will be no more updates for WebKit on Ubuntu 20.04 and Debian 11. We recommend updating your OS to a later version.
Package @playwright/experimental-ct-vue2 will no longer be updated.
Package @playwright/experimental-ct-solid will no longer be updated.

## 1.50.0 Breaking changes !! 
Breaking

-  [expect(locator).toBeEditable()](https://playwright.dev/docs/api/class-locatorassertions#locator-assertions-to-be-editable) and [locator.isEditable()](https://playwright.dev/docs/api/class-locator#locator-is-editable) now throw if the target element is not <input>, <select>, or a number of other editable elements.
- Option [testConfig.updateSnapshots](https://playwright.dev/docs/api/class-testconfig#test-config-update-snapshots) now updates all snapshots when set to all, rather than only the failed/changed snapshots. Use the new enum changed to keep the old functionality of only updating the changed snapshots.


dmlemeshko:
Hi folks, is there any blocker preventing the merge?

vigneshshanmugam (commented):
Code changes LGTM

Due to the new headless mode, we need to tackle the release a bit differently. We can 
 1. Release it without setting the `stack_version` tag to make sure we dont break the HB/Elastic Agent new images accidentally. 
 2. Build a new Elastic agent image with this new version and compare the previous runs with the current one to ensure the Performance metrics, duration all stays intact. 
 
 /cc @emilioalvap FYI

dmlemeshko:
Based on a meeting we had recently, I think we agreed that updating & releasing synthetics with a newer playwright version will give us sometime to run Scout with more recent playwright/test and Chromium versions.

Maybe it is feasible to update playwright  to the most recent one 1.52.0? https://playwright.dev/docs/release-notes#version-152

vigneshshanmugam:
@dmlemeshko We will run a few tests on our end and release a new version - for now we will merge and release as it is with `1.50.1` and do another by upgrading it to `1.52.0` just to be in the safer side.

emilioalvap:
Still running the performance comparison, for starters there's considerable change in size with the new headless version. Comparison of a snapshot build:

```
REPOSITORY         TAG       IMAGE ID       CREATED          SIZE
agent_synthetics   2.0.0     f55dd2ff38c1   32 minutes ago   3.92GB
agent_synthetics   1.0.1     505cad69678c   19 hours ago     3.37GB
```

emilioalvap (approved):

vigneshshanmugam:
> Still running the performance comparison, for starters there's considerable change in size with the new headless version. Comparison of a snapshot build:

Do we know the reason for the increase? Is it just merely coming from the new executables?

emilioalvap:
@vigneshshanmugam Yes, there are now two separate chromium binaries for headless/headful: https://github.com/microsoft/playwright/blame/55333c6f285c017b54e97e6d86363baaeeb079c2/packages/playwright-chromium/install.js#L27

PR #1002, opened 2025-02-25
feat: support output directory for attachments 

shahzad31:


src/plugins/attachments.ts
```diff
@@ -0,0 +1,124 @@
...
+  outputDir?: string;
+  screenshots?: RunOptions['screenshots'];
+};
+
+export class AttachmentsManager {
+  private attachments: Attachment[] = [];
+  #screenshotPath = join(CACHE_PATH, 'screenshots');
+
+  constructor(private driver: Driver, private options: AttachmentOptions) {
+    this.#screenshotPath = options.outputDir
```
vigneshshanmugam, line 43: We shouldn't let the user config touch our screenshotpath, otherwise it would break our blocks gen/reconstruction logic. 

The output dir is only taken in to account on the reporter side, where we extract and reconstruct the file paths based on the given value. Thoughts?

PR #1003, opened 2025-02-25
feat: expose page metrics via rich events !!

shahzad31:
expose page metrics via rich events !!


vigneshshanmugam (commented):
I am not against exposing this information, but the current API is not well thoughtout and currently lacks the page information where the metrics was gathered, so if a step has multiple pages, popups, iframes etc, the metric would be incorrect. We need to come up with a valid design for the Kibana UI and expose the page specific metric information along with step/url information etc. Please let me know if you have some other ideas.

PR #1052, opened 2025-07-18
feat: support multi space monitors !!

shahzad31:
To test the PR 

You can create a dummy project and point it to the source of the directory like this

    "@elastic/synthetics": "file:../../",

then you can specify spaces as an option in config or individula monitors using monitor.use

```
    monitor: {
      schedule: 10,
      locations: ['us_central'],
      spaces: ['test', 'default'],
    },
```

spaces in lightweight are also supported

```
heartbeat.monitors:
  - type: http
    name: Todos Lightweight Test 0
    id: todos-lightweight-test-0
    enabled: true
    urls: 'https://elastic.github.io/synthetics-demo'
    schedule: '@every 60m'
    timeout: 18s
    alert.status.enabled: true
    service.name: todos-lightweight-test
    namespace: food
    spaces:
      - '*'
```

### Notes

1. Kibana API will throw an error if user is trying to create a monitor in non existent space
2. Kibana API will throw an error if user doesn't have permission to create monitors in a space
3. KIbana API will throw error if user is trying to use spaces option before prev 8.19, or 9.1.0 version
4. Project space Id is always appended, that means monitor is always created in a project space, besides user can also choose other spaces
5. monitor.use option will take precedent over project.monitor option


src/push/kibana_api.ts
```diff
@@ -51,6 +51,7 @@ export async function bulkPutMonitors(
   schemas: MonitorSchema[]
 ) {
   const url = generateURL(options, 'bulk_update') + '/_bulk_update';
+  // '/_bulk_update?savedObjectType=synthetics-monitor';
```
shahzad31, line 54:
```suggestion
```

src/core/runner.ts
```diff
@@ -466,6 +467,9 @@ export default class Runner implements RunnerInfo {
       (journey.cb ?? journey.callback)({ params: options.params } as any);
       const monitor = journey.monitor ?? journey?._getMonitor();
       monitor.update(this.#monitor?.config);
+      monitor.setSpaces(
```
vigneshshanmugam, line 470: This should not be required as the .update call should handle the changes from the config. Please check once.

emilioalvap (approved):

PR #1106, opened 2025-11-20
chore: add missing openat2 to seccomp profiles

emilioalvap:
### Summary

Fixes e2e jobs by adding the missing `openat2` call to seccomp profiles.


mcapell (approved):
Nice find!

PR #1117, opened 2026-02-11
feat: consider "chore" commits for patch releases

mcapell:
"chore" commits are not considered by the release process. Instead, consider them for patch releases.


vigneshshanmugam (commented):
What was the motivation behind this change?

PR #1144, opened 2026-05-14
fix: skip husky install in non-prod environments

mcapell:
Currently, when installing only dependencies (not devDependencies), the "prepare": "husky" script fails because Husky isn't installed. This change was introduced [here](https://github.com/elastic/synthetics/pull/1133).

Replace `husky` prepare script with a wrapper script that exits when `husky` is not installed, preventing npm install failures in production and embedded installs.

More information: https://typicode.github.io/husky/how-to.html

### DIFF
--- a/.husky/install.mjs
+++ b/.husky/install.mjs
@@ -2,6 +2,6 @@
 try {
   const husky = await import('husky');
   husky.default();
-} catch {
-  // husky not a dependency, skip
+} catch (e) {
+  if (e.code !== 'ERR_MODULE_NOT_FOUND') throw e;
 }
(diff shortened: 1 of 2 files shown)


zmoog (dismissed):
Looks good, only non-blocking comments.

Suggestions

1. Narrow the catch to only handle missing module errors. The current catch hides all failures. Consider:

```js
try {
  const husky = await import('husky');
  husky.default();
} catch (e) {
  if (e.code !== 'ERR_MODULE_NOT_FOUND') throw e;
}
```

1. This preserves the silent skip for "not installed" while still surfacing real initialization failures.
2. Minor: the catch comment is slightly misleading. It says "husky not a dependency, skip" but currently catches all errors. If you keep the broad catch, something like `// not in a dev environment, skip` is more accurate.

zmoog (approved):

PR #1147, opened 2026-05-28
fix(e2e): unbreak uptime e2e against modern Heartbeat / Kibana

shahzad31:
## Summary

The `e2e-test` job in `ci.yml` has been red on every recent PR. The docker-compose stack was pinned to **8.3.0-SNAPSHOT**, whose Heartbeat ships Node 16 and no longer boots against the current `@elastic/synthetics` tree (`undici` needs Node ≥ 18). Bumping the default to **8.19.0-SNAPSHOT** in `setup.sh` exposed four further incompatibilities — fixed here in one go.

### Root causes & fixes

| # | Symptom | Fix |
|---|---|---|
| 1 | `ReadableStream is not defined` → Heartbeat container exits 1 | Bump default `STACK_VERSION` in `__tests__/e2e/scripts/setup.sh` from `8.3.0-SNAPSHOT` → `8.19.0-SNAPSHOT` (Node 16 → 20). Still overridable via env. |
| 2 | `runtime/cgo: pthread_create failed: Operation not permitted` | The bundled `examples/docker/seccomp_profile.json` doesn't allow `clone3`. Drop the `security_opt:` block in `docker-compose.yml` so Docker's default profile is used. |
| 3 | Heartbeat config validation error: "local browser monitors are now removed" | Remove the `source.local` `todos` monitor from `examples/docker/heartbeat.docker.yml`; the journey only needs `my-monitor`. |
| 4 | Legacy Uptime app hidden + index pattern mismatch (8.19 publishes to `synthetics-*` data streams, app queries `heartbeat-*`) | In `uptime.journey.ts#beforeAll`: enable `observability:enableLegacyUptimeApp` and widen Uptime's `heartbeatIndices` to `heartbeat-*,synthetics-*`. Wait-for-data query targets the same indices and requires ≥ 1 doc (we now ship a single monitor). |
| 5 | Snapshot still returns `total: 0` even though docs match the bool query | Inspector (`observability:enableInspectEsQueries`) showed one shard failing with `No field found for [observer.geo.name] in mapping` — the legacy `getSnapshotCount` scripted_metric hard-references `doc["observer.geo.name"]`, and on `synthetics-*` data streams that field is unmapped unless the monitor declares a location. Inject it via an `add_fields` processor in `heartbeat.docker.yml`. |

### Files

- `__tests__/e2e/scripts/setup.sh` — default `STACK_VERSION` 8.3 → 8.19, env-overridable.
- `__tests__/e2e/docker-compose.yml` — drop stale `seccomp_profile.json` mount.
- `examples/docker/heartbeat.docker.yml` — drop deprecated `source.local` monitor; add `add_fields` processor for `observer.geo.{name,location}`.
- `__tests__/e2e/uptime.journey.ts` — enable legacy Uptime app, widen indices, reorder waits, bump synthetics-data timeout 60s → 180s, accept ≥ 1 doc.

## Test plan

- [x] Local: `STACK_VERSION=8.19.0-SNAPSHOT npx @elastic/synthetics __tests__/e2e/uptime.journey.ts` — 4/4 steps pass, exit 0.

```
Journey: E2e test synthetics
   ✓  Step: 'Go to kibana uptime app' succeeded (2234 ms)
   ✓  Step: 'Check if there is table data' succeeded (163 ms)
   ✓  Step: 'Click on my monitor' succeeded (180 ms)
   ✓  Step: 'It navigates to details page' succeeded (102 ms)
 4 passed (3798 ms)
```

- [ ] CI: this PR's `e2e-test` job goes green.

## Notes

- The scheduled `e2e.yml` workflow (runs `synthetics.journey.ts` via `elastic-package stack`, pinned to `7.17.8` / `8.5.2` in `__tests__/e2e/versions`) is unrelated and currently green; this PR doesn't touch its code paths.
- The `add_fields` processor is the canonical Beats way to inject `observer.geo.*`. The values (`name: docker`, `location: "0, 0"`) are placeholders — they only need to exist for the Painless script not to throw.


__tests__/e2e/docker-compose.yml
```diff
@@ -16,7 +16,14 @@ services:
...
       - bootstrap.memory_lock=true
-      - xpack.security.enabled=false
+      # Security is required for Fleet (which we use to install the
+      # Synthetics integration package so synthetics-* data streams get
+      # the right mappings). Keep HTTP/transport SSL off to avoid
+      # certificate plumbing for an e2e stack.
+      - xpack.security.enabled=true
+      - xpack.security.http.ssl.enabled=false
+      - xpack.security.transport.ssl.enabled=false
+      - ELASTIC_PASSWORD=changeme
```
vigneshshanmugam, line 26: shouldnt this be ELASTICSEARCH_PASSWORD ?

__tests__/e2e/uptime.journey.ts
```diff
@@ -122,12 +141,50 @@ async function waitForKibana() {
...
+        auth: AUTH,
+      });
       esStatus = data?.status.overall.level === 'available';
     } catch (e) {}
   }
 }
 
+async function installSyntheticsIntegration() {
+  console.log('Installing Synthetics Fleet integration package');
+  await axios.post(
```
vigneshshanmugam, line 167: Dont love this, can we check if the integration exists and then do this? that would avoid the race right?

examples/docker/heartbeat.docker.yml
```diff
@@ -17,10 +17,3 @@ heartbeat.monitors:
               }
               await page.hover('css=[data-nav-item=products]');
           });
-  - type: browser
```
vigneshshanmugam, line 20: Any reason why we removed it?

vigneshshanmugam (commented):
LGTM, just some questions.

__tests__/e2e/docker-compose.yml
shahzad31: These are two different env vars on two different images:

- On the **elasticsearch** image (this line), `ELASTIC_PASSWORD` is the bootstrap variable read on first boot to set the `elastic` superuser's password — see [the ES docker docs](https://www.elastic.co/docs/deploy-manage/deploy/self-managed/install-elasticsearch-docker-basic#docker-bootstrap-password).
- On the **kibana** image (line 88, `ELASTICSEARCH_PASSWORD: changeme`) it's what Kibana uses to authenticate _to_ Elasticsearch as `kibana_system`.

So both are correct and not interchangeable.

__tests__/e2e/uptime.journey.ts
shahzad31: Good call — pushed [`5285ef3`](https://github.com/elastic/synthetics/pull/1147/commits/5285ef3). We now `GET /api/fleet/epm/packages/synthetics` first and short-circuit when `item.status === 'installed'`, so on a warm stack the destructive `_data_stream/synthetics-browser-default` delete is never reached. The install + delete only run on the cold-bootstrap path where the race is irrelevant.

Verified locally against an already-bootstrapped stack: log prints `Synthetics Fleet integration already installed, skipping` and the journey still passes 4/4.

examples/docker/heartbeat.docker.yml
shahzad31: The removed monitor was the second one in the file:

```yaml
- type: browser
  id: todos
  name: Todos
  schedule: '@every 1m'
  source:
    local:
      path: '/opt/elastic-synthetics/examples/todos'
```

8.19 Heartbeat removed `source.local` browser monitors entirely, so leaving it in causes config validation to fail and the container exits before ever publishing anything — which is part of why the e2e job was failing on `main`. The uptime journey only needs `my-monitor` for the assertions, so dropping `todos` is the minimal fix.

vigneshshanmugam (approved):

PR #1164, opened 2026-07-26
feat: capture per-phase resource timing for API journeys

shahzad31:
## Summary

Reuses Playwright's newly released [`APIResponse.timing()`](https://github.com/microsoft/playwright/pull/41800) (shipped in **1.62.0**) to give API journeys the same per-phase network timing breakdown that browser journeys already have.

- **Bump Playwright** `1.61.1` → `1.62.0` (`playwright`, `playwright-chromium`, `playwright-core`).
- **New shared helper** `src/plugins/network-timings.ts` — extracts the Resource Timing → phase-breakdown (`getResourceTimings`) and total calculation with wall-clock fallback (`calcTotalTime`). Since `APIResponse.timing()` returns the identical shape as `Request.timing()`, both `NetworkManager` (browser) and `APINetworkManager` (API) now share exactly the same math.
- **API journeys** (`api-network.ts`): replaced the coarse wall-clock stopwatch — which left `blocked/dns/ssl/connect` at `-1` and hard-coded `receive: 0` — with a real breakdown from `response.timing()`. Phases that don't happen (reused keep-alive socket) or are unknown (HAR replay) still report `-1`, and `calcTotalTime` falls back to the wall-clock span in that case.
- **Browser journeys** (`network.ts`): refactored to call the shared helper (removing the duplicated inline math and `_calcTotalTime`); behavior unchanged.

### Before vs after (API journeys)

| phase | before | after |
|---|---|---|
| blocked / dns / ssl / connect | `-1` | derived from `APIResponse.timing()` |
| wait | wall-clock fetch duration | TTFB from Resource Timing |
| receive | hard-coded `0` | download time from Resource Timing |

## Test plan

- [x] `network-timings.test.ts` — deterministic unit tests: full HTTPS timing, reused-socket/cached-DNS (`-1` phases), HAR-replay wall-clock fallback.
- [x] `api-network.test.ts` — asserts per-phase timings are populated from `APIResponse.timing()`.
- [x] `options.test.ts` — updated the hard-coded device user-agent that shifted with the Playwright bump (`Chrome/149…` → `151…`).
- [x] Full unit suite green (38 suites / 315 passed, 5 skipped), `tsc --noEmit`, ESLint, Prettier clean.
- [ ] Heartbeat/integration smoke recommended, since 1.61 → 1.62 pulls in all of 1.62's browser/behavior changes beyond this feature.


src/plugins/network.ts
```diff
@@ -197,45 +182,10 @@ export class NetworkManager {
       redirectURL: networkEntry.response.redirectURL,
     };
 
-    // Gather all resource timing information up until the
-    // TTFB(Time to first byte) is received
+    // Resource timing up to TTFB; `receive` fills in on `requestfinished`.
```
vigneshshanmugam: This comment is more robotic, lets reset to the previous version.

src/network-timings.ts
vigneshshanmugam, line 1: This file not a plugin per se, more of a helper for timing breakdown, lets move it somewhere else not under plugins directory.

src/plugins/network.ts
```diff
@@ -276,14 +226,10 @@ export class NetworkManager {
     if (!networkEntry) return;
 
     networkEntry.loadEndTime = epochTimeInSeconds();
-    // responseEnd is fired after the last byte of the response is received.
+    // Full timing (incl. `receive`) is available once the response completes.
```
vigneshshanmugam: same I prefer the old comment.

src/network-timings.ts
```diff
@@ -0,0 +1,129 @@
...
+    total: -1,
+  };
+}
+
+/**
+ * Sums the positive phases into `total`. Falls back to the wall-clock span
+ * between `requestSentTime` and `loadEndTime` when Resource Timing data is
+ * unavailable (`startTime <= 0`, e.g. HAR replay or aborted requests).
+ */
+export function calcTotalTime(
```
vigneshshanmugam, line 111: maybe we can change it to a pure func and return a separate object instead of overwiting the timing?

src/plugins/api-network.ts
vigneshshanmugam, line 98: I didnt review the piece that added this, any reason why we didnt use the Events like ApiRequest.Request and RequestFinished ? Wasnt that possible?

shahzad31: PW don't expose those unfortunately

src/plugins/network.ts
shahzad31: Done, restored the original comment 👍

src/plugins/network.ts
shahzad31: Reverted this one too.

src/network-timings.ts
shahzad31: Moved it out to `src/network-timings.ts` (out of the plugins dir).

src/network-timings.ts
shahzad31: Done — made it pure, it returns `total` now and the caller assigns it onto `timings.total`.

src/network-timings.ts
```diff
@@ -0,0 +1,130 @@
...
+  const dns =
+    rtiming.domainLookupEnd !== -1
+      ? roundMilliSecs(rtiming.domainLookupEnd - rtiming.domainLookupStart)
+      : -1;
+  const connect =
+    rtiming.connectEnd !== -1
+      ? roundMilliSecs(rtiming.connectEnd - rtiming.connectStart)
+      : -1;
+  const ssl =
+    rtiming.secureConnectionStart !== -1
```
mcapell, line 81: Should we check `rtiming.connectEnd` as well? something like:

```suggestion
    rtiming.secureConnectionStart !== -1 && rtiming.connectEnd !== -1
```

mcapell: I see it's guarded here: https://github.com/elastic/synthetics/pull/1164/changes#diff-8191a864bcd614940da81a1dbf38c09856ee074f07fc648eb3c47bcce77fc748R125 -- feel free to ignore it

mcapell (approved):
LGTM, added a small comment

PR #1166, opened 2026-08-04
test(e2e): add more stack versions to scheduled e2e matrix

shahzad31:
## 🎯 Summary

Widens e2e stack-version coverage, speeds up CI, and rewrites the tests to be
API-driven instead of Playwright-UI-driven — Kibana's Synthetics UI had
drifted out from under the old selectors. The version matrix is now resolved
live at run time instead of hand-maintained.

## 📦 What changed

| Area | Change |
|---|---|
| 🔢 Dynamic version matrix | Replaced the hand-curated `__tests__/e2e/versions` file with `__tests__/e2e/branches` (branch names, each with an optional permanently-pinned floor version, e.g. `8.19 8.19.19`) + `scripts/resolve_versions.js`, which resolves each branch's latest GA release and current dev snapshot **live** on every run — verified against the actual Docker registry (a HEAD request, no image data downloaded), since `artifacts-api`'s version list includes versions whose images haven't published yet. No more manual bumps, no drift-detection job needed (removed `check_new_snapshot.js`/`check-latest-snapshot` — there's nothing that can go stale). |
| 🔁 PR trigger | `e2e.yml` now also runs on `pull_request`, not just schedule/push to `main` |
| ⚡ Parallel CI | Sequential per-version loop → `prepare` job (resolves versions live) feeds a matrix `test` job, one runner per version, all concurrent |
| 🔕 Notify | Slack notify split into its own `notify` job, skipped on `pull_request` |
| 🧪 API-driven tests | `synthetics_api.journey.ts` — creates a private location + http/tcp/browser monitors via the Synthetics REST API, verifies **actual check status** (not just "a doc landed") in Elasticsearch. **No UI.** Replaces the old `synthetics.journey.ts` (deleted — its UI path only covered `<8.7.0`, and 7.17 — the only version below that threshold — is no longer backported to). |
| 🐳 Complete agent image | `setup_integration.sh` now forces `elastic-agent-complete` (namespace picked per version) — `elastic-package stack up` otherwise silently gives a non-"complete" image that can't run browser monitors at all |

`icmp` is intentionally not tested — it needs a raw-socket capability the
elastic-package-provisioned container doesn't grant, independent of stack
version.

## 🚨 Found a real elastic-agent regression, wider than originally reported

The new (correct) assertion — verifying the check actually succeeded, not
just that *some* result document landed — surfaced a real browser-monitor
regression across every currently active branch, well beyond the single
version reported in
[elastic/elastic-agent#15993](https://github.com/elastic/elastic-agent/issues/15993).
Every failure has the identical signature:
```
command '/usr/share/elastic-agent/.node/node/bin/elastic-synthetics ... --inline --rich-events --outfd 3' exited unexpectedly with code: 1
```

| Version | Result |
|---|---|
| `8.19.19` (pinned floor GA) | ✅ |
| `8.19.20-SNAPSHOT` | ❌ real regression |
| `9.4.4` (pinned floor GA) | ✅ |
| `9.4.5-SNAPSHOT` | ❌ real regression |
| `9.5.0` (GA) | ❌ real regression |
| `9.5.0-SNAPSHOT` | ❌ real regression |
| `9.6.0-SNAPSHOT` (main) | ❌ real regression |

Every GA/snapshot build from every currently active branch fails; only the
two permanently-pinned older-GA floors (predating the npm 12 bump) pass —
which is exactly their job: proving the test harness itself works even when
everything upstream is broken. (`8.19.20` and `9.4.5` GA don't have published
Docker images yet as of this writing, so they aren't in the matrix at all —
the resolver verifies that against the registry directly now, rather than
trusting `artifacts-api`'s version list, which lists versions before their
images actually exist — an issue caught the hard way during development of
this PR.)

Before this PR, none of this was visible in e2e — the old assertion couldn't
tell a fast-failing check from a successful one.

## ✅ Test plan

- [x] All matrix legs run in parallel, resolved live from `branches`
- [x] `8.19.19`, `9.4.4` (pinned floor versions) pass end-to-end, proving the harness itself works — [run 30994185246](https://github.com/elastic/synthetics/actions/runs/30994185246)
- [x] Every current GA/snapshot build correctly fails on the real elastic-agent regression


mcapell (dismissed):
This is awesome. Thank you!

mcapell (dismissed):

mcapell (approved):

PR #1172, opened 2026-08-13
fix: remove Husky from the published package

### LINKED ISSUE #1171: Published package runs unnecessary Husky prepare script
## Summary

Starting with @elastic/synthetics 1.24.1, the published package contains a Husky lifecycle script (https://github.com/elastic/synthetics/pull/1133):

```json
"prepare": "node .husky/install.mjs || true"
```

When consumers install or link the package, npm executes this script. In environments with restricted process creation, the additional shell/Node process can fail:

```text
npm error code 2
npm error command sh -c node .husky/install.mjs || true
npm error sh: 1: Cannot fork
```

Husky is a development tool, so its setup should not run when installing the published production package.

### Workaround

Consumers can disable lifecycle scripts:

```text
NPM_CONFIG_IGNORE_SCRIPTS=true
```

This is broad in scope and should only be considered a temporary workaround.

### Proposed fix

Remove the Husky prepare script and .husky/install.mjs from the published package while retaining Husky setup for repository development.

Add an npm pack test to verify that future published packages do not contain the Husky lifecycle script.

mcapell:
## Summary

 Remove the Husky prepare lifecycle script and lint-staged hook because
 linting and formatting are already enforced by CI. Document the equivalent
 local commands for contributors.

 Fixes #1171


shahzad31 (approved):
LGTM !!

PR #1187, opened 2026-09-10
fix: run inline api monitors as apiJourney, not journey

shahzad31:
## Summary

`--inline` mode (the only way Fleet/Kibana-authored monitors ever run) always wrapped the piped-in script in a plain `journey(...)` call, even for `api`-type monitors. This meant `journey.type` was always `"browser"`, so `Gatherer.setupDriver(options, journey.type)` always built a full browser `Driver` — never the `APIDriver` added in #997 — Chromium always launched, and network capture (`APINetworkManager`) never started. API Journey monitors produced no `journey/network_info` events at all, and silently ran as (slower, screenshot-taking) browser journeys instead of the fast, browser-less path #997 built.

Heartbeat already sets `ELASTIC_SYNTHETICS_MONITOR_TYPE=api` on the child process specifically to signal this (`x-pack/heartbeat/monitors/browser/synthexec/synthexec.go`), but nothing in this repo ever read it.

## Fix

`loadInlineScript` (in `src/loader.ts`) now checks `process.env.ELASTIC_SYNTHETICS_MONITOR_TYPE === 'api'` and wraps the legacy (non-`import`) inline script in `apiJourney('inline', ({ params, request }) => ...)` instead of `journey(...)` when true — matching what the `apiJourney(...)` module-import inline path already does correctly.

## How I verified this locally

- Built and ran the exact CLI invocation Heartbeat uses (`elastic-synthetics --inline --rich-events`) against a real Fleet-managed Elastic Agent running a Kibana-created API Journey monitor (elastic/kibana#270874) — before this fix, the run produced 64 `screenshot/block` events and zero `journey/network_info` events, confirming it silently executed as a browser journey.
- Added `cli.test.ts` › "runs inline api monitor (legacy step statements) without a browser", which sets `ELASTIC_SYNTHETICS_MONITOR_TYPE=api` and asserts a `journey/network_info` event is produced and no `screenshot/block` events are.
  - Confirmed this test **fails** (times out) on `main` without this fix, and on this branch with the fix reverted.
  - Confirmed it **passes** (in ~650ms, vs. the ~15s timeout without the fix) with the fix applied.
- `npm run build` (tsc) and `npm run lint` both pass clean.
- Full `__tests__/cli.test.ts` suite: all tests pass except a small number of pre-existing browser-launch timeouts (`runs inline tests`, `runs inline journey module with ESM imports`) that I confirmed **also fail identically on a clean, unmodified `main`** in this sandbox — unrelated to this change (Playwright/Chromium launch flakiness in the test environment, not something this PR touches).

## Related

- elastic/synthetics#997 (adds `apiJourney`/`APIDriver`/`APINetworkManager` — this PR is what actually lets Heartbeat reach that code path)
- elastic/beats#50802 (Heartbeat's `api` monitor type; sets the `ELASTIC_SYNTHETICS_MONITOR_TYPE` env var this PR finally reads)
- elastic/kibana#270874 (Kibana UI/Fleet support for Synthetics API Journey monitors; this is where I discovered the gap while testing end-to-end)

### DIFF
--- a/__tests__/cli.test.ts
+++ b/__tests__/cli.test.ts
@@ -120,6 +120,32 @@ describe('CLI', () => {
       expect(await cli.exitCode).toBe(0);
     });
 
+    it('runs inline api monitor via --inline-api without a browser', async () => {
+      // Same as above, but opting in with a flag instead of the
+      // Heartbeat-only ELASTIC_SYNTHETICS_MONITOR_TYPE env var, for callers
+      // invoking the CLI by hand.
+      const cli = new CLIMock()
+        .stdin(
+          `step('check body', async () => {
+          const resp = await request.get(params.url);
+          expect((await resp.body()).toString()).toMatch(/Synthetics/);
+        })`
+        )
+        .args([
+          '--inline',
+          '--inline-api',
+          '--rich-events',
+          '--params',
+          JSON.stringify(serverParams),
+        ])
+        .run();
+      await cli.waitFor('journey/end');
+      const output = cli.buffer().join('\n');
+      expect(getEvent(output, 'journey/network_info')).toBeTruthy();
+      expect(getEvent(output, 'screenshot/block')).toBeFalsy();
+      expect(await cli.exitCode).toBe(0);
+    });
+
     it('generate mfa totp token', async () => {
       const cli = new CLIMock()
         .stdin(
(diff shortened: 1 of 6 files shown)


mcapell:
I think we should add something here as well: https://github.com/elastic/synthetics/blob/main/src/cli.ts#L93

mcapell (approved):
LGTM!

